PR #19, opened 2021-08-04
SNAP

nmrodelo:


openfisca_us/entities.py
```diff
@@ -8,6 +8,16 @@
 
 from openfisca_core.entities import build_entity
 
+
+BenefitUnit = build_entity(
+    key="benefitunit",
+    plural="benefitunits",
```
nikhilwoodruff, line 14: Could separate with space

openfisca_us/entities.py
```diff
@@ -8,6 +8,16 @@
...
+    label="The benefit unit",
+    doc="""
+    Description of a benefit unit
+    """,
+)
+
 TaxUnit = build_entity(
```
nikhilwoodruff, line 21: I set taxunits as person-level entities, but eventually we'll probably need to return to them and make them groups with dependents in them?

openfisca_us/entities.py
```diff
@@ -8,6 +8,16 @@
...
 
+
+BenefitUnit = build_entity(
+    key="benefitunit",
+    plural="benefitunits",
+    label="The benefit unit",
+    doc="""
+    Description of a benefit unit
+    """,
+)
```
nikhilwoodruff, line 19: Error is being caused by `build_entity` needing a `roles` argument, see here from openfisca-uk:
```python
roles=[
        {
            "key": "adult",
            "plural": "adults",
            "label": u"Adult",
            "doc": u"The adults in the household.",
        },
        {
            "key": "child",
            "plural": "children",
            "label": u"Child",
            "doc": u"Dependent children in the household.",
        },
    ],
```

openfisca_us/variables/finance/benefit/inputs.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class gross_income(Variable):
+    value_type = float
+    entity = BenefitUnit
+    label = u"Income before SNAP program deductions are applied"
+    definition_period = YEAR
+
+
+class net_income(Variable):
```
nikhilwoodruff, line 13: Is this definitely an input variable? Will it ever have a formula?

openfisca_us/variables/finance/benefit/outputs.py
```diff
@@ -0,0 +1,27 @@
...
+        # Max Benefit
+        # Phase-out Rate
+        # for later, Employment Rules
+
+        # Get vars:
+        # net income
+        # gross income
+        # number of individuals
+
+        # for later, Employment Status
```
nikhilwoodruff, line 25: Good placeholder

nikhilwoodruff (commented):
Looks good! Added the reason for the test fail, and a few cosmetic suggestions but otherwise on the right track for sure.

openfisca_us/variables/finance/benefit/inputs.py
MaxGhenis: It'll probably be a formula

openfisca_us/entities.py
MaxGhenis: Does it need separate roles? I don't think there's a "head of household" concept for SPM units.

openfisca_us/variables/finance/benefit/outputs.py
MaxGhenis: although actually might not be possible since we're looking over a year, might be weeks unemployed instead

nikhilwoodruff:
@MaxGhenis , @nmrodelo just have a small question about the benefit unit entity here (please feel free to correct anything I've mixed up here). Looking at the entities we'll need to create for OpenFisca-US, so far I've come across:
- Person
- Tax Unit
- Benefit Unit
- CPS ASEC Family
- Household
- SPM Unit

I'd just like to make sure I've got the right definitions for these. 

The [CPS technical documentation](https://www.census.gov/programs-surveys/cps/technical-documentation/subject-definitions.html#family) defines families as:
> A family is a group of two people or more (one of whom is the householder) related by birth, marriage, or adoption and residing together; all such people (including related subfamily members) are considered as members of one family.

[This](https://www.masslegalhelp.org/income-benefits/food-stamps/advocacy-guide/part2/q30-snap-household-or-unit-article) implies people unrelated but living together can be in the same SNAP unit, so if benefit unit here distinct from the UK definition of one or two adults plus dependent children? If so, I'd favour using an alternative name. 

Also, I've looked around, but I'm not sure I really understand the relationship between SPM unit and household.

MaxGhenis:
Sorry, we should have named it SPM unit, which is the same as benefit unit. The hierarchy is:
* Person
* TaxUnit
* Family
* SPM Unit
* Household

openfisca_us/variables/finance/benefit/inputs.py
nmrodelo: rename to spm_gross_income and move to finance/benefits/outputs.py

MaxGhenis:
@nmrodelo let's go through the merge conflict tomorrow

openfisca_us/parameters/benefit/SNAP/amounts/household_size.yaml
```diff
@@ -0,0 +1,40 @@
+description:  Maximum monthly SNAP benefit by household size
+
+1:
+  values:
+    2021-01-01: 204
```
MaxGhenis, line 5: Per the changes scheduled for October (see #27), verify if these should be `2020-10-01` instead

openfisca_us/tests/baseline/finance/benefit/SNAP.yml
```diff
@@ -0,0 +1,131 @@
+- name: Individual, $50k income
+  period: 2019
```
MaxGhenis, line 2:
```suggestion
  period: 2021
```
and others below

openfisca_us/tests/baseline/finance/benefit/SNAP.yml
```diff
@@ -0,0 +1,131 @@
...
+    snap: 2448  # 204 * 12
+
+- name: CBPP family A  # TODO: can we add a source?
+  # https://www.cbpp.org/research/food-assistance/the-supplemental-nutrition-assistance-program-includes-earnings-incentives
+  period: 2019
+  absolute_error_margin: 0
+  input:
+    people:
+      person_1:
+        earned: 20784  # 1732 * 12
```
MaxGhenis, line 36:
```suggestion
        earned: 1732 * 12
```
and below

openfisca_us/parameters/benefit/SNAP/deductions/shelter_deduction.yaml
```diff
@@ -0,0 +1,318 @@
+description:  Shelter Deduction by State
+
+states:
+
+      [AL]:
```
MaxGhenis, line 5: please remove empty lines and the square brackets

openfisca_us/parameters/benefit/SNAP/deductions/standard_deduction.yaml
```diff
@@ -0,0 +1,1496 @@
...
+        5:
+          values:
+            2021-01-01: 244
+        
+        6:
+          values:
+            2021-01-01: 280
+        
+
+        NOTE: NEEDS TO BE SIX OR MORE FIXED AT 280
```
MaxGhenis, line 1487: yaml accepts `# comments`

MaxGhenis (commented):
@nmrodelo these are all the same except Alaska and Hawaii right? That would follow the poverty lines. If so I think we can simplify somehow...@nikhilwoodruff how would you suggest?

MaxGhenis:
Also if the gross and net income tests are multiples of the poverty line, let's just enter the poverty lines and then those multiples as policy parameters.

nikhilwoodruff:
> @nmrodelo these are all the same except Alaska and Hawaii right? That would follow the poverty lines. If so I think we can simplify somehow...@nikhilwoodruff how would you suggest?

OK I think I understand. I think a good way to deal with these situations in general with outlier states is to add a "default" value just like another state, and in the variable, instead of indexing with the state name, map the state name array to {"alaska" -> "alaska", ..., any of the default states -> "default}, and use that to index instead.

openfisca_us/variables/finance/benefit/outputs.py
```diff
@@ -0,0 +1,67 @@
...
+
+class maximum_snap_benefit(Variable):
+    value_type = float
+    entity = SPMUnit
+    label = u"Supplemental Nutrition Assistance Program, spm unit's max possible unit based on family size"
+    definition_period = YEAR
+
+    def formula(spm_unit, period, parameters):
+
+        max_by_household_size = parameters(
```
nikhilwoodruff, line 62: This applies to lots of other instances (including on the tax side), but bear in mind that because this formula is calculatable for any year after 0 AD (`formula(` is effectively `formula_0000(`), the model at the moment will break if trying to calculate anything involving this variable for years before 2021, because it won't find the parameter for that year. I'd suggest changing `formula` to `formula_2021` until we decide to add older values. This'll make SNAP zero before 2021.

openfisca_us/variables/finance/benefit/inputs.py
```diff
@@ -0,0 +1,24 @@
+from openfisca_core.model_api import *
+from openfisca_us.entities import *
+from openfisca_us.tools.general import *
+
+
+class gross_income(Variable):
```
nikhilwoodruff, line 6: This is a very general name so would advise either prepending `snap_` or moving elsewhere if it's general enough

openfisca_us/variables/finance/benefit/inputs.py
nikhilwoodruff, line 13: Also very general name, see above

openfisca_us/variables/finance/benefit/inputs.py
```diff
@@ -0,0 +1,24 @@
...
+    value_type = float
+    entity = BenefitUnit
```
nikhilwoodruff, line 8: I think this just would need to change to SPMUnit

openfisca_us/variables/finance/benefit/inputs.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class net_income(Variable):
+    value_type = float
+    entity = BenefitUnit
+    label = u"Income after SNAP program deductions are applied"
+    definition_period = YEAR
+
+
+class people(Variable):
```
nikhilwoodruff, line 20: I don't think this is used so we could just take it out, but ideally `people` is person-level, so that `sim.calc("people", map_to="any_entity")` gives person counts

openfisca_us/parameters/benefit/SNAP/amounts/household_size.yaml
```diff
@@ -0,0 +1,1977 @@
+description:  Maximum Monthly SNAP Benefit by State and Household Size
```
MaxGhenis, line 1: Per https://github.com/UBICenter/openfisca-us/issues/48#issuecomment-903856922 let's name this file `benefit/SNAP/amount.yaml`

MaxGhenis (changes requested):
I think we should also split this into multiple PRs so we can merge the parameters without waiting on the logic, which could be more difficult.

MaxGhenis:
Per today's meeting, confirming that we'll split this out into one PR for the parameters and one for the logic, maybe splitting out the logic into separate PRs (e.g. for net/gross income variables) as well.

We should also create a new variable for SNAP's geographic grouping, since that varies by contiguous 48 and also within Alaska.

PR #23, opened 2021-08-07
Add entities

nikhilwoodruff:
This adds the person, tax unit, family, SPM unit and household entities, and switches to using the `BaseCPS` dataset in `openfisca-data` (relevant [PR](https://github.com/UBICenter/openfisca-data/pull/1)).

### DIFF
--- a/openfisca_us/entities.py
+++ b/openfisca_us/entities.py
@@ -1,17 +1,9 @@
-"""
-This file defines the entities needed by our legislation.
-
-Taxes and benefits can be calculated for different entities: persons, household, companies, etc.
-
-See https://openfisca.org/doc/key-concepts/person,_entities,_role.html
-"""
-
 from openfisca_core.entities import build_entity
 
 TaxUnit = build_entity(
-    key="taxunit",
-    plural="taxunits",
-    label="Tax Unit",
+    key="tax_unit",
+    plural="tax_units",
+    label="Tax unit",
     doc="""
     A tax unit.
     """,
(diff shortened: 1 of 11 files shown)


openfisca_us/variables/entity/person.py
```diff
@@ -2,8 +2,22 @@
 from openfisca_us.entities import *
 
 
+class P_person_id(Variable):
```
MaxGhenis, line 5: why not `person_id`?

openfisca_us/variables/entity/taxunit.py
```diff
@@ -2,6 +2,27 @@
 from openfisca_us.entities import *
 
 
+class taxunit_id(Variable):
```
MaxGhenis, line 5: inconsistent with `spm_unit_id`

openfisca_us/variables/entity/person.py
nikhilwoodruff: Changed

openfisca_us/entities.py
```diff
@@ -1,44 +1,77 @@
...
+    plural="families",
+    label="Family",
+    doc="""
+    A family.
+    """,
+    roles=[
+        dict(
+            key="member",
+            label="Member",
+            doc="A member of the group",
```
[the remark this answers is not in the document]
nikhilwoodruff: Changed

openfisca_us/variables/entity/taxunit.py
nikhilwoodruff: Yep, changed

Issue #30, opened 2021-08-21
Remove segmentation of rates by filer type


MaxGhenis:
Only brackets vary with filer type, not rates. See https://github1s.com/UBICenter/openfisca-us/blob/master/openfisca_us/parameters/tax/income/bracket.yaml

PR #33, opened 2021-08-21
Add capital gains parameters

MaxGhenis:
Only baseline parameters (no AMT, no CG calculation logic, tests, etc.)

From https://github.com/PSLmodels/Tax-Calculator/blob/master/taxcalc/policy_current_law.json#L6685-L7353

NB: I used the current system of multiple rates per filer status, but this may change, see #30


nikhilwoodruff:
Looks good, thanks @MaxGhenis

PR #55, opened 2021-08-24
Add additional medicare tax parameter

MaxGhenis:
https://taxcalc.pslmodels.org/guide/policy_params.html#additional-medicare-fica

cc @fedderw 

todo: add metadata (see #57)

### DIFF
--- a/openfisca_us/parameters/tax/payroll/FICA/medicare/additional
+++ b/openfisca_us/parameters/tax/payroll/FICA/medicare/additional
@@ -1,0 +1,14 @@
+description: The Additional Medicare Tax rate is applied for earnings above an exclusion.
+rate:
+  2013-01-01: 0.009
+exclusion:
+  SINGLE:
+    2013-01-01: 200_000
+  JOINT:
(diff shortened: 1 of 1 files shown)


openfisca_us/parameters/tax/payroll/FICA/medicare/additional
```diff
@@ -0,0 +1,14 @@
+description: The Additional Medicare Tax rate is applied for earnings above an exclusion.
+rate:
+  2013-01-01: 0.009
+exclusion:
+  SINGLE:
+    2013-01-01: 200_000
```
MaxGhenis, line 6: @nikhilwoodruff wdyt about making this `_` a standard practice?

nikhilwoodruff: I like this, no downsides IMO, it's just more readable. Plus, it'll be trivial to convert all existing parameters into this format if we go with it.

nikhilwoodruff (approved):

PR #142, opened 2021-09-20
Add SPM income and poverty thresholds

nikhilwoodruff:
Also adds the poverty function.

### DIFF
--- a/openfisca_us/api/microsimulation.py
+++ b/openfisca_us/api/microsimulation.py
@@ -114,7 +114,7 @@ class Microsimulation:
         else:
             return arr
 
-    def df(self, variables, period=2019):
+    def df(self, variables, period=2020):
         df_dict = {}
         var_metadata = self.simulation.tax_benefit_system.variables[
             variables[0]
--- a/openfisca_us/api/microsimulation.py
+++ b/openfisca_us/api/microsimulation.py
@@ -2,7 +2,7 @@ from microdf.generic import MicroDataFrame
 import numpy as np
 from openfisca_core import periods
 from openfisca_core.model_api import *
-from openfisca_us_data import BaseCPS, RawCPS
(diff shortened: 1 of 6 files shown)


openfisca_us/api/microsimulation.py
```diff
@@ -30,65 +30,100 @@ def apply_reforms(self, reforms: list) -> None:
...
+        target_pop = self.simulation.populations[target_entity]
+        GROUP_ENTITIES = ("tax_unit", "family", "spm_unit", "household")
+        if entity == "person" and target_entity in GROUP_ENTITIES:
+            if how and how not in (
+                "sum",
+                "any",
+                "min",
+                "max",
+                "all",
+                "value_from_first_person",
```
MaxGhenis, line 74: not just `first`?

openfisca_us/api/microsimulation.py
```diff
@@ -30,65 +30,100 @@ def apply_reforms(self, reforms: list) -> None:
...
-                    f"{entity_key}_weight", period=period, weighted=False
+                    f"{entity}_weight", period=period, weighted=False
                 ),
             )
             return series
         else:
             return arr
 
-    def df(self, variables, year=2019):
+    def df(self, variables, period=2019):
```
MaxGhenis, line 117:
```suggestion
    def df(self, variables, period=2020):
```

MaxGhenis (approved):

openfisca_us/api/microsimulation.py
nikhilwoodruff: Unfortunately not, these are [openfisca-implemented functions](https://github1s.com/openfisca/openfisca-core/blob/HEAD/openfisca_core/populations/group_population.py) rather than pandas function.

PR #144, opened 2021-09-21
Implement small tax functions

### LINKED ISSUE #111: Add IITAX

### LINKED ISSUE #117: Add LumpSumTax

### LINKED ISSUE #118: Add ExpandIncome

### LINKED ISSUE #119: Add AfterTaxIncome

nikhilwoodruff:
Fixes #119 
Fixes #111 
Fixes #118 
Fixes #117 (assuming we should skip, as this seems to be a head tax that isn't in the baseline)

### DIFF
--- a/openfisca_us/parameters/tax/credits/child_tax_credit/refundable.yaml
+++ b/openfisca_us/parameters/tax/credits/child_tax_credit/refundable.yaml
@@ -1,0 +1,7 @@
+description: Whether the child tax credit is fully refundable
+metadata:
+  unit: bool
+values:
+  2013-01-01: false
+  2021-01-01: true
+  2022-01-01: false
--- a/openfisca_us/parameters/tax/credits/child_tax_credit/refundable.yaml
+++ b/openfisca_us/parameters/tax/credits/child_tax_credit/refundable.yaml
@@ -1,4 +1,4 @@
-description: Whether the child tax credit is fully refundable
+description: Whether the Child Tax Credit is fully refundable
 metadata:
   unit: bool
 values:
--- a/openfisca_us/variables/finance/tax/outputs.py
+++ b/openfisca_us/variables/finance/tax/outputs.py
@@ -94,7 +94,7 @@ class combined(Variable):
     value_type = float
     entity = TaxUnit
     definition_period = YEAR
-    documentation = """Sum of iitax and payrolltax and lumpsum_tax"""
+    documentation = """Sum of iitax and payrolltax"""
 
     def formula(tax_unit, period, parameters):
         return add(tax_unit, period, "iitax", "payrolltax")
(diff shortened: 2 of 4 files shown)


openfisca_us/parameters/tax/credits/child_tax_credit/refundable.yaml
```diff
@@ -0,0 +1,7 @@
+description: Whether the child tax credit is fully refundable
```
MaxGhenis, line 1:
```suggestion
description: Whether the Child Tax Credit is fully refundable
```

openfisca_us/variables/finance/tax/outputs.py
```diff
@@ -188,13 +194,47 @@ class expanded_income(Variable):
         """Broad income measure that includes benefit_value_total"""
     )
 
+    def formula(tax_unit, period, parameters):
```
MaxGhenis, line 197: Expanded income isn't used for tax calculations, but it is what taxcalc uses as their income measure for binning into deciles. Should we include it?

openfisca_us/variables/finance/tax/outputs.py
```diff
@@ -944,3 +994,22 @@ class nontaxable_ubi(Variable):
     entity = TaxUnit
     definition_period = YEAR
     documentation = """Amount of UBI benefit excluded from AGI"""
+
+
+class aftertax_income(Variable):
```
MaxGhenis, line 999: similarly, aftertax_income and benefit_value_total aren't used for calculations. aftertax_income is the net income without benefits, and therefore used for summarizing tax reforms. Could be important for testing for now but I think eventually at least we may want to remove it.

openfisca_us/variables/finance/tax/outputs.py
```diff
@@ -944,3 +994,22 @@ class nontaxable_ubi(Variable):
...
+    value_type = float
+    entity = TaxUnit
+    label = "After-tax income"
+    definition_period = YEAR
+
+    def formula(tax_unit, period, parameters):
+        expanded = tax_unit("expanded_income", period)
+        combined = tax_unit("combined", period)
```
MaxGhenis, line 1007:
```suggestion
        combined_tax = tax_unit("combined", period)
```

MaxGhenis (approved):

openfisca_us/variables/finance/tax/outputs.py
nikhilwoodruff: I can see benefit in including it (even if it's redundant/unused) for validation later on

openfisca_us/variables/finance/tax/outputs.py
nikhilwoodruff: Yeah I agree, but I think it might still be helpful when validating the model

PR #146, opened 2021-09-23
Age groups + speedup

nikhilwoodruff:
* Adds a variable `age_group` (child, working-age, senior)
* Fixes a bug which caused Microsimulation instances to take around 15 seconds to load (< 2 seconds now)


openfisca_us/variables/demographic/tax_unit.py
```diff
@@ -2,20 +2,35 @@
...
+    entity = Person
+    label = u"Unique reference for the tax unit of this person"
+    definition_period = ETERNITY
+
+
+class MARSType(Enum):
+    SINGLE = "Single"
+    JOINT = "Joint"
+    SEPARATE = "Separate"
+    HOUSEHOLD_HEAD = ("Head of household",)
```
MaxGhenis, line 30: Why parens and comma?

MaxGhenis (approved):

openfisca_us/variables/demographic/tax_unit.py
nikhilwoodruff: Dunno, ask `black`. Will revert

PR #156, opened 2021-09-28
Move duplicated code to PolicyEngine-Core

nikhilwoodruff:
This will ensure parity with the UK model.


openfisca_us/api/hypothetical.py
```diff
@@ -1,185 +1,10 @@
-import numpy as np
-import openfisca_us
-from openfisca_core.simulation_builder import SimulationBuilder
-from openfisca_core.periods import period
+from policyengine_core import IndividualSim as GeneralIndividualSim
```
MaxGhenis, line 1: Would it make sense to rename this in policyengine_core?

MaxGhenis (approved):
image
❤️

PR #177, opened 2021-10-09
Add NSLP parameters

tolaouk:
Add 130% and 185%

### DIFF
--- a/openfisca_us/tests/baseline/finance/benefit/school_meals.yaml
+++ b/openfisca_us/tests/baseline/finance/benefit/school_meals.yaml
@@ -9,7 +9,7 @@
         age: 10
     spm_units:
       spm_unit:
-        members: [parent,child]
+        members: [parent, child]
         spm_unit_state_group: CONTIGUOUS_US
         poverty_ratio: 0 
   output:
@@ -27,7 +27,7 @@
         age: 10
     spm_units:
       spm_unit:
-        members: [parent,child]
+        members: [parent, child]
         spm_unit_state_group: CONTIGUOUS_US
         poverty_ratio: 1.2
   output:
@@ -45,7 +45,7 @@
         age: 10
     spm_units:
       spm_unit:
-        members: [parent,child]
+        members: [parent, child]
         spm_unit_state_group: CONTIGUOUS_US
         poverty_ratio: 1.8
   output:
@@ -63,7 +63,7 @@
         age: 10
     spm_units:
       spm_unit:
-        members: [parent,child]
+        members: [parent, child]
         spm_unit_state_group: CONTIGUOUS_US
         poverty_ratio: 10
   output:
@@ -80,7 +80,7 @@
         age: 10
     spm_units:
       spm_unit:
-        members: [parent,child]
+        members: [parent, child]
         spm_unit_state_group: HI
         poverty_ratio: 0
   output:
@@ -98,7 +98,7 @@
         age: 10
     spm_units:
       spm_unit:
-        members: [parent,child]
+        members: [parent, child]
         spm_unit_state_group: HI
         poverty_ratio: 1.0
   output:
@@ -116,7 +116,7 @@
         age: 10
     spm_units:
       spm_unit:
-        members: [parent,child]
+        members: [parent, child]
         spm_unit_state_group: HI
         poverty_ratio: 1.5
   output:
@@ -134,7 +134,7 @@
         age: 10
     spm_units:
       spm_unit:
-        members: [parent,child]
+        members: [parent, child]
         spm_unit_state_group: HI
         poverty_ratio: 20
   output:
@@ -152,7 +152,7 @@
         age: 10
     spm_units:
       spm_unit:
-        members: [parent,child]
+        members: [parent, child]
         spm_unit_state_group: AK
         poverty_ratio: 0.8
   output:
@@ -170,7 +170,7 @@
         age: 10
     spm_units:
       spm_unit:
-        members: [parent,child]
+        members: [parent, child]
         spm_unit_state_group: AK
         poverty_ratio: 1.4
   output:
@@ -190,7 +190,7 @@
         age: 10
     spm_units:
       spm_unit:
-        members: [parent,child1,child2]
+        members: [parent, child1, child2]
         spm_unit_state_group: CONTIGUOUS_US
         poverty_ratio: 1.2
   output:
@@ -210,7 +210,7 @@
         age: 10
     spm_units:
       spm_unit:
-        members: [parent,child1,child2]
+        members: [parent, child1, child2]
         spm_unit_state_group: HI
         poverty_ratio: 1.7
   output:
@@ -252,7 +252,7 @@
         age: 6
     spm_units:
       spm_unit:
-        members: [parent,child1,child2,child3]
+        members: [parent, child1, child2, child3]
         spm_unit_state_group: CONTIGUOUS_US
         poverty_ratio: 15
   output:
@@ -274,7 +274,7 @@
         age: 5
     spm_units:
       spm_unit:
-        members: [parent,child1,child2,child3]
+        members: [parent, child1, child2, child3]
         spm_unit_state_group: AK
         poverty_ratio: 1.6
   output:
@@ -294,7 +294,7 @@
         age: 16
     spm_units:
       spm_unit:
-        members: [parent1,parent2,child]
+        members: [parent1, parent2, child]
         spm_unit_state_group: CONTIGUOUS_US
         poverty_ratio: 1.8
   output:
@@ -314,7 +314,7 @@
         age: 12
     spm_units:
       spm_unit:
-        members: [parent1,parent2,child]
+        members: [parent1, parent2, child]
         spm_unit_state_group: HI
         poverty_ratio: 1.1
   output:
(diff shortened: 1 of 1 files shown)


openfisca_us/parameters/benefit/NSLP/benefit.yaml
```diff
@@ -0,0 +1,14 @@
+description: Annual value of NSLP per child
+
+2021-07-16: 595.8
```
nikhilwoodruff, line 3: Nice - not a breaking issue but could we add parameters covering the start of 2021 and 2021? I had a look at the second reference (NSLP - Rates of Reimbursement) but I couldn't seem to find the values, or they weren't loading.

openfisca_us/parameters/benefit/NSLP/benefit.yaml
```diff
@@ -0,0 +1,14 @@
...
+
+
+metadata:
+  unit: currency-USD
+  reference: 
+  - title: Federal Reserve Bank of Atlanta - Policy Rules Database
+    href: https://www.atlantafed.org/-/media/documents/economic-mobility-and-resilience/advancing-careers-for-low-income-families/policy-rules-database/policy-rules-database-technical-manual.pdf
+  reference:
```
nikhilwoodruff, line 11: I think we just need to remove the `reference:` line here and this should be valid.

nikhilwoodruff (commented):
Thanks @tolaouk - looking good, added a few comments.

MaxGhenis:
@tolaouk I gave you access to the private https://github.com/PolicyEngine/PolicyRulesDatabase repo, which contains data from the Atlanta Fed. In https://github.com/PolicyEngine/PolicyRulesDatabase/blob/main/Version%2009-30-2021/CostDatabase.xlsx, the `schoolmealsCost` tab points to https://schoolnutrition.org/aboutschoolmeals/schoolmealtrendsstats/#3 which includes:
image

This indicates the 0.7 reduced-price copay from the PRD actually means $0.70, not 70%. So I'd suggest removing `copay.yaml` and changing `amount.yaml` to have two values for each level:
* Free: `180 * ((3.66 - 0.35) + (1.97 - 0.33))` (yaml accepts arithmetic expressions, this equals $891)
* Reduced price: `180 * ((3.26 - 0.35) + (1.67 - 0.33))` ($765)

Note that this exceeds the PRD's estimate, which is scaled up from 2017 with inflation: `4.15 * 180 * 1.02 ^ 4 = 809` for the free plan. I just emailed them to suggest adjusting the amount to these values.

Under this approach we should also rename the folder to `NSLP_SBP`.

Alternatively we could be more explicit and separate out each of the parameters for SBP, NSLP, and reimbursement rates for Free/Reduced price/Paid, and also add number of school days (180) as its own parameter. @nikhilwoodruff wdyt?

MaxGhenis:
Ah, [here's the ticket](https://www.federalregister.gov/documents/2021/07/16/2021-15107/national-school-lunch-special-milk-and-school-breakfast-programs-national-average-paymentsmaximum):
image

I think we'd ideally get this full table in a yaml structure like: `parameters/benefits/school_meals/{NSLP,SBP,milk,snacks}.yaml`, where each file is broken down into `contiguous_US,AK,HI` (like [SNAP parameters](https://github.com/PolicyEngine/openfisca-us/blob/master/openfisca_us/parameters/benefit/SNAP/standard_deduction.yaml)), and within that, `free,reduced_price,paid`. For simplicity we can just take the `LESS THAN 60%` column for now, i.e. assume all schools have fewer than 60% of kids below 185% of the poverty line.

Here's the [equivalent document](https://www.federalregister.gov/documents/2020/07/22/2020-15764/national-school-lunch-special-milk-and-school-breakfast-programs-national-average-paymentsmaximum) for the 2020-2021 school year.

tolaouk:
Hi @MaxGhenis, are we meeting today at 1?

MaxGhenis:
Yes https://meet.google.com/ihc-uwti-yfw

MaxGhenis:
Thanks @nikhilwoodruff for helping with these and demonstrating it in a [notebook](http://nbviewer.org/gist/nikhilwoodruff/70e9848ec42321993636de888f95a40b).

@tolaouk to run that kind of example on your machine, you'll need to
a) Install the latest version of policyengine-core:
```shell
pip install policyengine-core
```
b) Install the latest version of openfisca-us; this will be easiest done locally to reflect the changes you're making. So from your openfisca-us folder, do:
```shell
pip install -e .
```
`-e` keeps the installation updated based on the contents of the folder.

c) [Install Anaconda](https://www.anaconda.com/products/individual) to have the rest of the packages available.

openfisca_us/tests/baseline/finance/benefit/school_meals.yaml
```diff
@@ -0,0 +1,1826 @@
...
+        state: HI
+      child1:
+        age: 15
+      child2:
+        age: 12
+      child3:
+        age: 10
+      child4:
+        age: 7
+      child4:
```
MaxGhenis, line 1790:
```suggestion
      child5:
```
this should fix the breaking test

openfisca_us/variables/finance/benefit/school_meals.py
```diff
@@ -0,0 +1,43 @@
...
+        # Get state and poverty ratio for SPM unit.
+        state_group = spm_unit("spm_unit_state_group", period)
+        poverty_ratio = spm_unit("poverty_ratio", period)
+        # Get parameters.
+        p_school_meals = parameters(period).benefit.school_meals
+        p_income_limit = p_school_meals.income_limit
+        # Look up the free/reduced/full subsidy tier for each SPM unit by
+        # poverty ratio.
+        tier = where(
+            poverty_ratio < p_income_limit["free"],
```
nikhilwoodruff, line 24: I might suggest a `select` (alias for `np.select`) statement here to make it more readable:
```python
tier = select([
    poverty_ratio < p_income_limit.free,
    poverty_ratio < p_income_limit.reduced_price,
    poverty_ratio >= p_income_limit.reduced_price,
], ["free", "reduced_price", "paid"])

**Review by nikhilwoodruff — COMMENTED** [review]
Thanks @tolaouk - added a few changes to get it running and a minor suggestion here.

**Comment by tolaouk** [comment]
screenshot

hi @MaxGhenis  I think the program uses this date 2021-01-01 to calculate the subsidy for the 2021 test case while the effective date for the test case is 2021-07-01. That's why it's off by 41.4 which is the difference between the 2020-2021 and the 2021-2022 school year. Where can this be fixed?

**Comment by MaxGhenis** [comment]
@tolaouk as you discovered, tests for a given year use parameter values as of January 1 of that year. If you'd like to test the current parameter values, you'll have to use 2022 as the test year.

File: openfisca_us/parameters/benefit/school_meals/income_limit.yaml [hunk 5]
```diff
@@ -0,0 +1,15 @@
+description: Income limit as share of poverty line
+
+free:
+  2020-07-01: 1.3
+  2021-07-01: 1.3
```
MaxGhenis, line 5: @nikhilwoodruff I forgot what standard we're setting here, should we drop lines when the policy hasn't changed?

openfisca_us/parameters/benefit/school_meals/income_limit.yaml
```diff
@@ -0,0 +1,15 @@
...
+reduced_price:
+  2020-07-01: 1.85
+  2021-07-01: 1.85
+
+  
+metadata:
+  unit: /1
+  reference: 
+  - title: The National School Lunch Program
+    href: https://fns-prod.azureedge.net/sites/default/files/resource-files/NSLPFactSheet.pdf
```
MaxGhenis, line 15: could you add a reference for 2020?

openfisca_us/parameters/benefit/school_meals/amount.yaml
```diff
@@ -0,0 +1,76 @@
...
+      2021-07-01: 1.99
+    free:
+      2020-07-01: 2.21
+      2021-07-01: 2.29
+
+
+metadata:
+  unit: currency-USD
+  reference:
+  - title: National School Lunch, Special Milk, and School Breakfast Programs, National Average Payments/Maximum Reimbursement Rates
```
MaxGhenis, line 72: Could you add the year to the titles?

openfisca_us/tests/baseline/finance/benefit/school_meals.yaml
```diff
@@ -0,0 +1,468 @@
...
+      child5:
+        age: 5
+    spm_units:
+      spm_unit:
+        members: [parent1,parent2,child1,child2,child3,child4,child5]
+        spm_unit_state_group: CONTIGUOUS_US
+        poverty_ratio: 1.1
+  output:
+    school_meal_subsidy: 180 * (3.66 + 1.97) * 5
+
```
MaxGhenis, line 468: delete second trailing newline

openfisca_us/tests/baseline/finance/benefit/school_meals.yaml
```diff
@@ -0,0 +1,468 @@
...
+  absolute_error_margin: 0
+  input:
+    people:
+      parent:
+        age: 30
+      child:
+        age: 10
+    spm_units:
+      spm_unit:
+        members: [parent,child]
```
MaxGhenis, line 12: nit: @nikhilwoodruff should these have spaces?

openfisca_us/parameters/benefit/school_meals/income_limit.yaml
nikhilwoodruff: I think I remember when we talked about it we decided on keeping even redundant lines.

PR #178, opened 2021-10-10
SNAP amount

MaxGhenis:
Maximum allotment


MaxGhenis:
Merging so we can move ahead with variables, but we might want to revisit the `main`/`additional` elements

PR #194, opened 2021-10-16
CCDF for New York

tolaouk:
Adding CCDF eligibility parameters and variables for New York


MaxGhenis:
Ah actually the rates vary by county as well, per https://ocfs.ny.gov/main/policies/external/OCFS_2016/LCMs/16-OCFS-LCM-18.pdf#page=14

Though that report is from 2016, we should use a newer one.

These market rates aren't in the PRD, so we should also see how they do it.

openfisca_us/variables/finance/benefit/ccdf.py
```diff
@@ -0,0 +1,130 @@
...
+    entity = SPMUnit
+    definition_period = YEAR
+    documentation = ""
+
+    def formula(spm_unit, period, parameters):
+        # Get state and poverty ratio for SPM unit.
+        state = spm_unit("spm_unit_state", period)
+        assets = spm_unit("spm_unit_assets", period)
+        # Get parameters.
+        p_asset_limit = parameters(period).benefit.CCDF.eligibility.assets[
```
MaxGhenis, line 18:
```suggestion
        p_asset_limit = parameters(period).benefit.ccdf.eligibility.assets[
```
Once #229 is merged, all variables and parameters will be lowercase, so you'll need to update a few places here.

openfisca_us/parameters/benefit/CCDF/eligibility.yaml
```diff
@@ -0,0 +1,19 @@
+description: CCDF eligibility by income and assets.
+
+initial_income:
```
MaxGhenis, line 3: Make `eligibility` a folder and add an `income.yaml` file for initial and continuous eligibility by state and household size

openfisca_us/variables/finance/benefit/ccdf.py
```diff
@@ -1,6 +1,133 @@
...
+
+
+class ccdf_max_benefit_per_child(Variable):
+    value_type = float
+    entity = SPMUnit
+    definition_period = YEAR
+    documentation = ""
+
+
+class ccdf_child_age(Variable):
```
MaxGhenis, line 88: delete

MaxGhenis:
@tolaouk we can close this now right?

tolaouk:
> @tolaouk we can close this now right?

yes we can close this. @MaxGhenis

PR #195, opened 2021-10-17
Move state and state_group to household entity

MaxGhenis:


openfisca_us/variables/finance/benefit/school_meals.py
```diff
@@ -14,7 +15,9 @@ class school_meal_subsidy(Variable):
 
     def formula(spm_unit, period, parameters):
         # Get state and poverty ratio for SPM unit.
-        state_group = spm_unit("spm_unit_state_group", period)
+        state_group = spm_unit.value_from_first_person(
```
MaxGhenis, line 18: @nikhilwoodruff I think this is where tests are breaking. How does one get household-level variables for a SPM unit? I tried adapting some code from openfisca-uk here but it's not working.

nikhilwoodruff: I filed this a while ago in core: https://github.com/openfisca/openfisca-core/issues/1041 - the solution is in the first snippet, but I think the second snippet would be much better, will see if I should submit a pull request.

MaxGhenis: That approach breaks as it returns `[0]` as `state_group`:
```
state_group = spm_unit.value_from_first_person(
    spm_unit.members.household("state_group", period)
)
```

nikhilwoodruff: Oh yeah, I remember there's a few places in openfisca-uk where we have to use Series objects and groupbys, because the value_from_first_person doesn't preserve the Enum attributes.

MaxGhenis: Adding `decode_to_str` worked!

openfisca_us/tests/baseline/finance/benefit/school_meals.yaml
```diff
@@ -10,13 +10,15 @@
...
+        poverty_ratio: 0
+    households:
+      household:
+        state_group: CONTIGUOUS_US
   output:
     school_meal_subsidy: 180 * (3.66 + 1.97)
 
 
-- name: Low income, one parent, one child, contiguous_US
+- name: Low income, one parent, one child, contiguous_US (defined as default)
```
MaxGhenis, line 21: This test breaks without defining `state_group` explicitly. @nikhilwoodruff should this work?

MaxGhenis: Specifying the members of the household doesn't fix the error, but it does goes away if I specify `state_group` in the household, as per the other tests.

nikhilwoodruff:
@MaxGhenis Not sure if I've fixed all the bugs but found one - it was to do with the Enum value type needing some non-obvious encode/decode calls. I think the State variable should be an Enum here too - made a start to pass the tests.

MaxGhenis:
Thanks, I needed the `[0]` to get it working in previous runs. I'll add the remaining state codes.

openfisca_us/variables/demographic/household.py
```diff
@@ -32,3 +33,104 @@ class is_household_head(Variable):
...
+    definition_period = ETERNITY
+
+
+class StateGroup(Enum):
+    CONTIGUOUS_US = "Contiguous US"
+    AK = "Alaska"
+    HI = "Hawaii"
+    GU = "Guam"
+    PR = "Puerto Rico"
+    # Omit other territories for now.
```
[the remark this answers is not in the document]
MaxGhenis: The CPS only has the 50 states and DC. If any new ones show up, the current logic will slot them into CONTIGUOUS_US. To avoid that, we'd have to make a list of all the states, which I guess we could do but ideally I think we'd use something like the us package to automate it.

openfisca_us/variables/demographic/household.py
```diff
@@ -32,3 +33,104 @@ class is_household_head(Variable):
...
+    default_value = StateGroup.CONTIGUOUS_US
+    entity = Household
+    label = u"State group"
+    definition_period = ETERNITY
+
+    def formula(household, period, parameters):
+        NON_CONTIGUOUS_STATES = ("AK", "HI", "GU", "PR", "VI")
+        state_code = household("state_code", period).decode_to_str()
+        return where(
+            np.isin(state_code, NON_CONTIGUOUS_STATES),
```
nikhilwoodruff, line 133: Very minor but I'd prefer to have an `isin` alias documented in `tools/general.py` rather than importing numpy.

PR #222, opened 2021-10-31
Child care age group

### LINKED ISSUE #202: Create variable for childcare age group in NY
Per https://view.officeapps.live.com/op/view.aspx?src=https%3A%2F%2Focfs.ny.gov%2Fmain%2Fpolicies%2Fexternal%2Focfs_2019%2FLCM%2F19-OCFS-LCM-23.docx&wdOrigin=BROWSELINK

* Infant: <1.5 (center-based) or under 2 years (home-based)
* Toddler: 1.5-2 (center-based) or 2 (home-based)
* Preschooler: 3-5
* School aged: 6-12

tolaouk:
fixes #202


openfisca_us/variables/demographic/person.py
```diff
@@ -43,3 +43,57 @@ def formula(person, period, parameters):
...
+    value_type = float
+    entity = Person
+    label = u"CCDF age"
+    definition_period = YEAR
+
+
+class CCDFCareLocation(Enum):
+    CENTER_BASED = "Center-based"
+    HOME_BASED = "Home-based"
+
```
nikhilwoodruff, line 58: @MaxGhenis @nmrodelo feel free to override or if we should make this a general pattern, but I wonder if for all enums we should have an "UNSPECIFIED"/"UNKNOWN" value?

openfisca_us/variables/demographic/person.py
```diff
@@ -43,3 +43,57 @@ def formula(person, period, parameters):
...
+    SA = "School age"
+
+
+class ccdf_age_group(Variable):
+    value_type = Enum
+    possible_values = CCDFAgeGroup
+    default_value = CCDFAgeGroup.I
+    entity = Person
+    label = u"CCDF age group"
+    definition_period = YEAR
```
nikhilwoodruff, line 82: Could you add the reference link? I know it's a word online document - is there an official hosted page somewhere we can link to?

openfisca_us/variables/demographic/person.py
```diff
@@ -43,3 +43,57 @@ def formula(person, period, parameters):
...
+
+    def formula(person, period, parameters):
+        ccdf_age = person("ccdf_age", period)
+        ccdf_care_location = person(
+            "ccdf_care_location", period
+        ).decode_to_str()
```
nikhilwoodruff, line 88: Ideally we'd want to compare Enums instead of comparing strings (it's slightly more efficient). For example:
```python
ccdf_care_location = person("ccdf_care_location", period)
care_locations = ccdf_care_location.possible_values # links to the Enum class
...
return select(
    [
        ((ccdf_age < 1.5) & (ccdf_care_location == care_locations.CENTER_BASED))
...
```

openfisca_us/variables/demographic/person.py
```diff
@@ -43,3 +43,57 @@ def formula(person, period, parameters):
...
+        return select(
+            [
+                (ccdf_age < 1.5 and ccdf_care_location == "CENTER_BASED")
```
nikhilwoodruff, line 91: Could you use `&` and `|` instead of `and` and `or`? I think the vectorisation requires it.

nikhilwoodruff (changes requested):
Thanks @tolaouk! Just left a few comments.

openfisca_us/variables/demographic/person.py
tolaouk: found pdf link for it :)
https://ocfs.ny.gov/main/policies/external/ocfs_2019/LCM/19-OCFS-LCM-23.pdf

openfisca_us/variables/demographic/person.py
MaxGhenis: What logic would we use in that case though?

openfisca_us/variables/demographic/person.py
```diff
@@ -43,3 +43,76 @@ def formula(person, period, parameters):
...
+    HOME_BASED = "Home-based"
+
+
+class ccdf_care_location(Variable):
+    value_type = Enum
+    possible_values = CCDFCareLocation
+    default_value = CCDFCareLocation.CENTER_BASED
+    entity = Person
+    label = u"CCDF care location"
+    definition_period = YEAR
```
MaxGhenis, line 66: Is this a function of `provider_type`?

tolaouk: Yeah actually they are
image

tolaouk: DCC_SACC -> center-based
FDC_GFDC -> home-based
LE_STD -> home-based
LE_ENH -> home-based

tolaouk: I'll need to change the formula then.

tolaouk: My current branch, child_care_age_group, doesn't have the enums for provider_type. How should I approach with this?

openfisca_us/variables/demographic/person.py
nikhilwoodruff: True, maybe it's better not to- it'd need an UNKNOWN path all the way through the computation tree.

openfisca_us/variables/demographic/person.py
```diff
@@ -17,7 +17,7 @@ class person_weight(Variable):
 
 
 class age(Variable):
-    value_type = int
+    value_type = float
```
MaxGhenis, line 20: @nikhilwoodruff fyi, since the rules are based on age more fine than int, e.g. 1.5 years

openfisca_us/variables/demographic/person.py
```diff
@@ -45,6 +45,50 @@ def formula(person, period, parameters):
...
+    definition_period = YEAR
+
+    def formula(person, period, parameters):
+        return (
+            person("provider_type_group", period) != ProviderTypeGroup.DCC_SACC
+        )
+
+
+class CCDFAgeGroup(Enum):
+    I = "Infant"
```
MaxGhenis, line 62: should we spell these out?

MaxGhenis (commented):
@nikhilwoodruff any tips on diagnosing the test failures?

openfisca_us/variables/demographic/person.py
```diff
@@ -45,6 +45,50 @@ def formula(person, period, parameters):
...
+    default_value = CCDFAgeGroup.I
+    entity = Person
+    label = u"CCDF age group"
+    definition_period = YEAR
+
+    reference = "https://ocfs.ny.gov/main/policies/external/ocfs_2019/LCM/19-OCFS-LCM-23.pdf"
+
+    def formula(person, period, parameters):
+        age = person("age", period)
+        home_based = person("is_ccdf_home_based", period, parameters)
```
nikhilwoodruff:
```suggestion
        home_based = person("is_ccdf_home_based", period)
```

This just needs to only have the variable name and period arguments (parameters get passed automatically by OpenFisca).

openfisca_us/variables/demographic/person.py
```diff
@@ -45,6 +45,50 @@ def formula(person, period, parameters):
...
+        return select(
+            [
+                (age < 1.5 & ~home_based) | (age < 2 & home_based),
```
nikhilwoodruff:
```suggestion
                ((age < 1.5) & ~home_based) | (age < 2 & home_based),
```

I think the `&` operator has precedence over `<`, so this was evaluated as `age < (1.5 & (~home_based))` (which breaks).

openfisca_us/variables/demographic/person.py
```diff
@@ -45,6 +45,50 @@ def formula(person, period, parameters):
...
+                (age < 2 & ~home_based) | (age < 3 & home_based),
```
nikhilwoodruff: Same here as above

nikhilwoodruff (commented):
Thanks both- just had a look at the test fails (issues were one operator precedence bug and one argument-related bug).

openfisca_us/variables/demographic/person.py
```diff
@@ -45,6 +45,55 @@ def formula(person, period, parameters):
...
+    INFANT = "Infant"
+    TODDLER = "Toddler"
+    PRESCHOOLER = "Preschooler"
+    SCHOOL_AGE = "School age"
+
+
+class ccdf_age_group(Variable):
+    value_type = Enum
+    possible_values = CCDFAgeGroup
+    default_value = CCDFAgeGroup.INFANT
```
MaxGhenis, line 71: @nikhilwoodruff should we have default values for variables that descend from other variables with their own default values? e.g. this is based on age (default 0) and home-based (default False), so it would default to infant anyway.

MaxGhenis (approved):
Good to merge after Nikhil clarifies this question from me. Thanks!

openfisca_us/variables/demographic/person.py
MaxGhenis: answer: yes

MaxGhenis:
@nikhilwoodruff I think you'll also need to update your approval

openfisca_us/variables/demographic/person.py
MaxGhenis: Let's actually deal with this question separately. Merging.

nikhilwoodruff (approved):

PR #224, opened 2021-11-05
Snap gross income

nmrodelo:
Add variable for snap_gross_income

### DIFF
--- a/openfisca_us/variables/finance/benefit/snap.py
+++ b/openfisca_us/variables/finance/benefit/snap.py
@@ -5,12 +5,9 @@ from openfisca_us.tools.general import *
 
 class snap_gross_income(Variable):
     value_type = float
-    entity = Person
+    entity = SPMUnit
     definition_period = YEAR
-    label = "Gross income for the person"
-
-    def formula(persons, period, parameters):
-        return persons("snap_gross_income", period)
+    label = "Gross income for the SPM Unit"
 
 
 class snap_earnings_deduction(Variable):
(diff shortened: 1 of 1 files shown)


openfisca_us/variables/finance/benefit/snap.py
```diff
@@ -3,6 +3,16 @@
 from openfisca_us.tools.general import *
 
 
+class snap_gross_income(Variable):
+    value_type = float
+    entity = Person
```
MaxGhenis, line 8: Should we calculate this at the person or spm unit?

openfisca_us/variables/finance/benefit/snap.py
```diff
@@ -3,6 +3,16 @@
...
+    definition_period = YEAR
+    label = "Gross income for the person"
+
+    def formula(persons, period, parameters):
+        return persons("snap_gross_income", period)
```
MaxGhenis, line 13: this recursion won't work. if it's not a formula based on other variables, you can omit the `formula` section.

openfisca_us/variables/finance/benefit/snap.py
```diff
@@ -3,6 +3,13 @@
 from openfisca_us.tools.general import *
 
 
+class snap_gross_income(Variable):
+    value_type = float
+    entity = SPMUnit
+    definition_period = YEAR
+    label = "Gross income for the SPM Unit"
```
MaxGhenis, line 10:
```suggestion
    label = "SPM unit's gross income for calculating SNAP eligibility"
```

MaxGhenis (approved):
Good to go after this I think!

Issue #233, opened 2021-11-06
Add entity encapsulation


MaxGhenis:
person -> tax_unit -> family -> spm_unit -> household

PR #237, opened 2021-11-06
Code health improvements

nikhilwoodruff:
Adds:
* A test for lower-case variable names
* Group encapsulation
* (also, a test for entity_variable_name names, but I've disabled for now due to the tax variables)

Applies autopep8, black and linecheck.


openfisca_us/tests/code_health/variable_names.py
```diff
@@ -0,0 +1,20 @@
...
+
+    system = CountryTaxBenefitSystem()
+    for instance in system.variables.values():
+        variable = type(instance)
+        assert (
+            variable.__name__.islower()
+        ), f"{variable.__name__} is not all lowercase"
+        if variable.entity.key in GROUP_ENTITY_NAMES:
+            pass
+            # assert variable.__name__[:len(variable.entity.key)] == variable.entity.key, f"{variable.__name__} does not begin with the specified group entity"
```
MaxGhenis, line 20: Still a draft?

openfisca_us/entities.py
```diff
@@ -32,6 +15,7 @@
             doc="A member of the household",
         ),
     ],
+    containing_entities=["spm_unit"],
```
MaxGhenis, line 18: I don't think the set of `containing_entities` arguments here consistently represents the hierarchy:

`person` -> `tax_unit` -> `family` -> `spm_unit` -> `household`

Should this include just the immediate predecessor in the hierarchy? SPMUnit having both family and household seems off.

MaxGhenis (commented):
Thanks @nikhilwoodruff this will make our lives much easier!

openfisca_us/tests/code_health/variable_names.py
nikhilwoodruff: Yeah, the taxcalc variables in particular. We just need to add `tax_unit` to each variable name, but not sure how that'll work with the non-dep sums.

openfisca_us/entities.py
nikhilwoodruff: I mean, I should know which is the correct encapsulation, but in https://github.com/PolicyEngine/openfisca-us/issues/233 `spm_unit` and `family` are switched.

MaxGhenis: Sorry - yes #233 was wrong (edited to fix). See https://www.census.gov/content/dam/Census/library/publications/2021/demo/p60-275.pdf#page=3

image

PR #251, opened 2021-11-13
ccdf market rate NY

### LINKED ISSUE #210: Add market rate for CCDF NY
From https://ocfs.ny.gov/main/reports/2019-Child-Care-Market-Rate-Survey.pdf#page=24

Keyed on county cluster, provider type group, duration of care, and child age.

tolaouk:
fixes #210

### DIFF
--- a/openfisca_us/parameters/benefit/ccdf/amount.yaml
+++ b/openfisca_us/parameters/benefit/ccdf/amount.yaml
@@ -1,0 +1,878 @@
+description: CCDF benefit rate by county cluster, provider type, duration of care, and child age
+
+1:
+  DCC_SACC:
+    WEEKLY:
+      INFANT:
+        2021-10-01: 364
+      TODDLER:
+        2021-10-01: 338
+      PRESCHOOLER:
+        2021-10-01: 310
+      SCHOOL_AGE:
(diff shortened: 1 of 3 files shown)


MaxGhenis (approved):

MaxGhenis:
Looks good! Can I merge this?

tolaouk:
> Looks good! Can I merge this?

Yes you can :)

nikhilwoodruff:
One thing before merging (and we should formalise this in policyengine.dev) - could we bump the version (even by a patch) with this? I'd also support starting a proper changelog like in openfisca-core, @MaxGhenis wdyt?

MaxGhenis:
I'd prefer to define our versioning strategy first. PolicyEngine needs consistent version bumps because of its continuous deployment and caching. Adopting that could be cleaner, but keeping track of versions across PRs ends up being a slightly-taxing way of tagging every PR as either a bug fix or new feature. An alternative could be bunchubg them here (though this would admittedly be arbitrary and we may eventually need to switch). Could we wait until a next PR?

nikhilwoodruff:
Ok, yeah. I also do agree on the information duplication. But this PR is definitely good to merge!

PR #254, opened 2021-11-16
Add AGI 'above the line' deductions

### LINKED ISSUE #53: Add Adj
Add the tax policy logic from the [Adj](https://github1s.com/pslmodels/tax-calculator/blob/HEAD/taxcalc/calcfunctions.py#L315) function in `tax-calculator`, including parameters, intermediate and output variables.

nikhilwoodruff:
This adds the logic and parameters from the [Adj](https://github1s.com/pslmodels/tax-calculator/blob/HEAD/taxcalc/calcfunctions.py#L315) function in `Tax-Calculator`. I've also added a unit test (there isn't one AFAICT in taxcalc. Necessary intermediate variables (the `filer_` variables) have also been added.

Uncertainties:
* Some of the haircut rates are constant - do we need them?
* Is the parameter organisation correct?

As for the naming, I'm continuing with the pattern of exempting tax-calculator variables from our naming conventions while we port the logic in order to avoid overcomplicating it.

Fixes #53


openfisca_us/parameters/tax/ald/misc/alimony/paid.yaml
```diff
@@ -0,0 +1,7 @@
+description: Adjustment for alimony-paid haircut
+metadata:
+  unit: currency-USD
```
MaxGhenis, line 3: This and other haircuts are proportions
```suggestion
  unit: /1
```

MaxGhenis (approved):
Slick `c02900` formula!

There are a couple trailing lines in the files, not sure if your utility catches that.

openfisca_us/variables/finance/tax/outputs.py
```diff
@@ -438,10 +438,37 @@ class c02900(Variable):
...
+    unit = USD
     documentation = (
         """Total of all 'above the line' income adjustments to get AGI"""
     )
 
+    def formula(tax_unit, period, parameters):
+        misc_haircuts = parameters(period).tax.ald.misc
+        other_deductions = add(tax_unit, period, "c03260", "care_deduction")
+        return (
+            sum(
```
MaxGhenis, line 451: Nit but I think it'd be clearer if this part was its own variable, like `other_deductions` so you just have to `return haircut_deductions + other_deductions`

nikhilwoodruff: OK I've split it up, are the variable names OK?

openfisca_us/parameters/tax/ald/misc/alimony/paid.yaml
nikhilwoodruff: Fixed

openfisca_us/variables/finance/tax/outputs.py
MaxGhenis: I was thinking Python variables. Right now it's kind of arbitrary which deductions allow haircuts, since some had full haircuts in the past (i.e. they didn't count as deductions against AGI) and some have always been fully deducted.

Rather than adding these variables, how about adding two new parameters for haircuts on `c03260` and `care_deduction`, to make it a single consistent variable?

nikhilwoodruff: OK, yeah that makes sense - how about now?

openfisca_us/variables/finance/tax/outputs.py
```diff
@@ -439,10 +439,35 @@ class c02900(Variable):
...
         """Total of all 'above the line' income adjustments to get AGI"""
     )
 
+    def formula(tax_unit, period, parameters):
+        misc_haircuts = parameters(period).tax.ald.misc
+        return sum(
+            [
+                (1 - haircut) * tax_unit(variable, period)
+                for haircut, variable in (
+                    (misc_haircuts.c03260, "c03260"),
```
MaxGhenis, line 454: Do `c03260` and `care_deduction` not apply the filer logic (i.e. not `filer_c03260`)?

openfisca_us/variables/finance/tax/outputs.py
```diff
@@ -439,10 +439,35 @@ class c02900(Variable):
...
+                    (misc_haircuts.care_deduction, "care_deduction"),
+                    (misc_haircuts.student_loan, "filer_e03210"),
+                    (misc_haircuts.early_withdrawal, "filer_e03400"),
```
MaxGhenis, line 457: Sorry to keep this PR going but could you align the haircut parameter name to the variable? e.g., `misc_haircuts.e03400` here. Would that then enable something simpler like:
```python
HAIRCUT_VARS = ["e03210", "e03400", ...]
return sum([(1 - misc_haircuts[variable]) * tax_unit(var, period) for variable in HAIRCUT_VARS])
```

openfisca_us/variables/finance/tax/outputs.py
nikhilwoodruff: They don't as they're already tax-unit level

openfisca_us/variables/finance/tax/outputs.py
nikhilwoodruff: OK, how does it look now?

MaxGhenis (approved):

openfisca_us/variables/finance/tax/outputs.py
```diff
@@ -439,10 +439,32 @@ class c02900(Variable):
...
     definition_period = YEAR
+    label = "'Above the line' AGI deductions"
+    unit = USD
     documentation = (
         """Total of all 'above the line' income adjustments to get AGI"""
     )
 
+    def formula(tax_unit, period, parameters):
+        misc_haircuts = parameters(period).tax.ald.misc.haircut
+        HAIRCUT_VARS = (
```
MaxGhenis, line 450: Do these have to be tuples?
```suggestion
        BASE_HAIRCUT_VARS = ["c03260", "care_deduction"]
        FILER_HAIRCUT_VARS = [
            "e03210",
            "e03400",
            "e03500",
            "e00800",
            "e03220",
            "e03230",
            "e03240",
            "e03290",
            "e03270",
            "e03150",
            "e03300",
        ]
        haircut_vars = BASE_HAIRCUT_VARS + ["filer_" + name for name in FILER_HAIRCUT_VARS]
```

PR #270, opened 2021-11-19
Update SNAP Params and Variables [ max() to max_() ]

### LINKED ISSUE #223: Handle negative net income values when calculating SNAP Variables

nmrodelo:
Fixes #223

### DIFF
--- a/openfisca_us/variables/finance/benefit/snap.py
+++ b/openfisca_us/variables/finance/benefit/snap.py
@@ -54,10 +54,13 @@ class snap_net_income_pre_shelter(Variable):
 
     def formula(spm_unit, period, parameters):
 
-        return (
-            spm_unit("gross_income", period)
-            - spm_unit("snap_standard_deduction", period)
-            - spm_unit("snap_earnings_deduction", period)
+        return max_(
+            (
+                spm_unit("gross_income", period)
+                - spm_unit("snap_standard_deduction", period)
+                - spm_unit("snap_earnings_deduction", period)
+            ),
+            0,
         )
 
 
@@ -102,8 +105,10 @@ class snap_net_income(Variable):
 
     def formula(spm_unit, period, parameters):
 
-        return spm_unit("snap_net_income_pre_shelter", period) - spm_unit(
-            "snap_shelter_deduction", period
+        return max_(
+            spm_unit("snap_net_income_pre_shelter", period)
+            - spm_unit("snap_shelter_deduction", period),
+            0,
         )
 
 
--- a/openfisca_us/variables/finance/benefit/snap.py
+++ b/openfisca_us/variables/finance/benefit/snap.py
@@ -54,14 +54,12 @@ class snap_net_income_pre_shelter(Variable):
 
     def formula(spm_unit, period, parameters):
 
-        return max_(
-            (
-                spm_unit("gross_income", period)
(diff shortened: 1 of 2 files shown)


openfisca_us/variables/finance/benefit/snap.py
```diff
@@ -54,10 +54,13 @@ class snap_net_income_pre_shelter(Variable):
 
     def formula(spm_unit, period, parameters):
 
-        return (
-            spm_unit("gross_income", period)
-            - spm_unit("snap_standard_deduction", period)
-            - spm_unit("snap_earnings_deduction", period)
+        return max_(
+            (
+                spm_unit("gross_income", period)
```
MaxGhenis, line 59: make this a new python variable, e.g. `res`

same with below

MaxGhenis (commented):
Could you add a test for this?

nmrodelo:
> Could you add a test for this?

For what exactly?

nikhilwoodruff:
> > Could you add a test for this?
> 
> For what exactly?

I think just something that fails with the old `max` statement, and that passes with the new `max_` fix (probably an input with more than one SPM unit will do this)

openfisca_us/variables/finance/benefit/snap.py
```diff
@@ -151,10 +153,13 @@ class snap(Variable):
 
     def formula(spm_unit, period, parameters):
         # TODO: Add gross and net income checks.
-        return spm_unit("snap_max_benefit", period) - spm_unit(
+
+        res = spm_unit("snap_max_benefit", period) - spm_unit(
```
MaxGhenis, line 157:
```suggestion
        return spm_unit("snap_max_benefit", period) - spm_unit(
```
don't need the `res` here

MaxGhenis (commented):
I don't think `max_`works in yaml. Might be `max`?

But what we were suggesting is a test that would have failed under the previous logic and succeeded under the current logic. The extra `max_` call changes the results, so we want to include a test that shows that the new result is what we want.

openfisca_us/variables/finance/benefit/snap.py
```diff
@@ -54,11 +54,12 @@ class snap_net_income_pre_shelter(Variable):
...
     def formula(spm_unit, period, parameters):
 
-        return (
-            spm_unit("gross_income", period)
-            - spm_unit("snap_standard_deduction", period)
-            - spm_unit("snap_earnings_deduction", period)
+        res = spm_unit("gross_income", period) - spm_unit(
+            "snap_standard_deduction", period
         )
+        -spm_unit("snap_earnings_deduction", period)
```
MaxGhenis, line 60: move to line above

openfisca_us/variables/finance/benefit/snap.py
```diff
@@ -102,9 +103,10 @@ class snap_net_income(Variable):
 
     def formula(spm_unit, period, parameters):
 
-        return spm_unit("snap_net_income_pre_shelter", period) - spm_unit(
-            "snap_shelter_deduction", period
-        )
+        res = spm_unit("snap_net_income_pre_shelter", period)
+        -spm_unit("snap_shelter_deduction", period)
```
MaxGhenis, line 107: move to line above

PR #279, opened 2021-11-20
Add SchXYZTax

### LINKED ISSUE #92: Add SchXYZTax

nikhilwoodruff:
Fixes #92


openfisca_us/variables/finance/tax/outputs.py
```diff
@@ -621,8 +621,29 @@ class c05200(Variable):
...
+        pass_through = parameters(period).tax.income.pass_through
+        e26270 = tax_unit("filer_e26270", period)
+        e00200 = tax_unit("filer_e00200", period)
+        pt_passive = pass_through.business_income.eligible_rate.passive * (
+            e00200
+            - e26270
+        )
+        e00900 = tax_unit("e00900", period)
+        pt_active_gross = e00900 + e26270
+        if (pt_active_gross > 0) and pass_through.wages_active_income:
```
MaxGhenis, line 641:
```suggestion
        if pass_through.wages_active_income:
```

openfisca_us/variables/finance/tax/outputs.py
```diff
@@ -621,8 +621,29 @@ class c05200(Variable):
...
+            pt_active_gross += e00200
```
MaxGhenis, line 642:
```suggestion
            pt_active_gross += where(pt_active_gross > 0, e00200, 0)
```

MaxGhenis (commented):
looks like still a draft but here's one fix

nikhilwoodruff:
Thanks for the review- actually I think this is SchXYZ.

nikhilwoodruff:
Definitely think after this adding a unit test (even just default = 0) for every tax PR is necessary even when taxcalc doesn't have one - a lot of bugs from previous additions by me.

openfisca_us/variables/finance/tax/outputs.py
nikhilwoodruff: Done, thanks

openfisca_us/variables/finance/tax/outputs.py
nikhilwoodruff: Added

nikhilwoodruff:
@MaxGhenis I've now removed the pass-through income reform parameters as these all seemed to be reform-only (1 or 0, for all time).

PR #283, opened 2021-11-20
Add Top Level ssi variables

Jason017:


openfisca_us/variables/finance/benefit/ssi.py
```diff
@@ -0,0 +1,32 @@
...
+    entity = Person
+    definition_period = YEAR
+    documentation = "Eligibility for Supplemental Security Income"
+
+
+class ssi_amount_if_eligible(Variable):
+    value_type = float
+    entity = Person
+    definition_period = YEAR
+    documentation = "Supplemental Security Income amount if someone is eligible"
```
MaxGhenis, line 32: newline

MaxGhenis (commented):
Let's also move this from `variables/finance/benefit/ssi.py` to `variables/finance/ssa/ssi.py` since it's in the Social Security Administration.

PR #281, opened 2021-11-20
Add templates

nikhilwoodruff:
Adds a pull request and issue template.

PR #286, opened 2021-11-20
Add core income definitions

nikhilwoodruff:
Somewhat ironically, the first PR filed after #281 doesn't fit in any of the options. This PR adds a few variables for our core income definitions, with the idea that when we're working on new programs that use income definitions, we should either use the core variables (mapped to the appropriate entity) or create program-specific income definitions. I've also added `unit` and `documentation` to the VS Code variable snippet.

### DIFF
--- a/.vscode/python.code-snippets
+++ b/.vscode/python.code-snippets
@@ -18,8 +18,10 @@
 		 		"class ${1:name}(Variable):",
 				"	value_type = ${2:float}",
 				"	entity = ${3:TaxUnit}",
-				"	label = u\"${4:Label}\"",
-				"	definition_period = ${5:YEAR}",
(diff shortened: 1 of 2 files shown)


MaxGhenis:
Are these for demonstration purposes? We won't be able to calculate taxes and (all) benefits at the person level.

nikhilwoodruff:
I was imagining we'd calculate at the appropriate entity and give to the head like in the UK?

nikhilwoodruff:
Per a discussion with @MaxGhenis we'll not give group-level benefits to individuals, and rather only aggregate upwards and not downwards.

nikhilwoodruff:
We decided not to include these, since they'd probably be misleading or unclear.

PR #301, opened 2021-11-21
Added formula for tanf_max_amount

yvvijay121:
added formula for tanf_max_amount, but the family_size function does not exist.


openfisca_us/variables/finance/benefit/tanf.py
```diff
@@ -38,6 +38,12 @@ class tanf_max_amount(Variable):
     documentation = "The maximum benefit amount a family could receive from Temporary Assistance for Needy Families given their state and family size."
     unit = "currency-USD"
 
+    def formula(spm_unit, period, parameters):
+        family_size = spm_unit("family_size", period)
```
nikhilwoodruff, line 42: Nice, could you add an `spm_unit_num_people` OpenFisca variable if that's what this is?

openfisca_us/variables/finance/benefit/tanf.py
```diff
@@ -38,6 +38,12 @@ class tanf_max_amount(Variable):
...
+        state = spm_unit("state", period)
```
nikhilwoodruff, line 43:
```suggestion
        state = spm_unit.household("state_code", period)
```

Though @MaxGhenis should we rename this variable to simply `state`?

nikhilwoodruff (commented):
Thanks @yvvijay121 , still a draft I know but here's a few comments in case they're useful.

openfisca_us/variables/finance/benefit/tanf.py
MaxGhenis: Should we do that or just use `spm_unit.nb_persons()`?

nikhilwoodruff: Good point, that'd work

yvvijay121:
Thank you for the comments! I made the changes that you suggested, and I will mark this pull request for review.

openfisca_us/parameters/hhs/tanf/max_amounts.yaml
```diff
@@ -0,0 +1,47 @@
...
+    2021-10-01: 2024
+  17:
+    2021-10-01: 2138
+  18:
+    2021-10-01: 2252
+metadata:
+  unit: currency-USD
+  reference: 
+  - title: "Illinois TANF Eligibility and Benefit Amount: WAG 25-03-05: Table of Monthly Allowances"
+    href: https://www.dhs.state.il.us/page.aspx?item=12673
```
MaxGhenis, line 43: Could you add a comment showing which page indicates that these values took effect on 2021-10-01?

MaxGhenis (commented):
Great, could you add a couple tests, e.g. modeled after the [school meal tests](https://github.com/PolicyEngine/openfisca-us/blob/master/openfisca_us/tests/policy/baseline/finance/benefit/school_meals.yaml)?

openfisca_us/variables/finance/benefit/tanf.py
```diff
@@ -38,6 +38,12 @@ class tanf_max_amount(Variable):
     documentation = "The maximum benefit amount a family could receive from Temporary Assistance for Needy Families given their state and family size."
     unit = "currency-USD"
 
+    def formula(spm_unit, period, parameters):
+        family_size = spm_unit.nb_persons()
```
MaxGhenis, line 42: This might fix the current test failure
```suggestion
        family_size = spm_unit.nb_persons().astype(str)
```

MaxGhenis (commented):
.

nikhilwoodruff:
@yvvijay121 looks like the test is failing, because the `state_code` variable, as an Enum, returns only the numeric index. We usually would fix this by adding `decode_to_str()`, but that looks like it's not working because of a [bug in OpenFisca-Core](https://github.com/openfisca/openfisca-core/issues/1085). Filed and issue and it shouldn't be too hard for me to find a fix, but in the meantime I'd use this as a quick fix:

```python
class state_code_str(Variable):
  value_type = str
  entity = Household
  definition_period = ETERNITY

  def formula(household, period):
    return household("state_code", period).decode_to_str()
```

and then replace your call to `state_code` with `state_code_str`.

openfisca_us/variables/hhs/tanf.py
```diff
@@ -59,4 +75,4 @@ class tanf_amount_if_eligible(Variable):
     def formula(spm_unit, period, parameters):
         max_amount = spm_unit("tanf_max_amount", period)
         countable_income = spm_unit("tanf_countable_income", period)
-        return max_(0, max_amount - countable_income)
+        return max_(0, max_amount - (0.25 * countable_income))
```
MaxGhenis, line 78: I thought the countable income was after the phase-out? If not, please make the 0.25 a parameter so we can have references and avoid hard-coding.

openfisca_us/variables/finance/benefit/tanf.py
MaxGhenis: We can discuss in https://github.com/PolicyEngine/openfisca-us/discussions/326

openfisca_us/variables/finance/benefit/tanf.py
MaxGhenis: Resolving but also pointing to #324

openfisca_us/variables/hhs/tanf.py
yvvijay121: Yes, you're right. I'll change this back to what it was.

nikhilwoodruff (approved):
Thanks @yvvijay121 - this looks good to merge to me.

PR #304, opened 2021-11-21
Add CapGains variables

nikhilwoodruff:
# New variables: `c01000`, `ymod`, `ymod`

- [x] Label field added
- [x] Unit field added
- [x] Default value field added if relevant
- [x] Unit test(s) added
- [x] Integration test(s) added if relevant

## What's changed

Added the parameters and logic from the taxcalc `CapGains` function. This also adds a missing `filer_` variable and also `invinc_agi_ec`, though this is reform-only (but to keep variable existence in line with taxcalc until we are confident we have the full logic).


openfisca_us/variables/finance/tax/outputs.py
```diff
@@ -1244,18 +1253,71 @@ class ymod(Variable):
...
+    documentation = "Variable that is included in AGI"
+
+    def formula(tax_unit, period, parameters):
+        direct_inputs = add(
+            tax_unit,
+            period,
+            *[
+                "filer_" + variable
+                for variable in (
+                    "e00200",
```
MaxGhenis, line 1286: Make this something like `DIRECT_INPUT_COLS` (same with investment below) to reduce nesting

MaxGhenis (approved):

openfisca_us/variables/finance/tax/outputs.py
nikhilwoodruff: Thanks, added

PR #309, opened 2021-11-22
Fixes

### LINKED ISSUE #308: EITC isn't phased out

nikhilwoodruff:
# Bug fix

- [x] Regression test added
- [x] Regression test passing

Fixes #308 - the three-way minimum failed to broadcast the singular `eitc.max` parameter to the same array as the variables, causing it to always be recognised as the minimum.


nikhilwoodruff:
My bad, forgot to check back into master. Now requires #307

PR #310, opened 2021-11-23
add ccdf eligibility variable

### LINKED ISSUE #311: add ccdf eligibility variable
CCDF eligibility depends on
> asset eligibility
> age eligibility
> initial income eligibility if not currently enrolled
> continuous income eligibility if currently enrolled

https://www.ocfs.ny.gov/programs/childcare/stateplan/assets/2022-plan/FFY2022-2024-CCDF-Plan.pdf

tolaouk:
fixes #311

### DIFF
--- a/openfisca_us/tests/policy/baseline/demographic/person/ccdf_eligibility.yaml
+++ b/openfisca_us/tests/policy/baseline/demographic/person/ccdf_eligibility.yaml
@@ -5,8 +5,10 @@
     is_ccdf_age_eligible: True
     is_enrolled_in_ccdf: False
     is_ccdf_initial_income_eligible: True
+    is_ccdf_continuous_income_eligible: True
   output:
     is_ccdf_eligible: True
+
 - name: Asset eligible, age eligible, not enrolled, initial income not eligible
   period: 2022
   input:
@@ -16,6 +18,7 @@
     is_ccdf_initial_income_eligible: False
   output:
     is_ccdf_eligible: False
+
 - name: Asset eligible, age eligible, enrolled, continuous income eligible
   period: 2022
   input:
@@ -25,6 +28,7 @@
     is_ccdf_continuous_income_eligible: True
   output:
     is_ccdf_eligible: True
+
 - name: Asset eligible, age eligible, enrolled, continuous income not eligible
   period: 2022
   input:
@@ -34,6 +38,7 @@
     is_ccdf_continuous_income_eligible: False
   output:
     is_ccdf_eligible: False
+
 - name: Asset not eligible, age eligible, not enrolled, initial income eligible
   period: 2022
   input:
@@ -43,6 +48,7 @@
     is_ccdf_initial_income_eligible: True
   output:
     is_ccdf_eligible: False
+
 - name: Asset eligible, age not eligible, not enrolled, initial income eligible
   period: 2022
   input:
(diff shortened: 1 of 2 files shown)


tolaouk:
Hi @MaxGhenis This is ready for review. I also added Age Eligibility since max age is 12 for this program.

openfisca_us/tests/policy/baseline/demographic/person/ccdf_eligibility.yaml
```diff
@@ -0,0 +1,54 @@
...
+- name: Asset not eligible, age eligible, not enrolled, initial income eligible
+  period: 2022
+  input:
+    is_ccdf_asset_eligible: False
+    is_ccdf_age_eligible: True
+    is_enrolled_in_ccdf: False
+    is_ccdf_initial_income_eligible: True
+  output:
+    is_ccdf_eligible: False
+- name: Asset eligible, age not eligible, not enrolled, initial income eligible
```
MaxGhenis, line 46: Could you add one more test for not-enrolled and only continuous eligible? This basically tests the false default for initial eligibility.

openfisca_us/tests/policy/baseline/demographic/person/ccdf_eligibility.yaml
```diff
@@ -0,0 +1,54 @@
+- name: Asset eligible, age eligible, not enrolled, initial income eligible
+  period: 2022
+  input:
+    is_ccdf_asset_eligible: True
+    is_ccdf_age_eligible: True
+    is_enrolled_in_ccdf: False
+    is_ccdf_initial_income_eligible: True
+  output:
+    is_ccdf_eligible: True
+- name: Asset eligible, age eligible, not enrolled, initial income not eligible
```
MaxGhenis, line 10: Could you add a line between tests? I think it could improve readability.

MaxGhenis (approved):
LGTM, just adding @nikhilwoodruff to take a look too.

openfisca_us/tests/policy/baseline/demographic/person/ccdf_eligibility.yaml
tolaouk: Great catch. I was thinking just New York. Other states can have different criteria between initial and continuous.

openfisca_us/tests/policy/baseline/demographic/person/ccdf_eligibility.yaml
tolaouk: I decided to have both values for each test since users would input the income, we can determine both initial and continuous eligibility from that. What do you think?

nikhilwoodruff (approved):
Thanks @tolaouk - all looks good! I am slightly curious about the best unit for the income eligibility (given this is person-level, like the age criteria, but as I understand it, no person will have both their age and income looked at? i.e. children won't have their income assessed, and adults won't have their age assessed), but these changes are good to merge and we can figure that out later.

tolaouk:
> Thanks @tolaouk - all looks good! I am slightly curious about the best unit for the income eligibility (given this is person-level, like the age criteria, but as I understand it, no person will have both their age and income looked at? i.e. children won't have their income assessed, and adults won't have their age assessed), but these changes are good to merge and we can figure that out later.

Yes you're right. Income eligibility is probably more fitting with household, since children in same household would have the same income eligibility status. I guess we'll see how it goes when I work on those eligibility variables.

MaxGhenis:
Interesting, this should be the first program where we have eligibility flags across multiple entities for the same program!

PR #313, opened 2021-11-23
Reorganize into agencies and calculate SPM unit poverty ratios

### LINKED ISSUE #172: Rename current SPM poverty threshold variables to include SPM
We need to distinguish from federal poverty guideline

https://github.com/PolicyEngine/openfisca-us/blob/master/openfisca_us/variables/finance/income/spm_unit.py

### LINKED ISSUE #276: Layer benefit parameters by the agency that administers them (USDA, HHS, HUD)
Benefit programs often share features with others in the agency that administers them. Examples:
* Household size adjustments to state median income computations differ between HHS (#263) and HUD (#264)
* USDA programs share definitions of elderly and disabled

I'd suggest we add a layer to `parameters/benefit` with folders for the agency. This would move the following folders/files under `parameters/`:
* `benefit/ccdf` -> `benefit/hhs/ccdf`
* `benefit/school_meals` -> `benefit/usda/school_meals`
* `benefit/snap` -> `benefit/usda/snap`
* `poverty/fpg.yaml` -> `benefit/hhs/fpg.yaml`

*Note that `benefit/unemployment_insurance/exemption` should be moved into `tax` (#152)*

A more extreme version of this could be removing `benefit/` and organizing `parameters/` by agency, replacing `tax` with `irs`, or perhaps even `treasury/irs` if we want to mirror the structure of the federal government.

Similarly, we may want a `parameters/states` folder for state-level programs like California Alternative Rates for Energy #259.

### LINKED ISSUE #314: Create enum variable for school meal tier
Free/reduced/full

MaxGhenis:
* Fixes #276 
* Fixes #172 
* Fixes #314


MaxGhenis:
@nikhilwoodruff do you know what these errors mean? I'm not sure why they'd be triggered by these file renamings, maybe some tests weren't running in their previous location?

nikhilwoodruff:
OK, looking at the SNAP test fails - @nmrodelo do you know whether these are correct?

nmrodelo:
> OK, looking at the SNAP test fails - @nmrodelo do you know whether these are correct?

@nikhilwoodruff is there only 1 of the SNAP tests failing? I don't understand from the output if only that test is failing or if there are more tests after this one that simply aren't bein run due to this error

nikhilwoodruff:
> > OK, looking at the SNAP test fails - @nmrodelo do you know whether these are correct?
> 
> @nikhilwoodruff is there only 1 of the SNAP tests failing? I don't understand from the output if only that test is failing or if there are more tests after this one that simply aren't bein run due to this error

There are more I think that aren't being run, it just fails at the first I think.

MaxGhenis:
Merging so others can adopt the new organization framework. @nmrodelo I deleted several failing SNAP tests, but filed https://github.com/PolicyEngine/openfisca-us/issues/323 with the relevant commits to reinstate them.

PR #332, opened 2021-11-28
Add automated test for each variable

nikhilwoodruff:
## Automated variable tests

This PR adds a `pytest` module adding a parametrized test for each variable in the system - not checking the value, but just that each variable doesn't error when calculating for a situation with one person.


MaxGhenis (approved):

nikhilwoodruff:
Added an exemption for three TANF variables, because the default state is California and we don't yet have parameters for it.

PR #342, opened 2021-11-28
Add GainsTax

nikhilwoodruff:
Adds the functionality and variables of [`GainsTax`](https://github1s.com/pslmodels/tax-calculator/blob/HEAD/taxcalc/calcfunctions.py#L1568). This required converting a multi-output formula into (the minimum number of) single-output variables.


nikhilwoodruff:
@MaxGhenis seems like at some point between the tests of the AMT PR and the AMT PR merge, something changed in the coverage action. Fixed by adding the `-i` option as per [S.O.](https://stackoverflow.com/questions/2386975/no-source-for-code-message-in-coverage-py)

openfisca_us/tests/policy/baseline/calcfunctions/gainstax.yaml
```diff
@@ -0,0 +1,261 @@
+- name: GainsTax unit test 0
```
MaxGhenis, line 1: Is this generated from the script with true CPS tax units? If so could you add a comment indicating as such?

It may also be helpful to include the CPS record ID in the test name or a comment.

openfisca_us/tests/policy/baseline/calcfunctions/gainstax.yaml
```diff
@@ -0,0 +1,261 @@
...
+    e26270: 0
+    e02000: 0
+  output:
+    dwks10: 3.0
+    dwks13: 3.0
+    dwks14: 2779.0
+    dwks19: 2779.0
+    c05700: 0.0
+    taxbc: 277.90000000000003
+
```
MaxGhenis, line 261:
```suggestion
```

openfisca_us/variables/irs/intermediate.py
```diff
@@ -211,6 +211,29 @@ def formula(tax_unit, period, parameters):
...
+
+    def formula(tax_unit, period, parameters):
+        INCOME_SOURCES = [
+            "c01000",
+            "c23650",
+            "filer_p23250",
+            "filer_e01100",
+            "filer_e00650",
+        ]
+        return np.any(
```
MaxGhenis, line 229: Unless there are negatives this could simplify. If it does need to capture negatives could you add a comment?
```suggestion
        return add(tax_unit, period, INCOME_SOURCES) > 0
```

MaxGhenis (approved):

openfisca_us/variables/irs/intermediate.py
```diff
@@ -219,6 +219,7 @@ class hasqdivltcg(Variable):
     definition_period = YEAR
 
     def formula(tax_unit, period, parameters):
+        # Negatives cannot offset other income sources
```
MaxGhenis, line 222: Could you add a unit test for this?

PR #347, opened 2021-11-29
Add metadata for PolicyEngine US

nikhilwoodruff:
These are just a few fixes for a functional household page on PolicyEngine.
* Added a label, unit and documentation field for `school_meal_subsidy`
* Fixed the MARS formula (for now, just calculating SINGLE/JOINT status - we can add more logic later).

### DIFF
--- a/openfisca_us/variables/irs/inputs.py
+++ b/openfisca_us/variables/irs/inputs.py
@@ -42,10 +42,11 @@ class mars(Variable):
     label = "MARS Status for the tax unit"
 
     def formula(tax_unit, period, parameters):
-        return tax_unit.max(
+        has_spouse = tax_unit.max(
             tax_unit.members("is_tax_unit_spouse", period)
-            * tax_unit.members("age", period)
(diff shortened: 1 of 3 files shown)


openfisca_us/variables/irs/inputs.py
```diff
@@ -42,10 +42,11 @@ class mars(Variable):
     label = "MARS Status for the tax unit"
```
MaxGhenis, line 42:
```suggestion
    label = "Marital status for the tax unit"
```

openfisca_us/variables/usda/school_meals.py
```diff
@@ -55,7 +55,9 @@ class school_meal_subsidy(Variable):
     value_type = float
     entity = SPMUnit
     definition_period = YEAR
-    documentation = ""
+    label = "School meal subsidy"
+    unit = "currency-GBP"
```
MaxGhenis, line 59:
```suggestion
    unit = "currency-USD"
```

MaxGhenis (approved):

PR #358, opened 2021-11-29
SNAP Homeless deduction

### LINKED ISSUE #317: SNAP Homeless Shelter Deduction
Per my exploration of the quality control data, 4.92% percent of individuals participating in SNAP are homeless, and therefore likely qualify and take the homeless deduction. Interesting to note that the takeup for this benefit/frequency of homeless individuals in SNAP is about as high (or low, depending on how you view it) as individuals who claim the medical expense deduction.

nmrodelo:
# New variable

- [x] Label field added
- [x] Unit field added
- [x] Unit test(s) added

fixes #317 
adds homeless shelter deduction variable, tests, and params

### DIFF
--- a/openfisca_us/parameters/usda/snap/homeless_shelter_deduction.yaml
+++ b/openfisca_us/parameters/usda/snap/homeless_shelter_deduction.yaml
@@ -1,12 +1,12 @@
 description: SNAP Homeless Shelter Deduction Parameters
 
-deduction:
+amount:
   2021-01-01: 143
   metadata:
     unit: currency-USD
 
 eligibility:
-  all_persons_homeless:
+  has_all_persons_homeless_for_snap:
     2021-01-01: True
   metadata:
     unit: bool
--- a/openfisca_us/parameters/usda/snap/homeless_shelter_deduction.yaml
+++ b/openfisca_us/parameters/usda/snap/homeless_shelter_deduction.yaml
@@ -5,17 +5,6 @@ values:
   metadata:
     unit: currency-USD
 
-eligibility:
-  has_all_persons_homeless_for_snap:
-    2021-01-01: True
-  metadata:
-    unit: bool
-
-#NOTE:
-#A State agency may provide a standard homeless shelter
-#deduction of $143 a month to households in which all members 
-#are homeless individuals but are not receiving free shelter throughout the month. 
-
 metadata:
   reference:
   - title: '7 CFR § 273.9 - Income and deductions.'
--- a/openfisca_us/parameters/usda/snap/homeless_shelter_deduction.yaml
+++ b/openfisca_us/parameters/usda/snap/homeless_shelter_deduction.yaml
@@ -1,6 +1,6 @@
 description: SNAP Homeless Shelter Deduction Parameters
 
-amount:
+values:
   2021-01-01: 143
   metadata:
     unit: currency-USD
--- a/openfisca_us/parameters/usda/snap/homeless_shelter_deduction.yaml
+++ b/openfisca_us/parameters/usda/snap/homeless_shelter_deduction.yaml
@@ -1,7 +1,7 @@
 description: SNAP Homeless Shelter Deduction Parameters
 
 values:
-  2021-01-01: 143 * 12
+  2021-01-01: 143
 
 metadata:
   unit: currency-USD
--- a/openfisca_us/parameters/usda/snap/homeless_shelter_deduction.yaml
+++ b/openfisca_us/parameters/usda/snap/homeless_shelter_deduction.yaml
@@ -1,7 +1,7 @@
 description: SNAP Homeless Shelter Deduction Parameters
 
 values:
-  2021-01-01: 143
+  2021-01-01: 143 * 12
 
 metadata:
   unit: currency-USD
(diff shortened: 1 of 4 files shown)


openfisca_us/parameters/usda/snap/homeless_shelter_deduction.yaml
```diff
@@ -0,0 +1,22 @@
+description: SNAP Homeless Shelter Deduction Parameters
+
+deduction:
```
MaxGhenis, line 3:
```suggestion
values:
```

openfisca_us/parameters/usda/snap/homeless_shelter_deduction.yaml
```diff
@@ -0,0 +1,22 @@
...
+  2021-01-01: 143
+  metadata:
+    unit: currency-USD
+
+eligibility:
```
MaxGhenis, line 8: remove this and make it part of the logic (homelessness is household-level)

openfisca_us/parameters/usda/snap/homeless_shelter_deduction.yaml
```diff
@@ -0,0 +1,22 @@
+description: SNAP Homeless Shelter Deduction Parameters
+
+deduction:
+  2021-01-01: 143
+  metadata:
```
MaxGhenis, line 5: move with other metadata and add `period: week`

openfisca_us/variables/usda/snap.py
```diff
@@ -175,3 +175,16 @@ class is_disabled_or_elderly_for_snap(Variable):
     entity = Person
     definition_period = YEAR
     documentation = "Indicates that a person is defined as disabled or elderly based on the USDA definition"
+
+
+class is_homeless_for_snap(Variable):
```
MaxGhenis, line 180: Let's make this a household-level `is_homeless` variable in https://github.com/PolicyEngine/openfisca-us/blob/master/openfisca_us/variables/demographic/household.py

We're a ways out from needing to model different homelessness definitions across agencies.

This variable doesn't need a formula.

openfisca_us/parameters/usda/snap/homeless_shelter_deduction.yaml
nmrodelo: Homelessness is defined at the person level per the law:
"A State agency may provide a standard homeless shelter deduction of $143 a month to households in which all members are homeless individuals but are not receiving free shelter throughout the month."
https://www.law.cornell.edu/cfr/text/7/273.9

I was under the impression it would be household level, but it seems the law included this pretty explicitly. What do you think?

MaxGhenis: Let's simplify to household-level for now, can always make more complicated later. But I think the microdata would show homelessness at the household level, and it's reasonable to ask users at the household level for the household calculator.

openfisca_us/parameters/usda/snap/homeless_shelter_deduction.yaml
nmrodelo: week? I don't follow, is there something I'm missing in the law?

nikhilwoodruff: Yeah it's monthly right? In that case, `period: month`

openfisca_us/variables/usda/snap.py
```diff
@@ -175,3 +175,19 @@ class is_disabled_or_elderly_for_snap(Variable):
...
+class snap_homeless_shelter_deduction(Variable):
+    value_type = float
+    entity = SPMUnit
+    definition_period = YEAR
+    label = "Homeless shelter deduction"
+    documentation = "https://www.law.cornell.edu/cfr/text/7/273.9"
+
+    def formula(household, period, parameters):
+
+        return where(
```
MaxGhenis, line 189:
```suggestion
        is_homeless = spm_unit.household("is_homeless", period)
        return is_homeless * parameters(period).usda.snap.homeless_shelter_deduction
```

openfisca_us/tests/policy/baseline/usda/snap/homeless_shelter_deduction.yaml
```diff
@@ -0,0 +1,19 @@
+- name: Family of 4, income of 24k/year (Standard Test), homeless
```
MaxGhenis, line 1: update test names

openfisca_us/variables/usda/snap.py
```diff
@@ -175,3 +175,19 @@ class is_disabled_or_elderly_for_snap(Variable):
...
+
+
+class snap_homeless_shelter_deduction(Variable):
+    value_type = float
+    entity = SPMUnit
+    definition_period = YEAR
+    label = "Homeless shelter deduction"
+    documentation = "https://www.law.cornell.edu/cfr/text/7/273.9"
+
+    def formula(household, period, parameters):
```
MaxGhenis, line 187:
```suggestion
    def formula(spm_unit, period, parameters):
```

nikhilwoodruff (approved):
Looks great to merge after these suggestions - thanks @nmrodelo!

Issue #325, opened 2021-11-27
Backdate CCDF age eligibility to 1991


MaxGhenis:
I looked around for CCDF legislation, and found that it was created in the Child Care and Development Block Grant Act of 1990, which became part of the [Omnibus Budget Reconciliation Act of 1990](https://www.congress.gov/bill/101st-congress/senate-bill/3209/text). This predated sectioned PDF legislation, but you can see the relevant section by searching for `who is less than 13`.

The [Child Care and Development Block Grant Act of 2014](https://www.congress.gov/bill/113th-congress/senate-bill/1086/text) reauthorized the program, and kept the same phrase.

PR #360, opened 2021-11-30
Update CCDF Eligibility Formula

tolaouk:
Update using federal rules.

Per sources from #325 
https://www.congress.gov/bill/113th-congress/senate-bill/1086/text


openfisca_us/variables/hhs/ccdf.py
```diff
@@ -67,18 +67,32 @@ class is_enrolled_in_ccdf(Variable):
...
 
 
+class is_ccdf_smi_eligible(Variable):
+    value_type = bool
+    entity = SPMUnit
+    definition_period = YEAR
+    label = u"SMI eligibility for CCDF"
+
+
+class is_ccdf_reason_for_care_eligible(Variable):
```
MaxGhenis, line 89: what does this mean?

tolaouk: Sorry forgot to mention that. It's a requirement that the kid is living with parents who are working or in job training or in educational program, or the kid is receiving or needs protective services. I used 'reason for care' because New York plan uses that term.

image

PR #363, opened 2021-12-02
move ccdf variables from demographic/person.py to hhs/ccdf.py

### LINKED ISSUE #327: Move CCDF variables from demographic/person.py into hhs/ccdf.py
https://github.com/PolicyEngine/openfisca-us/blob/master/openfisca_us/variables/demographic/person.py

tolaouk:
fixes #327


MaxGhenis (commented):
Could you move these into `variables/expense/person.py` instead? Let's reserve program-level scripts for variables that perform logic related to the program.
* provider_type_group
* childcare_hours_per_week
* childcare_hours_per_day
* childcare_days_per_week

PR #382, opened 2021-12-09
Add SNAP references, metadata and reorganise

nikhilwoodruff:
This adds references to the United States Code, Title 7 for the SNAP federal calculations. Since that legislation also references other parts of the Code providing State-level discretion, it seems like we should cite it for all components that include a federal aspect.

@nmrodelo sorry if anything here treads on your other PR/@MaxGhenis also FYI.


openfisca_us/variables/usda/snap/income/deductions/deductions.py
```diff
@@ -0,0 +1,26 @@
...
+    value_type = float
+    entity = SPMUnit
+    label = "SNAP income deductions"
+    unit = "currency-USD"
+    documentation = "Deductions made from gross income for SNAP benefits"
+    definition_period = YEAR
+    reference = "United States Code, Title 7, Section 2014(e)"
+
+    def formula(spm_unit, period, parameters):
+        return sum(
```
MaxGhenis, line 14: Would this work?
```python
DEDUCTIONS = ["standard", "earnings", "dependent_care", "child_support", "medical_expense", "shelter"]
return add(spm_unit, period, ["snap_" + i + "_deduction" for i in DEDUCTIONS])
```

openfisca_us/variables/usda/snap/income/deductions/deductions.py
```diff
@@ -0,0 +1,26 @@
+from openfisca_us.model_api import *
+
+
+class snap_deductions(Variable):
```
MaxGhenis, line 4: Could we add a unit test for this?

openfisca_us/variables/usda/snap/income/deductions/shelter_expense.py
```diff
@@ -0,0 +1,81 @@
...
+
+class snap_shelter_deduction(Variable):
+    value_type = float
+    entity = SPMUnit
+    definition_period = YEAR
+    documentation = (
+        "Excess shelter deduction for calculating SNAP benefit amount"
+    )
+    label = "SNAP shelter deduction"
+    reference = ("United States Code, Title 7, Section 2014(d)(6)",)
```
MaxGhenis, line 12: Is this right? I'm seeing that it points to child support: https://www.law.cornell.edu/uscode/text/7/2014#d_6

Also pointed to elsewhere in this PR.

MaxGhenis, line 12: Can we point to href/title pairs like in parameters? If not I think I'd prefer URLs.

openfisca_us/variables/usda/snap/income/deductions/shelter_expense.py
```diff
@@ -0,0 +1,81 @@
...
+
+        has_elderly_disabled = spm_unit("has_elderly_disabled", period)
+        # Cap for all but elderly/disabled people.
+        non_homeless_shelter_deduction = where(
+            has_elderly_disabled, uncapped_ded, min_(uncapped_ded, ded_cap)
+        )
+        homeless_shelter_deduction = spm_unit(
+            "snap_homeless_shelter_deduction", period
+        )
+        return homeless_shelter_deduction + non_homeless_shelter_deduction
```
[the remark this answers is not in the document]
nikhilwoodruff: Yep, you're right: 
image

openfisca_us/variables/usda/snap/income/deductions/shelter_expense.py
nikhilwoodruff: Unfortunately not - blocked by Core

nikhilwoodruff: I see your point, but I don't think we can link to specific lines with as much precision with URLS?

nikhilwoodruff: Good catch, it was (e) rather than (d)

MaxGhenis: We can! https://www.law.cornell.edu/uscode/text/7/2014#d_6

openfisca_us/variables/usda/snap/income/deductions/deductions.py
nikhilwoodruff: How about a list parameter?

openfisca_us/variables/usda/snap/income/deductions/deductions.py
nikhilwoodruff: Added

openfisca_us/variables/usda/snap/income/deductions/shelter_expense.py
nikhilwoodruff: Brain fail, yeah let's use those

openfisca_us/variables/usda/snap/income/deductions/deductions.py
nikhilwoodruff: In general seems like we should use list parameters instead of list constants

openfisca_us/variables/usda/snap/income/deductions/shelter_expense.py
nikhilwoodruff: Updated

nikhilwoodruff:
Here's how this is looking on PolicyEngine: 
image

PR #406, opened 2021-12-25
SNAP gross and income limits, and unit+integration tests

nmrodelo:
Unit tests for all existing SNAP variables


MaxGhenis:
This is failing because it adds `usda/snap.py`, which duplicates `usda/snap/snap.py`. Try deleting that file from the PR.

openfisca_us/variables/usda/snap/income/deductions/shelter_expense.py
```diff
@@ -163,10 +164,5 @@ def formula(spm_unit, period, parameters):
                 allowance_type == SNAPUttilityAllowanceType.TUA,
                 True,
             ],
-            [
-                utility.sua[state],
-                utility.lua[state],
-                utility.tua[state],
-                0,
-            ],
+            [utility.sua[state], utility.lua[state], utility.tua[state], 0,],
```
MaxGhenis, line 167: Should fix the lint error
```suggestion
            [utility.sua[state], utility.lua[state], utility.tua[state], 0],
```

openfisca_us/tests/policy/baseline/usda/snap/integration.yaml
```diff
@@ -13,18 +14,18 @@
         age: 15
     spm_units:
       spm_unit:
-        snap_gross_income: 2000 * 12
         members: [person_1, person_2, person_3, person_4]
+        is_snap_eligible: true
```
MaxGhenis, line 18: Without this, `snap` was getting set to zero. The $2,000/mo income is below the gross income limit of $2,839 (130% of FPL for a family of four) and the net income limit of $2,184 (100% of FPL) so I haven't figured out why this was necessary.

MaxGhenis (commented):
Thanks @nmrodelo! I removed and modified a couple tests to get everything to pass. It'd still be good to have unit tests for the variables here, and we may want to create new separate variables for the gross and net income tests, but I think it's good to merge as something to build on in future PRs if you agree @nikhilwoodruff.

openfisca_us/tests/policy/baseline/usda/snap/integration.yaml
nikhilwoodruff: Just took a look - @nmrodelo I think this was because the net/gross income limit percentages were getting multiplied by 12, rather than by the FPL.

openfisca_us/tests/policy/baseline/usda/snap/is_snap_eligible.yaml
```diff
@@ -0,0 +1,49 @@
...
+      person_4:
+        age: 15
+    spm_units:
+      spm_unit:
+        members: [person_1, person_2, person_3, person_4]
+        snap_gross_income: 20000 * 12
+        snap_net_income: 100 * 12
+    households:
+      household:
+        members: [person_1, person_2, person_3, person_4]
```
nikhilwoodruff, line 21: I just needed to add the `members` argument here, to ensure they're all under the same household (and therefore FPL).

nikhilwoodruff (approved):
Looks great - thanks @nmrodelo!

PR #410, opened 2021-12-26
Add semver actions

nikhilwoodruff:


nikhilwoodruff:
The code coverage action for the project failed because removing a redundant (but hit by tests) line of code increased the relative size of the untested code lines. I think we should make `patch` (test that the changes are at least as tested as the rest of the repo) mandatory but `project` (test that the changes don't make the repo as a whole less relatively tested) optional, because removing lines of code can decrease percentage coverage despite e.g. this case demonstrating that doesn't always mean the repo is less tested.

PR #421, opened 2021-12-29
Add education tax credits

nikhilwoodruff:
This implements the American Opportunity Credit and the Lifetime Learning Credit (from the [law](https://www.law.cornell.edu/uscode/text/26/25A#i), rather than taxcalc). This should fix both #103 and #105.

[tax-calc](https://github1s.com/pslmodels/tax-calculator/blob/HEAD/taxcalc/calcfunctions.py#L2599-L2600)

- [x] Add parameters
- [x] Add logic
- [x] Add tests

### DIFF
--- a/openfisca_us/parameters/irs/credits/education/phaseout.yaml
+++ b/openfisca_us/parameters/irs/credits/education/phaseout.yaml
@@ -12,10 +12,16 @@ start:
       unit: currency-USD
 length:
   description: Length of the phaseout for both the American Opportunity Credit and Lifetime Learning Credit on AGI (when excess income totals this amount, the credit is reduced to zero)
-  values:
-    2010-01-01: 10_000
-  metadata:
-    unit: currency-USD
+  single:
+    values:
+      2010-01-01: 10_000
+    metadata:
+      unit: currency-USD
+  joint:
+    values:
+      2010-01-01: 20_000
+    metadata:
+      unit: currency-USD
 metadata:
   reference:
     - title: 26 U.S. Code § 25A(d)(1)
--- a/openfisca_us/parameters/irs/credits/education/phaseout.yaml
+++ b/openfisca_us/parameters/irs/credits/education/phaseout.yaml
@@ -1,0 +1,22 @@
+start:
+  description: Phaseout start for both the American Opportunity Credit and Lifetime Learning Credit on AGI
+  single:
+    values:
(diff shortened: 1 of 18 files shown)


openfisca_us/variables/irs/credits/education/american_opportunity_credit.py
```diff
@@ -0,0 +1,82 @@
...
+    label = "American Opportunity Credit"
+    unit = "currency-USD"
+    documentation = "Total value of the American Opportunity Credit"
+    definition_period = YEAR
+    reference = "https://www.law.cornell.edu/uscode/text/26/25A#b"
+
+    def formula(tax_unit, period, parameters):
+        education = parameters(period).irs.credits.education
+        aoc = education.american_opportunity_credit
+        is_eligible = tax_unit.members(
```
MaxGhenis, line 16: Also in Lifetime Learning Credit section since used twice
```suggestion
        person = tax_unit.members
        is_eligible = person(
```

openfisca_us/parameters/irs/credits/education/phaseout.yaml
```diff
@@ -0,0 +1,22 @@
...
+      unit: currency-USD
+  joint:
+    values:
+      2010-01-01: 160_000
+    metadata:
+      unit: currency-USD
+length:
+  description: Length of the phaseout for both the American Opportunity Credit and Lifetime Learning Credit on AGI (when excess income totals this amount, the credit is reduced to zero)
+  values:
+    2010-01-01: 10_000
```
MaxGhenis, line 16: 20k for joint filers

openfisca_us/tools/general.py
```diff
@@ -35,3 +35,13 @@ def sum_contained_tax_units(var, population, period):
 infinity = np.inf
 select = np.select
 where = np.where
+
+
+def variable_alias(name: str, variable_cls: type) -> type:
+    """
+    Copy a variable class and return a new class.
+    """
+    new_variable_cls = type(
```
MaxGhenis, line 44:
```suggestion
    return type(
```

nikhilwoodruff:
To clarify on the changelog in ignored diffs: when the changelog is ignored in the functional changes check, the version action always fails because it thinks the changelog hasn't been updated. We should separate out the check for the changelog update from the check for functional changes.

PR #444, opened 2022-01-05
Use USDA elderly and disabled definitions in SNAP calculations.

MaxGhenis:
This replaces prior SNAP variables for elderly/disabled with `has_usda_elderly_disabled`, which is a formula based on `is_usda_elderly` and `is_usda_disabled`. This hooks up the SNAP rules to our formulas for USDA elderly and disabled status.

Also sets bools to lowercase in yaml files.


openfisca_us/tests/policy/baseline/usda/snap/expected_contribution.yaml
```diff
@@ -19,7 +19,7 @@
         snap_net_income_pre_shelter: (2000 - (2000 * 0.2) - 181) * 12
         snap_shelter_deduction: 586 * 12
         housing_cost: 1700 * 12
-        has_elderly_disabled: False
+        has_usda_elderly_disabled: False
```
MaxGhenis, line 22: change False to false

PR #450, opened 2022-01-08
add ccdf family copay formula

tolaouk:


MaxGhenis:
I made a couple updates to this PR following my #451

openfisca_us/variables/demographic/geographic/county_str.py
```diff
@@ -0,0 +1,12 @@
+from openfisca_us.model_api import *
+
+
+class county_str(Variable):
+    value_type = str
+    entity = Household
+    label = "County (string)"
+    documentation = "County variable, stored as a string"
+    definition_period = YEAR
```
MaxGhenis, line 9: I think this will fix the test issue. Otherwise try changing `ETERNITY` to `2022` in the test.
```suggestion
    definition_period = ETERNITY
```

openfisca_us/variables/assets/spm_unit_assets.py
```diff
@@ -4,6 +4,6 @@
 class spm_unit_assets(Variable):
     value_type = float
     entity = SPMUnit
-    label = "SPM unit assets"
+    label = u"SPM unit assets"
```
nikhilwoodruff, line 7: Very minor, but the `u` isn't necessary here (and other places).

openfisca_us/variables/demographic/geographic/county_str.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class county_str(Variable):
+    value_type = str
+    entity = Household
+    label = "County (string)"
+    documentation = "County variable, stored as a string"
+    definition_period = ETERNITY
+
+    def formula(household, period, parameters):
```
nikhilwoodruff, line 11: Technically we don't actually need `formula` to be a function of `parameters` - could just be `def formula(household, period):`. But we're not consistent on this - @MaxGhenis do you think we should start doing this?

nikhilwoodruff (approved):
Looks nice, thanks @tolaouk . I think the metadata additions of `u` are unnecessary, but that's a minor point.

openfisca_us/variables/demographic/geographic/county_str.py
MaxGhenis: Filed https://github.com/PolicyEngine/openfisca-us/issues/467 to decide and document. Let's leave as-is for now.

PR #466, opened 2022-01-12
String tests

nikhilwoodruff:
# Bug fix: string tests

The test runner now attempts to evaluate string, but if the result is still a string, it'll assert equality.


nikhilwoodruff:
Merging to try and get tests working on #450

PR #482, opened 2022-01-16
SNAP refactoring and further implementation

MaxGhenis:
# New variable formula: `snap_excess_medical_expense_deduction`

- [x] Label field added
- [x] Unit field added
- [x] Unit test(s) added
- [x] Issues this PR fixes linked

## What's changed

Adds SNAP excess medical deduction logic, including the standard medical deduction for California. Note the [USDA State Options Report](https://fns-prod.azureedge.net/sites/default/files/snap/14-State-Options.pdf#33) doesn't provide the actual amounts by state, only lists the states with standard medical deductions, so I had to cite the SNAP Screener data.

This also:
* Adds a person-level `medical_out_of_pocket_expenses` variable, since the existing SPM-unit level one is insufficient for this deduction, which only applies to expenses incurring by elderly or disabled members.
* Renames variables to align with legislation (I don't think we need to do this everywhere but these seemed like low-cost opportunities):
  * `snap_medical_expense_deduction` -> `snap_excess_medical_expense_deduction`
  * `snap_shelter_deduction` -> `snap_excess_shelter_expense_deduction`
  * `snap_earnings_deduction` -> `snap_earned_income_deduction`
* Adds a notebook with SNAP examples


MaxGhenis:
This is pretty close to done but just want to try the [notebook](https://nbviewer.org/github/PolicyEngine/openfisca-us/blob/c8b3a927e7dcf44595324d64725b5f73d366552d/docs/book/programs/snap.ipynb) tomorrow to make sure jupyterbook picks it up, update the changelog, and give it another look. Reviews welcome though. TIL that SNAP is effectively a 30% housing subsidy for certain ranges of housing costs with respect to income!

MaxGhenis:
Update: Many more SNAP changes, including splitting out earned from unearned income. Two things left:
1. Figure out why the [notebook](https://github.com/PolicyEngine/openfisca-us/blob/6cd3cd91eaf57122d115cdc1359b085259de2f14/docs/book/programs/snap.ipynb) is only showing the plotly charts as static images. I've also tried building the jupyterbook locally, and it's shown up either static or not at all. I've tried all combinations of setting execute to auto/force/false, and saving the notebook from jupyterlab and vscode. @nikhilwoodruff the index.ipynb charts showed up as interactive in the jupyterbook, can you think of anything different you might have done there? (BTW I removed all those charts because the code no longer works.)
2. Debug the $16/mo benefit for high-income people who should be ineligible. I'll add some unit tests as I dig in.

openfisca_us/parameters/usda/snap/income/deductions/earned_income.yaml
```diff
@@ -0,0 +1,10 @@
+description: Share of earned income that can be deducted from gross income for SNAP
+values:
+  2005-01-01: 0.2
+metadata:
+  unit: /1
+  name: snap_earned_income_deduction
```
nikhilwoodruff, line 6: Thanks - we'll just need to change this on the next update to PolicyEngine (won't break it as we're hard fixing the dependency)

nikhilwoodruff (approved):

nikhilwoodruff:
> Update: Many more SNAP changes, including splitting out earned from unearned income. Two things left:
> 
> 1. Figure out why the [notebook](https://github.com/PolicyEngine/openfisca-us/blob/6cd3cd91eaf57122d115cdc1359b085259de2f14/docs/book/programs/snap.ipynb) is only showing the plotly charts as static images. I've also tried building the jupyterbook locally, and it's shown up either static or not at all. I've tried all combinations of setting execute to auto/force/false, and saving the notebook from jupyterlab and vscode. @nikhilwoodruff the index.ipynb charts showed up as interactive in the jupyterbook, can you think of anything different you might have done there? (BTW I removed all those charts because the code no longer works.)
> 2. Debug the $16/mo benefit for high-income people who should be ineligible. I'll add some unit tests as I dig in.

Thanks - I wonder if the solution is in the difference between this config and [the UK's config file](https://github1s.com/policyengine/openfisca-uk/blob/HEAD/docs/book/_config.yml)?

docs/book/_config.yml
```diff
@@ -23,4 +23,4 @@ html:
 sphinx:
   config:
     html_js_files:
-    - https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.4/require.min.js
+      - https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.4/require.min.js
```
nikhilwoodruff, line 26: @MaxGhenis I know this seems like a syntax fix but the UK config has the old version, so might be either unnecessary or could be the reason for Plotly's failure?

openfisca_us/tests/policy/baseline/usda/snap/integration.yaml
```diff
@@ -18,13 +18,30 @@
...
     snap_expected_contribution: 5_108.4 # 17028 * 0.3 (flat parameter).
     snap_net_income: 17_028 # 17028 - 0.
     snap_net_income_pre_shelter: 17_028 # 24000 - 2172 - 4800.
-    snap_shelter_deduction: 0 # No shelter costs reported.
+    snap_excess_shelter_expense_deduction: 0 # No shelter costs reported.
     snap_standard_deduction: 2_172 # 181 (parameter for family of 4 in contiguous US) * 12.
-    snap_earnings_deduction: 4_800 # 2000 * 0.2 (flat 20% parameter) * 12.
+    snap_earned_income_deduction: 4_800 # 2000 * 0.2 (flat 20% parameter) * 12.
+
+- name: Single person in CA at 200% of FPL gets minimum benefit via BBCE.
```
MaxGhenis, line 31: @nikhilwoodruff this and the below test check the values for the `vary` issue we were chatting about. They pass as expected.

openfisca_us/parameters/usda/snap/income/deductions/earned_income.yaml
MaxGhenis: Yeah it'll need a few changes, e.g. income by source instead of just market income. So far I've implemented `employment_income`, `self_employment_income`, `dividend_income`, and `interest_income`, all at the person level, though might add more to cover the legislation.

docs/book/_config.yml
MaxGhenis: Yeah I'll try the UK config. This is my autoformatter btw, filed https://github.com/PolicyEngine/openfisca-us/issues/488 to look into this, ideally we'd run a formatter on all files once and no longer have these pop up.

openfisca_us/tests/policy/baseline/usda/snap/integration.yaml
MaxGhenis: Thanks for identifying the issue as `np.any` in `meets_snap_categorical_eligibility` needs an `axis=0` arg. Fixed now.

MaxGhenis:
I still couldn't get the graphs to show up in built jupyterbook docs, despite using the openfisca-uk _config.yml, but otherwise I think this PR is ready for review and merging. It ended up being quite a significant refactor, mostly regarding SNAP but not exclusively.

MaxGhenis:
The error might have something to do with https://github.com/executablebooks/jupyter-book/issues/1348

MaxGhenis:
Error resolved when @nikhilwoodruff ran the notebook. Something to do with my conda environment. Otherwise we went through it and all looks good - we'll see if the jupyterbook shows the plotly charts after merging.

PR #489, opened 2022-01-16
Add nested parameters

nikhilwoodruff:
Adds automated nesting tools from  OpenFisca-Tools to OpenFisca-US.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,7 +35,7 @@ setup(
         "pandas",
         "tqdm",
         "requests",
-        "OpenFisca-Tools>=0.1.9",
+        "OpenFisca-Tools>=0.2.0,<0.3.0",
         "pyyaml",
     ],
     extras_require={


openfisca_us/variables/demographic/household/household_size.py
```diff
@@ -0,0 +1,36 @@
...
+class household_size(Variable):
+    value_type = Enum
+    possible_values = HouseholdSize
+    default_value = HouseholdSize.ONE
+    entity = Household
+    label = "Household size"
+    definition_period = ETERNITY
+
+    def formula(household, period, parameters):
+        return household.nb_persons()
```
MaxGhenis, line 36: is this just for testing? i dont think we need this anywhere atm.

openfisca_us/variables/demographic/geographic/state_code.py
```diff
@@ -66,7 +66,81 @@ class State(Enum):
...
+    default_value = StateCode.CA
+    entity = Household
+    label = "State code"
+    definition_period = ETERNITY
+
+    def formula(household, period, parameters):
+        return StateCode.encode(household("state", period).decode_to_str())
+
+
+class state(Variable):
```
MaxGhenis, line 143: Will a variable named `state` ever create problems given it can be reserved?

In general I'd prefer more descriptiveness, e.g. to indicate it's the two-letter code.

openfisca_us/variables/demographic/household/household_size.py
nikhilwoodruff: Oh I added this for the household_size breakdown - should it be SPM unit size instead?

openfisca_us/variables/demographic/geographic/state_code.py
nikhilwoodruff: Sure, so maybe we just need the state_code variable. Or I could rename this to state_name?

openfisca_us/variables/demographic/household/household_size.py
MaxGhenis: Yeah we already have spm_unit_size

openfisca_us/variables/demographic/geographic/state_code.py
MaxGhenis: state_name sounds good. Could you put these in separate files?

nikhilwoodruff:
@MaxGhenis this should now be ready - I've fixed the bugs that caused test failures and ensured the `str` dummy variables all work - these dummy variables will be removed once the Core PR is merged.

openfisca_us/parameters/hhs/fpg.yaml
```diff
@@ -1,44 +1,48 @@
...
+    breakdown: state_group
+  Contiguous US:
     2017-01-01: 12_060
     2018-01-01: 12_140
     2019-01-01: 12_490
     2020-01-01: 12_760
     2021-01-01: 12_880
     2022-01-12: 13_590
-  AK:
+  Alaska:
```
MaxGhenis, line 12: Do we have to use values instead of codes for this to work? I prefer the conciseness of codes.

MaxGhenis (approved):

PR #507, opened 2022-01-23
School meal improvements

MaxGhenis:
* Categorical eligibility for school meals (except TANF which doesn't work in California yet)
* Legislative references
* Renaming variables

Currently integration tests are failing because the `school_meal_fpg_ratio` that governs economic eligibility doesn't affect income, so households are eligible for SNAP having $0 income, which then makes them categorically eligible. I'll use the same income concepts as SNAP and put those into the integration tests instead to resolve this.


CHANGELOG.md
```diff
@@ -5,249 +5,261 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0).
 
+## [0.26.0] - 2022-01-25
+
+### Added
+
+* Categorical eligibility to school meal subsidies.
```
MaxGhenis, line 12: my formatter changed all the `-` to `*` - @nikhilwoodruff could we try to figure this out when we meet? vscode also just started indenting my yaml files four spaces instead of two...

openfisca_us/tests/policy/baseline/income/spm_unit/spm_unit_market_income.yaml
```diff
@@ -0,0 +1,13 @@
+- name: SPM unit market income is summed person-level market income.
```
MaxGhenis, line 1: I created this because I thought we needed a separate variable to present in the PolicyEngine accounting table. Is that right, or will it sum person-level variables up?

nikhilwoodruff: It'll sum person-level variables up, yes

openfisca_us/variables/income/spm_unit/spm_unit_net_income.py
```diff
@@ -4,26 +4,12 @@
 class spm_unit_net_income(Variable):
     value_type = float
     entity = SPMUnit
-    label = "SPM unit net income"
+    label = "Net income"
```
nikhilwoodruff, line 7: Think we probably already went over this, but just to get it down, I do support this generally - I think we should avoid putting the entity in the label.

nikhilwoodruff (approved):
Thanks @MaxGhenis - moving over to policyengine now.

PR #535, opened 2022-02-02
Metadata for tax logic

nikhilwoodruff:
This fixes bugs and enables PolicyEngine US to show federal tax liability.


openfisca_us/variables/expense/spm_unit_federal_tax.py
```diff
@@ -7,3 +7,6 @@ class spm_unit_federal_tax(Variable):
     label = "SPM unit federal tax"
```
MaxGhenis, line 7: Also for FICA while you're here please.
```suggestion
    label = "Federal income tax"
```

Issue #582, opened 2022-02-21
Confirm Lifeline as a categorical eligibility program for ACP


MaxGhenis:
In describing ACP eligibility, [47 USC 1752(a)(6)](https://uscode.house.gov/view.xhtml?req=granuleid:USC-prelim-title47-section1752&edition=prelim) says:
>The term "eligible household" means, regardless of whether the household or any member of the household receives support under subpart E of part 54 of title 47, Code of Federal Regulations (or any successor regulation), and regardless of whether any member of the household has any past or present arrearages with a broadband provider, a household in which-

It goes on to list other forms of categorical eligibility, which do not include Lifeline itself, though it includes _eligibility_ for Lifeline.

https://www.fcc.gov/acp, however, says that Lifeline participation qualifies a household for ACP.

I've included Lifeline as a categorical eligibility program in #572, pointing to fcc.gov/acp, but it'd be good to understand the legislative justification for that.

DTrim99:
@MaxGhenis Closing as obsolete — this asked about Lifeline categorical eligibility for the Affordable Connectivity Program, but ACP ended in 2024, so the question is moot (no recent activity). Reopen if relevant.

PR #590, opened 2022-02-21
Subtract Lifeline from broadband costs before calculating EBB/ACP

### LINKED ISSUE #588: Subtract Lifeline from broadband costs before calculating ACP/EBB
Currently, Lifeline is `min(9.25, broadband+phone)` and ACP is `min(30, broadband)`. This will overestimate Lifeline+ACP for households with `broadband+phone<39.25`.

I don't see any resources on which order goes first, and it's not necessarily part of the law since the subsidies go to the providers. So I'd suggest we adjust the Lifeline program to be `min(9.25, broadband+phone-ACP)`.

### LINKED ISSUE #589: Check free/reduced price enum values instead of strings
This compares strings instead of enum values:

https://github.com/PolicyEngine/openfisca-us/blob/2605343194d36cc319aa52511c797c927cb759b3/openfisca_us/variables/usda/school_meals/free_school_meals.py#L12-L15

MaxGhenis:
Fixes #588

Fixes #589

### DIFF
--- a/openfisca_us/variables/fcc/lifeline/broadband_cost_after_lifeline.py
+++ b/openfisca_us/variables/fcc/lifeline/broadband_cost_after_lifeline.py
@@ -1,0 +1,16 @@
+from openfisca_us.model_api import *
+
+
+class broadband_cost_after_lifeline(Variable):
+    value_type = float
+    entity = SPMUnit
+    label = "Broadband costs after Lifeline"
+    documentation = "Broadband costs after Lifeline benefits"
+    definition_period = YEAR
+    unit = USD
+    reference = "https://www.law.cornell.edu/cfr/text/47/54.403"
+
+    def formula(spm_unit, period, parameters):
+        broadband = spm_unit("broadband_cost", period)
+        lifeline = spm_unit("lifeline", period)
+        return max_(broadband - lifeline, 0)
--- a/openfisca_us/variables/usda/school_meals/free_school_meals.py
+++ b/openfisca_us/variables/usda/school_meals/free_school_meals.py
@@ -10,6 +10,6 @@ class free_school_meals(Variable):
     documentation = "Value of free school meals"
 
     def formula(spm_unit, period, parameters):
-        tier = spm_unit("school_meal_tier", period).decode_to_str()
-        is_free = tier == "FREE"
+        tier = spm_unit("school_meal_tier", period)
+        is_free = (tier == tier.possible_values.FREE)
         return is_free * spm_unit("school_meal_net_subsidy", period)
--- a/openfisca_us/variables/usda/school_meals/free_school_meals.py
+++ b/openfisca_us/variables/usda/school_meals/free_school_meals.py
@@ -11,5 +11,5 @@ class free_school_meals(Variable):
 
     def formula(spm_unit, period, parameters):
         tier = spm_unit("school_meal_tier", period)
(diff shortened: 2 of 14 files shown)


MaxGhenis:
Ah, the circular reference is that `lifeline > 0` -> `acp_categorical_eligibility` -> `acp` -> `lifeline`. I can either flip the order, which will be more complicated because Lifeline applies to both phone and broadband, or drop the Lifeline categorical eligibility for ACP, which is only specified in the FCC website and not in the law (#582). The latter should only affect a very marginal set of people who enrolled in Lifeline previously and are no longer eligible, since its income limits are more stringent than ACP's.

MaxGhenis:
Another FCC site, https://acpbenefit.org/do-i-qualify, makes it even more clear that Lifeline provides eligibility for ACP:
>If You Already Receive a Lifeline Benefit
Lifeline is a [Federal Communications Commission program](https://www.lifelinesupport.org/) to help make communications services more affordable for low-income consumers. If you currently receive a Lifeline benefit, you automatically qualify for the ACP, and you can receive both benefits at the same time. You can apply your ACP benefit and your Lifeline benefit to the same or separate services. Talk to your internet company to learn how to start receiving ACP discounts.

So I'll keep that categorical eligibility in place and instead subtract from Lifeline first.

MaxGhenis:
I resolved the circular formula by subtracting Lifeline before calculating ACP and EBB. This is ready for review.

openfisca_us/variables/fcc/lifeline/broadband_cost_after_lifeline.py
```diff
@@ -0,0 +1,16 @@
...
+    label = "Broadband costs after Lifeline"
+    documentation = "Broadband costs after Lifeline benefits"
+    definition_period = YEAR
+    unit = USD
+    reference = "https://www.law.cornell.edu/cfr/text/47/54.403"
+
+    def formula(spm_unit, period, parameters):
+        broadband = spm_unit("broadband_cost", period)
+        lifeline = spm_unit("lifeline", period)
+        return max_(broadband - lifeline, 0)
```
nikhilwoodruff, line 16: Should we add a `positive(x) -> max_(x, 0)` function in openfisca-tools?

openfisca_us/variables/usda/school_meals/free_school_meals.py
```diff
@@ -10,6 +10,6 @@ class free_school_meals(Variable):
     documentation = "Value of free school meals"
 
     def formula(spm_unit, period, parameters):
-        tier = spm_unit("school_meal_tier", period).decode_to_str()
-        is_free = tier == "FREE"
+        tier = spm_unit("school_meal_tier", period)
+        is_free = (tier == tier.possible_values.FREE)
```
nikhilwoodruff, line 14: Nice code quality upgrade

nikhilwoodruff (approved):

openfisca_us/variables/fcc/lifeline/broadband_cost_after_lifeline.py
MaxGhenis: I'd prefer not, for the same reasons removing `amount_over` (https://github.com/PolicyEngine/openfisca-tools/issues/13). But feel free to add more rationale in an issue if you'd like.

Issue #439, opened 2022-01-04
Add parameters for TANF California


MaxGhenis:
Bumping this up because we're now implementing categorical eligibility for SNAP in #438, which includes TANF, and since SNAP in turn provides categorical eligibility for other programs like EBB and Lifeline, the absence of TANF in the default state of California is breaking some tests.

We have a couple workaround options:
1. Omit TANF from SNAP categorical eligibility for now
2. Rename the current `tanf` variable to `tanf_if_claimed` and then make a new `tanf` variable which will ultimately be `claims_tanf * tanf_if_claimed`, but omitting that formula for now.

But some rough CA TANF parameters would be best to start with I think. We could always still do (2) here and default `claims_tanf` to false, effectively dropping it from calculations until we're confident in it, or just allowing users to enter their TANF benefit or bool participation rather than calculating it for them.

Relevant state code: https://leginfo.legislature.ca.gov/faces/codes_displayText.xhtml?lawCode=WIC&division=9.&title=&part=3.&chapter=2.&article=6.

MaxGhenis:
Fixed

PR #604, opened 2022-03-06
Added Income Deductions Parameters for California TANF

yvvijay121:
# Added Income Deductions Parameters for California TANF

> Still working on this, just wanted to make a draft PR to help lay out the rest of my work.

The following have already been satisfied through the creation of `tanf_countable_income` from the IL TANF (#369) implementation:
- [x] Label field added
- [x] Unit field added

These still need to be completed for CA TANF (#439)
- [ ] Unit test(s) added
- [ ] Integration test(s) added if relevant

## What's changed

- Changed formula for `tanf_countable_income` variable to include flat earnings deduction that California has, but Illinois does not.
- Added percentage and flat income deductions for California TANF
- Added maximum amounts of TANF aid that California provides in `max_amount`

## What needs to be done

- Still need to add California TANF countable income unit tests.
- Doesn't pass tests yet, still need to debug.
- The maximum amounts for California TANF are only for Region 1, as shown [here](https://leginfo.legislature.ca.gov/faces/codes_displaySection.xhtml?lawCode=WIC&sectionNum=11450.023.). These regions are defined by counties in California; open to any suggestions for determining regions.
- For the test that uses MBSAC (initial employment deduction): $90 
- For the test that uses MAP (earned income deduction): $225 and 50% of remainder
- NOTE: Change CA TANF IED to be based on "per earner", not "per family member"


yvvijay121:
We're currently changing the folder structure of TANF parameters and variables, so it'll take a lot of work to get it to function again. In progress!

openfisca_us/parameters/hhs/tanf/cash/eligibility/initial/deductions/employment/household.yaml
```diff
@@ -36,12 +36,52 @@ IL:
     2021-10-01: 1_426
   18:
     2021-10-01: 1_501
+CA:
+  1:
```
MaxGhenis, line 40: We can remove CA from this and move it to a per-person IED parameter

PR #613, opened 2022-03-09
Update tax parameters for 2022

nikhilwoodruff:
# Added

* Tax parameters for 2022 as per IRS data releases


MaxGhenis (commented):
Will this break 2021 values for parameters where only e.g. 2019 and 2022 are provided? That is, it used to uprate from 2019 to 2021 and now it won't?

nikhilwoodruff:
> Will this break 2021 values for parameters where only e.g. 2019 and 2022 are provided? That is, it used to uprate from 2019 to 2021 and now it won't?

Not any more - added the values from IRS releases in 2019-2022 inclusive.

PR #617, opened 2022-03-11
Simplify WIC uprating and add February 2022 chained CPI-U

MaxGhenis:


MaxGhenis:
The uprating test passes with the original code, but not with the new `propagate_metadata_to_children: true`. @nikhilwoodruff could you advise?

nikhilwoodruff:
> The uprating test passes with the original code, but not with the new `propagate_metadata_to_children: true` (2022 value just equals 2018 value). @nikhilwoodruff could you advise?

The problem was a missing import from openfisca-tools to apply the function that reads that flag and actually applies the metadata downwards. Not sure why it was missing (it means other instances where it was used - nothing policyengine-facing - didn't have all metadata passed down), but added now.

PR #624, opened 2022-03-16
Organise and add tax tests

nikhilwoodruff:
This PR is to separate out tax implementation parts into a more logical structure and add tests where needed (also to try and track down the cause of https://github.com/PolicyEngine/policyengine/issues/560)


MaxGhenis (commented):
Could you make one .py file per variable (with the same name), for the ones you're reorganizing?

nikhilwoodruff:
> Could you make one .py file per variable (with the same name), for the ones you're reorganizing?

Renamed some files, but many of these variables will be renamed in future - left those for another PR.

PR #686, opened 2022-04-15
Fix tax unit head and spouse logic

### LINKED ISSUE #685: `is_tax_unit_head` doesn't return true for at least one member
This should replace `sum` with `min`:

https://github.com/PolicyEngine/openfisca-us/blob/cd0d74db8636c212089be9a77c0c81c51ca498b7/openfisca_us/variables/demographic/tax_unit/is_tax_unit_head.py#L16

This causes other downstream bugs e.g. with the ODC

MaxGhenis:
Fixes #685


openfisca_us/tests/policy/baseline/demographic/tax_unit/is_tax_unit_spouse.yaml
```diff
@@ -0,0 +1,34 @@
...
+      head:
+        is_tax_unit_head: true
+      spouse: {}
+      # Based on sequence of members.
+      other_dep: {}
+    tax_units:
+      tax_unit:
+        members: [child, head, spouse, other_dep]
+  output:
+    is_tax_unit_head: [false, false, true, false]
```
nikhilwoodruff, line 34:
```suggestion
    is_tax_unit_spouse [false, false, true, false]
```

Should this be `spouse`?

PR #687, opened 2022-04-15
Use CDCC max dependents parameter and other minor changes

### LINKED ISSUE #650: Remove tax-unit-level payroll taxes
Payroll taxes are paid at the individual level (except the additional Medicare tax). They're only aggregated to the tax unit level for taxdata/taxcalc.

Doing this will also avoid the issue with dropping dependents' payroll taxes #648

### LINKED ISSUE #684: Use `cdcc.eligibility.max` parameter
Currently it's hard-coded at 2:

https://github.com/PolicyEngine/openfisca-us/blob/cd0d74db8636c212089be9a77c0c81c51ca498b7/openfisca_us/variables/irs/credits/cdcc/cdcc.py#L40

which is the value of the parameter:

https://github.com/PolicyEngine/openfisca-us/blob/cd0d74db8636c212089be9a77c0c81c51ca498b7/openfisca_us/parameters/irs/credits/cdcc/eligibility/max.yaml#L3

That parameter is currently unused.

MaxGhenis:
* Fixes #684 
* Fixes #650 
* Adds some parameter metadata, including legislative references for CDCC parameters and variables
* Split up `cdcc.py` into one file per variable
* Replaces `"currency-USD"` with `USD`
* Minor CDCC refactoring

You can look at the first three commits for the substantive changes, before I split the variables into separate files.


nikhilwoodruff (approved):
Nice, thanks @MaxGhenis. No improvements here.

PR #701, opened 2022-04-19
HUD total tenant payment

### LINKED ISSUE #659: HUD total tenant payment
https://www.law.cornell.edu/cfr/text/24/5.628
Calculated monthly:
`max(hud_adjusted_income * 0.3, hud_annual_income * 0.1, (3rd piece - skip), hud_minimum_rent)`

Add parameters for 0.3, 0.1, and minimum rent (https://www.law.cornell.edu/cfr/text/24/5.630).

[PRD section](https://www.atlantafed.org/-/media/documents/economic-mobility-and-resilience/advancing-careers-for-low-income-families/policy-rules-database/policy-rules-database-technical-manual.pdf#page=39)

kvanalebeek:
Fixes #659


MaxGhenis:
@nikhilwoodruff could you advise on [this `UnboundLocalError: local variable 'possible_values' referenced before assignment` error](https://github.com/PolicyEngine/openfisca-us/runs/6188254257?check_suite_focus=true#step:5:204)? I think it has to do with us listing `hud_income_level` as a breakdown for the [`hud.ami_limit.family_size` parameter](https://github.com/PolicyEngine/openfisca-us/pull/701/files#diff-391be75d6f46e7b36fdfeb9feeb65b851cbd06e9fc2adc7d8e97008a251291de) before it's assigned. Unlike other smart-indexing uses, this is just a way to organize parameters rather than being a person/household-level variable to break down the parameter by. [Here's the relevant CFR](https://www.law.cornell.edu/cfr/text/24/81.17), should we do it as separate parameter files instead?

Also, do you know why we're getting the changelog failure?

nikhilwoodruff:
Hey @kvanalebeek, @MaxGhenis - the issue actually turned out to be unrelated to the `hud_income_level` parameter, and was instead caused by the `spm_unit_size` breakdown option. I've changed this to `range(1, 5)` given the data, and I think in general it's better to use the `range` syntax over variables with type `int` (as the breakdown logic doesn't know the possible values for an `int`).

nikhilwoodruff:
Oh, just found the cause of the check-version fail. It's because the PR is from a fork, rather than a branch. On @kvanalebeek's fork, the version tags on the main repo don't exist, and the check-version action uses these tags to find the functional changes in commits since the last version. I think we can ignore the check-version result here - the changelog preview looks fine to me.

MaxGhenis:
Thanks @nikhilwoodruff! The current failures seem to be from using the wrong variable names.

MaxGhenis:
Test failures should now be fixed, but we encountered some unexpected results in the notebook, so we'll add some more tests before opening this for review and merging.

openfisca_us/parameters/hud/fpl_income_limit/low.yaml
```diff
@@ -0,0 +1,10 @@
+description: Poverty line limit to qualify as low income for HUD programs.
```
MaxGhenis, line 1: Let's delete this since we have ami_limit/family_size.yaml

MaxGhenis (approved):

PR #723, opened 2022-05-01
Tie state income taxes to the SALT deduction

### LINKED ISSUE #722: Tie state income tax to SALT deduction
Currently the calculated value `state_income_tax` (MA only for now) is disconnected from `filer_e18400`, which is "Itemizable state and local income/sales taxes", so it's not getting into the SALT deduction.

MaxGhenis:
Fixes #722


MaxGhenis:
@nikhilwoodruff any tips on what might be breaking this? The change is pretty simple so I'm surprised, but it's getting into state income tax parameters. Here's the [verbose output](https://gist.github.com/MaxGhenis/54be6b538f4063fc4756cba13caaf56b) from the failing test.

MaxGhenis:
Looks like it may be an openfisca-tools issue. @nikhilwoodruff to investigate.

nikhilwoodruff:
It turned out to be because the local state parameters were being accessed in 2020 - they just needed some value in place.

PR #734, opened 2022-05-02
Add 50-state SNAP parameters

nikhilwoodruff:
Adds 50-state data from snapscreener.com.


openfisca_us/variables/usda/snap/income/deductions/shelter/snap_excess_shelter_expense_deduction.py
```diff
@@ -34,8 +34,14 @@ def formula(spm_unit, period, parameters):
...
         # Homeless shelter deduction is flat and has no utility component.
+        state = spm_unit.household("state_code_str", period)
+        homeless_deduction = (
+            p.homeless.deduction * WEEKS_IN_YEAR * p.homeless.available[state]
+        )
         return where(
-            spm_unit.household("is_homeless", period),
-            p.homeless * 12,
+            spm_unit.household("is_homeless", period)
+            & (housing_cost > 0)
```
nikhilwoodruff, line 43: Added this line here because of [the law](https://www.law.cornell.edu/uscode/text/7/2014#e_4_D) and also [this report](https://fns-prod.azureedge.us/sites/default/files/snap/14-State-Options.pdf#page=20).

openfisca_us/variables/expense/child_support.py
```diff
@@ -0,0 +1,10 @@
+from openfisca_us.model_api import *
+
+
+class child_support_expense(Variable):
+    value_type = float
+    entity = Person
+    label = "Child support expense"
+    unit = "currency-USD"
```
MaxGhenis, line 8:
```suggestion
    unit = USD
```

openfisca_us/variables/expense/count_distinct_utility_expenses.py
```diff
@@ -0,0 +1,26 @@
...
+
+class count_distinct_utility_expenses(Variable):
+    value_type = int
+    entity = SPMUnit
+    label = "Number of distinct utility expenses"
+    documentation = "The number of distinct utility expenses."
+    definition_period = YEAR
+
+    def formula(spm_unit, period, parameters):
+        utilities = [
```
MaxGhenis, line 12:
```suggestion
        UTILITIES = [
```

openfisca_us/variables/expense/phone_expense.py
```diff
@@ -0,0 +1,12 @@
+from openfisca_us.model_api import *
+
+
+class telephone_expense(Variable):
```
MaxGhenis, line 4: Not `phone`?

openfisca_us/variables/expense/phone_expense.py
```diff
@@ -0,0 +1,12 @@
...
+    value_type = float
+    entity = SPMUnit
+    label = "Phone expense"
+    documentation = "Phone line cost for this SPM unit"
+    unit = "currency-USD"
+    definition_period = YEAR
+
+    formula = sum_of_variables(["phone_cost"])  # For compatibility
```
MaxGhenis, line 12: See https://github.com/PolicyEngine/openfisca-us/issues/735

MaxGhenis (approved):

openfisca_us/variables/expense/phone_expense.py
nikhilwoodruff: I think telephone is more specific to the phone line and not mobile.

openfisca_us/variables/usda/snap/income/deductions/shelter/snap_excess_shelter_expense_deduction.py
MaxGhenis: Good catch. For posterity, it's https://www.law.cornell.edu/uscode/text/7/2014#e_6_D_i_II:
>that are not receiving free shelter throughout the month

openfisca_us/variables/expense/phone_expense.py
MaxGhenis: USDA allows deducting mobile expenses too: https://www.fns.usda.gov/snap/eligibility/resources/non-land-line-telephone-expenses-allowances-deductions

This variable also applies to Lifeline, which is for mobile, so if we need landline specifically it should be a separate variable.

MaxGhenis:
Did you use a Python script that could go in `tools/dev` to write these yaml files?

openfisca_us/variables/expense/phone_expense.py
nikhilwoodruff: Ah, OK. Reverted, thanks

nikhilwoodruff:
> Did you use a Python script that could go in `tools/dev` to write these yaml files?

I just used a notebook with about 4 lines of Python code in them (read the csv file, print out the min for each state for a particular column). Probably not worth adding IMO.

nikhilwoodruff:
Ah, just realised the emergency allotment needs scraping/

MaxGhenis:
Could you drop the 4 lines here for posterity?

nikhilwoodruff:
Sure, given the CSV file from snapscreener.com as `snap_parameters.csv`:
```python
import pandas as pd
df = pd.read_csv("snap_parameters.csv")
df["state"] = df.STATE.apply(lambda x: x[:2])
df = df.groupby("state").min()
for state, value in zip(df.index, df.ELECTRICITY): # Replace with each column name and re-run
  print(f"{state}:\n  2010-01-01: 0\n  2021-10-01: {value:_.0f}\n")
```

PR #796, opened 2022-05-20
Add ACA tax credits

nikhilwoodruff:
## Premium Tax Credit

This PR adds the Premium Tax Credit and logic necessary to calculate it in Massachusetts.


openfisca_us/variables/household/demographic/geographic/zip_code.py
```diff
@@ -0,0 +1,8 @@
+from openfisca_us.model_api import *
+
+
+class zip_code(Variable):
+    value_type = str
+    entity = Household
+    label = "Zip code"
```
MaxGhenis, line 7: Acronym for Zone Improvement Plan
```suggestion
    label = "ZIP code"
```

openfisca_us/variables/household/demographic/geographic/county_zip_code.py
```diff
@@ -0,0 +1,12 @@
+from openfisca_us.model_api import *
+
+
+class county_zip_code(Variable):
+    value_type = str
+    entity = Household
+    label = "County code from zip code"
```
MaxGhenis, line 7: Is this standard across the country? Could you add a reference?

nikhilwoodruff: Probably a bit more involved for other states - added a reference

Issue #837, opened 2022-05-30
Incorrect CDCC amounts


martinholmer:
The recently revised CDCC formula still has problems: it seems to not always (ever?) cap eligible care expenses to $3000 (for one eligible dependent) or $6000 (for two or more eligible dependents).

Here's my environment:
```
openfisca-core            35.8.0                   pypi_0    pypi
openfisca-tools           0.10.4                   pypi_0    pypi
openfisca-us              0.68.0                   pypi_0    pypi
```
Here is a test case that illustrates the problem:
```
(openfisca-us) OpenFisca-US% cat f18-id483.test.ymal
# OpenFisca-US test file derived from the f18.its.csv and f18.ots.csv files
# Expected test output from NBER TAXSIM Model v35 (05/18/22) With TCJA

- name: Tax unit with taxsimid 483 from f18.its.csv
  absolute_error_margin: 0.01
  period: 2018
  input:
    people:
      person1:
        is_tax_unit_head: 1
        is_tax_unit_spouse: 0
        is_tax_unit_dependent: 0
        age: 41
        employment_income: 292000
        unemployment_compensation: 5000
        social_security: 19000
        pension_income: 43000
        has_marketplace_health_coverage: false
      person2:
        is_tax_unit_head: 0
        is_tax_unit_spouse: 1
        is_tax_unit_dependent: 0
        age: 45
        employment_income: 139000
        has_marketplace_health_coverage: false
      person3:
        is_tax_unit_head: 0
        is_tax_unit_spouse: 0
        is_tax_unit_dependent: 1
        age: 12
        has_marketplace_health_coverage: false
      person4:
        is_tax_unit_head: 0
        is_tax_unit_spouse: 0
        is_tax_unit_dependent: 1
        age: 12
        has_marketplace_health_coverage: false
      person5:
        is_tax_unit_head: 0
        is_tax_unit_spouse: 0
        is_tax_unit_dependent: 1
        age: 12
        has_marketplace_health_coverage: false
      person6:
        is_tax_unit_head: 0
        is_tax_unit_spouse: 0
        is_tax_unit_dependent: 1
        age: 12
        has_marketplace_health_coverage: false
    tax_units:
      tax_unit:
        members: [person1, person2, person3, person4, person5, person6]
        tax_unit_childcare_expenses: 10000
    households:
      household:
        members: [person1, person2, person3, person4, person5, person6]
        state_code: TX
  output:
[...]

PR #838, opened 2022-05-30
Fix CDCC bug missing childcare expense cap

nikhilwoodruff:
This PR aims to fix the bug found by @martinholmer relating to the CDCC childcare expenses. I've added the integration test used to identify the bug to the test folder here.

### DIFF
--- a/changelog.yaml
+++ b/changelog.yaml
@@ -733,3 +733,7 @@
     added:
     - EITC parameters for 2017 and 2018.
   date: 2022-05-28 06:59:45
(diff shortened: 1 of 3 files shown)


MaxGhenis (approved):
As you mentioned, it'd be good to implement this by describing a cap of expenses rather than cap of children (per the law), but this works for now. Thanks.

martinholmer:
@nikhilwoodruff, Thanks for the quick response on issue #837.

PR #843, opened 2022-05-31
Add early termination logic to Premium Tax Credit formula

nikhilwoodruff:
Adds in @martinholmer's local change, which I think is a good idea to add in the meantime before we figure out a better solution to the intensive SLSPC matching process.


MaxGhenis:
I'm not sure about this. It sets an inconsistent precedent, and users have other options to bypass the computation like setting premium_tax_credit to zero. This would also only save considerable time when doing a lot of individual runs separately instead of the suggested approach of vectorizing.

PR #917, opened 2022-06-13
Add childcare from the CPS

nikhilwoodruff:
Raises the budgetary impact for the CDCC to $3.47bn, compared to the JCT's $4.7bn. The main things causing the underestimation are likely:
* Employer childcare exclusion (unknown proportion, probably low)
* PUF vs CPS income reporting
* Childcare expenses not being proportionate to the number of children in the tax unit compared to the SPM unit


nikhilwoodruff:
Also fixed the tax unit childcare formula to split per-child, rather than per-person.

PR #939, opened 2022-06-17
Bug fixes

nikhilwoodruff:
Fixes a bug relating to payroll taxes and market income, which caused some households to have negative income (this was affecting deep poverty estimates).


openfisca_us/variables/gov/irs/tax/payroll/payroll_tax_gross_wages.py
```diff
@@ -8,4 +8,8 @@ class payroll_tax_gross_wages(Variable):
     definition_period = YEAR
     unit = USD
 
-    formula = sum_of_variables(["employment_income", "pension_contributions"])
+    def formula(person, period, parameters):
+        earnings = person("employment_income", period)
+        pension_contributions = person("pension_contributions", period)
+        return max_(0, earnings - pension_contributions)
```
MaxGhenis, line 14: Just confirming this is intended to be a subtraction?

nikhilwoodruff: Corrected - CPS earnings are before pension contributions, think we should follow that.

PR #1006, opened 2022-06-29
Don't stop testing after the first failure

nikhilwoodruff:
cc @MattHJensen


MattHJensen:
Thanks @nikhilwoodruff!

PR #1017, opened 2022-07-01
Add 40-state ACA subsidies

nikhilwoodruff:
This uses the OSPC scraper linked by @martinholmer. The scraper worked until state 40, after which something crashed. Probably not too hard to fix it when we need to (and a very nice surprise that a web scraper works this much, this many years after it was originally designed).


nikhilwoodruff:
Looks like it's just South Carolina that broke it, so we can get to 49.

nikhilwoodruff:
Superseded by #1080

PR #1046, opened 2022-07-05
Virginia income tax

andersonfrailey:
Started working on coding up Virginia's tax code. Finished most of the 760 form (their primary income tax form) but have a few questions.

1. I can't tell if the tests I've added are running. Even after adding a variety of new tests, pytest shows the same number of tests running as before. Is this because all of the variable tests are run via one function so no matter how many individual tests we have pytest interprets it as one test?
2. Lines 20-34 seem like variables  we'll never actually have data on. Should I make placeholder variables any way, or ignore them? Are there established guidelines for when a variable is worth adding, or is it just a judgement call?
Screen Shot 2022-07-05 at 11 34 31 AM
3. For filing status, I wasn't able to tell what the convention was. I know in Tax-Calc we use 1, 2, 3, 4. Here it seemed like the parameter files where filing status was important there was a mix of using `single`, `joint, and `separate` in all caps or lower case. Does the distinction matter?

I think there was something else... but I've forgotten what it is so I'll post another comment when I think of it.


nikhilwoodruff:
Thanks @andersonfrailey! Responses inline.

> 1. I can't tell if the tests I've added are running. Even after adding a variety of new tests, pytest shows the same number of tests running as before. Is this because all of the variable tests are run via one function so no matter how many individual tests we have pytest interprets it as one test?

I think your tests are definitely running on GH actions - see the below screenshot. Do you get the same result locally as on GitHub?

image

> 2. Lines 20-34 seem like variables  we'll never actually have data on. Should I make placeholder variables any way, or ignore them? Are there established guidelines for when a variable is worth adding, or is it just a judgement call?

So far it's a judgement call, though I think we'll eventually have enough collective experience to figure out exactly what strategy works best. I think we should prioritise, in this order:
* Enable all information you can put in a tax form to be able to be input into OpenFisca-US (not necessarily as an individual input, maybe summing multiple before inputting: e.g. an input variable for ("other income sources for VA" with a comment saying to add up lines X through Y in the tax form)).
* Minimising the number of variables.

> 3. For filing status, I wasn't able to tell what the convention was. I know in Tax-Calc we use 1, 2, 3, 4. Here it seemed like the parameter files where filing status was important there was a mix of using `single`, `joint, and `separate` in all caps or lower case. Does the distinction matter?

Yep - I think we need to use all caps wherever it's used. [See here for an example](https://github.com/PolicyEngine/openfisca-us/blob/master/openfisca_us/parameters/gov/states/ma/tax/income/exemptions/personal.yaml).

openfisca_us/variables/gov/states/va/tax/deductions/va_age_deduction.py
```diff
@@ -0,0 +1,40 @@
...
+            gov.states.va.tax.deductions.income_based_age_deduction
+        )
+        age_thd = va_deductions_path.age
+        # number of tax payers claiming an income based deduction
+        filing_status = tax_unit("filing_status", period)
+        age_head = tax_unit("age_head", period)
+        age_spouse = tax_unit("age_spouse", period)
+
+        n_filers = (age_head >= age_thd) + (age_spouse >= age_thd)
+        if n_filers == 0:
```
MaxGhenis, line 23: OpenFisca is numpy-based, so you'll need to replace `if` statements with `where`s. In this case, since you're multiplying by the number of aged filers, you can delete these lines.

openfisca_us/variables/gov/states/va/tax/income/adjusted_gross_income/va_adjusted_gross_income.py
```diff
@@ -0,0 +1,27 @@
...
+    value_type = float
+    entity = TaxUnit
+    label = "VA adjusted gross income"
+    unit = USD
+    definition_period = YEAR
+
+    def formula(tax_unit, period, parameters):
+        agi = tax_unit("adjusted_gross_income", period)
+        additions = tax_unit("va_income_additions", period)
+        age_deduction = tax_unit("va_age_deduction", period)
```
MaxGhenis, line 15: Could you make a parameter listing the deductions? Then you can do this:
```suggestion
        deductions = add(tax_unit, period, parameters(period).gov.states.va.[...].deductions)
```
followed by
```
return agi + additions - deductions
```

openfisca_us/variables/gov/states/va/tax/income/va_income_subtractions.py
```diff
@@ -0,0 +1,9 @@
+from openfisca_us.model_api import *
+
+
+class va_income_substractions(Variable):
```
MaxGhenis, line 4:
```suggestion
class va_income_subtractions(Variable):
```

openfisca_us/variables/gov/states/va/tax/income/va_income_tax_main_rates.py
```diff
@@ -0,0 +1,35 @@
...
+        va_brackets = va_income.brackets
+        va_agi = tax_unit("va_adjusted_gross_income", period)
+        filing_status = tax_unit("filing_status", period)
+        if va_agi < va_income.va_filing_thresholds[filing_status]:
+            return 0
+        va_taxable_income = tax_unit("va_taxable_income", period)
+
+        reg_tax = 0
+        last_reg_threshold = 0
+        for i in range(1, 3):
```
MaxGhenis, line 22: Could you use [OpenFisca scale parameters](https://openfisca.org/doc/coding-the-legislation/legislation_parameters.html#creating-scales) for this? Simpler and faster.

openfisca_us/variables/gov/states/va/tax/deductions/va_age_deduction.py
```diff
@@ -0,0 +1,40 @@
...
+        va_deductions_path = (
+            gov.states.va.tax.deductions.income_based_age_deduction
+        )
+        age_thd = va_deductions_path.age
+        # number of tax payers claiming an income based deduction
+        filing_status = tax_unit("filing_status", period)
+        age_head = tax_unit("age_head", period)
+        age_spouse = tax_unit("age_spouse", period)
+
+        n_filers = (age_head >= age_thd) + (age_spouse >= age_thd)
```
MaxGhenis, line 22:
```suggestion
        aged_filer_count = (age_head >= age_thd) + (age_spouse >= age_thd)
```

openfisca_us/variables/gov/states/va/tax/deductions/va_age_deduction.py
```diff
@@ -0,0 +1,40 @@
...
+    label = "VA age deduction"
+    unit = USD
+    definition_period = YEAR
+
+    def formula(tax_unit, period, parameters):
+        gov = parameters(period).gov
+        va_deductions_path = (
+            gov.states.va.tax.deductions.income_based_age_deduction
+        )
+        age_thd = va_deductions_path.age
```
MaxGhenis, line 16:
```suggestion
        age_threshold = va_deductions_path.age
```

openfisca_us/tests/policy/baseline/states/va/tax/income/va_taxable_income.yaml
```diff
@@ -0,0 +1,65 @@
...
+    va_adjusted_gross_income: 75_000
+  output:
+    va_taxable_income: 75_000 - 5_000 - 5_000
+
+- name: VA taxable income test - separate standard filer
+  period: 2021
+  absolute_error_margin: 0
+  input:
+    va_itemized_deductions: 2_000
+    filing_status: "separate"
```
MaxGhenis, line 61:
```suggestion
    filing_status: SEPARATE
```

openfisca_us/parameters/gov/states/va/tax/deductions/income_based_age_deduction.yaml
```diff
@@ -0,0 +1,29 @@
+age:
```
MaxGhenis, line 1: Could you split this into three files in the same folder?

openfisca_us/parameters/gov/states/va/tax/deductions/income_based_age_deduction.yaml
```diff
@@ -0,0 +1,29 @@
...
+  description: Minimum age a tax payer must be to claim the income-based age deduction
+  values:
+    2021-01-01: 65
+  metadata:
+    unit: year
+    label: VA income-based age deduction threshold
+income:
```
MaxGhenis, line 8: Also please add a `phase_out_rate` parameter, equal to `1` if I'm reading correctly.
```suggestion
phase_out_start:
```

openfisca_us/parameters/gov/states/va/tax/deductions/income_based_age_deduction.yaml
```diff
@@ -0,0 +1,29 @@
...
+  single:
+    values:
+      2021-01-01: 50_000
+  joint:
+    values:
+      2021-01-01: 75_000
+  metadata:
+    unit: currency-USD
+    label: VA income-based age deduction income limits
+deduction:
```
MaxGhenis, line 19:
```suggestion
amount:
```

MaxGhenis (commented):
Thanks Anderson, looks great so far. I reverted this to draft (we use this instead of WIP in the titles) and gave some initial feedback following on Nikhil's comments. Feel free to ping when you'd like another look.

MaxGhenis:
Are you installing with `pip install -e .` from the root folder?

PR #1063, opened 2022-07-09
Impute CPS ages from 80 to 84 as integers

### LINKED ISSUE #1062: Assign age as an int in CPS
It gets converted to a float with non-integer values here:

https://github.com/PolicyEngine/openfisca-us/blob/2897abd67a48629184b8f7107b4ddd66e00e5b0a/openfisca_us/data/datasets/cps/cps.py#L141-L145

We can replace `random.rand` with `random.randint`.

MaxGhenis:
Fixes #1062

Comments verbatim from the [CPS documentation](https://www2.census.gov/programs-surveys/cps/datasets/2021/march/asec2021_ddl_pub_full.pdf#page=27)

### DIFF
--- a/changelog.yaml
+++ b/changelog.yaml
@@ -1064,3 +1064,7 @@
     added:
     - MA COVID-19 Essential Employee Premium Pay Program.
   date: 2022-07-08 12:26:04
+- bump: patch
+  changes:
+    changed:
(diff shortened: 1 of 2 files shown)


nikhilwoodruff (approved):
Thanks @MaxGhenis!

PR #1084, opened 2022-07-12
IN deductions 

MattHJensen:


openfisca_us/variables/gov/states/in/tax/income/deductions/in_num_children_nonpublic_school.py
```diff
@@ -2,10 +2,9 @@
 
 
 class in_num_children_nonpublic_school(Variable):
```
MaxGhenis, line 4:
```suggestion
class in_count_children_nonpublic_school(Variable):
```

MattHJensen:
@nikhilwoodruff and @MaxGhenis this is ready for your review. Thanks in advance.

openfisca_us/variables/gov/states/in/tax/income/deductions/in_military_service_deduction.py
```diff
@@ -0,0 +1,17 @@
...
+
+class in_military_service_deduction(Variable):
+    value_type = float
+    entity = TaxUnit
+    label = "IN military service deduction"
+    definition_period = YEAR
+    unit = USD
+    reference = "http://iga.in.gov/legislative/laws/2021/ic/titles/006#6-3-2-4"  # (a)(1)
+
+    def formula(tax_unit, period, parameters):
```
MaxGhenis, line 12: I think we can unify this with `in_person_military_service` and make it tax-unit-level, for consistency with exemptions and to save memory. To do this, start with `person = tax_unit.members` and then port everything from the other formula, and `return tax_unit.sum(person_deduction)`.

But for future reference, `formula = sum_of_variables(["in_person_military_service_deduction"])` would have worked here.

openfisca_us/variables/gov/states/in/tax/income/deductions/in_nonpublic_school_deduction.py
```diff
@@ -0,0 +1,20 @@
+from openfisca_us.model_api import *
+
+
+class in_nonpublic_school_deduction(Variable):
+    value_type = float
+    entity = TaxUnit
+    label = "Nonpublic school expenditures deduction for IN"
+    definition_period = YEAR
+    unit = USD
+    reference = "http://iga.in.gov/legislative/laws/2021/ic/titles/006#6-3-2-4"  # (d)(1)
```
MaxGhenis, line 10: Not seeing it here

openfisca_us/variables/gov/states/in/tax/income/deductions/in_unemployment_compensation_deduction.py
```diff
@@ -0,0 +1,31 @@
...
+    label = "IN Unemployment compensation deduction"
+    unit = USD
+    definition_period = YEAR
+    reference = (
+        "http://iga.in.gov/legislative/laws/2021/ic/titles/006#6-3-2-10"
+    )
+
+    def formula(tax_unit, period, parameters):
+        p = parameters(period).gov.states["in"].tax.income.deductions
+        unemployment_compensation_in_federal_AGI = tax_unit(
```
MaxGhenis, line 16: And elsewhere; we always use lowercase.
```suggestion
        unemployment_compensation_in_federal_agi = tax_unit(
```

openfisca_us/variables/gov/states/in/tax/income/deductions/in_unemployment_compensation_deduction.py
```diff
@@ -0,0 +1,31 @@
...
+            "tax_unit_taxable_unemployment_compensation", period
+        )
+        federal_AGI = tax_unit("adjusted_gross_income", period)
+        filing_status = tax_unit("filing_status", period)
+        AGI_reduction = p.unemployment_compensation.agi_reduction[
+            filing_status
+        ]
+        reduced_AGI = max(0, federal_AGI - AGI_reduction)
+        in_taxable_unemployment_compensation = min(
+            (0.5 * reduced_AGI), unemployment_compensation_in_federal_AGI
```
MaxGhenis, line 26: Where is the 0.5 in the law? Could you make that a parameter?

openfisca_us/variables/gov/states/in/tax/income/exemptions/in_qualifying_child_count.py
```diff
@@ -5,7 +5,6 @@ class in_qualifying_child_count(Variable):
     value_type = int
     entity = TaxUnit
     label = "IN qualifying depdendent child count"
```
MaxGhenis, line 7: We can have arbitrary strings here.
```suggestion
    label = "IN qualifying dependent child count"
    unit = "child"
```

openfisca_us/variables/gov/states/in/tax/income/deductions/in_rent.py
```diff
@@ -1,10 +1,11 @@
 from openfisca_us.model_api import *
 
 
-class in_deductions(Variable):
+class in_rent(Variable):
```
MaxGhenis, line 4: Good to be precise, but since many states have rental credits/deductions and taxsim unifies them, let's use our existing `rent` variable here (this will also save space). Could you add a comment in the `in_renters_deduction` variable indicating the restrictions (principal place of residence subject to IN property tax)?

openfisca_us/variables/gov/states/in/tax/income/deductions/in_count_children_nonpublic_school.py
```diff
@@ -0,0 +1,10 @@
+from openfisca_us.model_api import *
+
+
+class in_count_children_nonpublic_school(Variable):
```
MaxGhenis, line 4: Can you change this to person-level and move it to https://github.com/PolicyEngine/openfisca-us/blob/master/openfisca_us/variables/household/demographic/person as `is_in_k12_nonpublic_school`, a la the existing `is_in_k12_school`? Similar to rent, making this US-wide will save space and serve similar state laws, at the cost of some precision.

openfisca_us/parameters/gov/states/in/tax/income/deductions/homeowners_property_tax/max.yaml
```diff
@@ -0,0 +1,21 @@
...
+  2021-01-01: 2_500
+HEAD_OF_HOUSEHOLD:
+  2021-01-01: 2_500
+JOINT:
+  2021-01-01: 2_500
+WIDOW:
+  2021-01-01: 2_500
+SEPARATE:
+  2021-01-01: 1_250
+reference:
```
MaxGhenis, line 19: move to metadata

openfisca_us/parameters/gov/states/in/tax/income/deductions/military_service/max.yaml
```diff
@@ -0,0 +1,13 @@
...
+  2021-01-01: 5_000
+metadata:
+  period: year
+  unit: currency-USD
+  name: in_max_military_service_deduction
+  label: IN max military service deduction
+  reference:
+    - title: IC 6-3-2-4	Military service deduction; retirement income or survivor's benefits deduction 
+      href: http://iga.in.gov/legislative/laws/2021/ic/titles/006#6-3-2-4 (a)(1)
+
```
MaxGhenis, line 12: remove spare lines

openfisca_us/parameters/gov/states/in/tax/income/deductions/nonpublic_school/amount.yaml
```diff
@@ -0,0 +1,12 @@
...
+  2021-01-01: 1_000
+metadata:
+  period: year
+  unit: currency-USD
+  name: in_nonpublic_school_deduction
+  label: IN nonpublic school deduction
+  reference:
+    - title: IC 6-3-2-22 Deduction; unreimbursed education expenditures
+      href: http://iga.in.gov/legislative/laws/2021/ic/titles/006#6-3-2-4 (d)(1)
+
```
MaxGhenis, line 12: remove spare line

openfisca_us/parameters/gov/states/in/tax/income/deductions/nonpublic_school/amount.yaml
```diff
@@ -0,0 +1,12 @@
...
+values:
+  2021-01-01: 1_000
+metadata:
+  period: year
+  unit: currency-USD
+  name: in_nonpublic_school_deduction
+  label: IN nonpublic school deduction
+  reference:
+    - title: IC 6-3-2-22 Deduction; unreimbursed education expenditures
+      href: http://iga.in.gov/legislative/laws/2021/ic/titles/006#6-3-2-4 (d)(1)
```
MaxGhenis, line 11: Updating and making a valid URL (it gets linked in https://policyengine.org/us/api-explorer)
```suggestion
      href: http://iga.in.gov/legislative/laws/2021/ic/titles/006#6-3-2-22
```

openfisca_us/parameters/gov/states/in/tax/income/deductions/nonpublic_school/amount.yaml
```diff
@@ -0,0 +1,12 @@
+description: IN deduction for qualified dependent children in private, parochial, or home school with unreimbursed expense
+values:
+  2021-01-01: 1_000
+metadata:
+  period: year
+  unit: currency-USD
+  name: in_nonpublic_school_deduction
+  label: IN nonpublic school deduction
+  reference:
+    - title: IC 6-3-2-22 Deduction; unreimbursed education expenditures
```
MaxGhenis, line 10:
```suggestion
    - title: IC 6-3-2-22 Deduction; unreimbursed education expenditures (d)(1)
```

openfisca_us/parameters/gov/states/in/tax/income/deductions/unemployment_compensation/agi_reduction.yaml
```diff
@@ -0,0 +1,21 @@
...
+SINGLE:
+  2021-01-01: 12_000
+HEAD_OF_HOUSEHOLD:
+  2021-01-01: 12_000
+JOINT:
+  2021-01-01: 18_000
+WIDOW:
+  2021-01-01: 12_000
+SEPARATE:
+  2021-01-01: 12_000
```
MaxGhenis, line 18: Does (3)(B) refer to separate filers?
>(B) Zero dollars ($0) if the individual:
(i) is married at the close of the taxable year, as determined under Section 7703 of the Internal Revenue Code;
(ii) does not file a joint return for the taxable year; and
(iii) does not live apart from the individual's spouse at all times during the taxable year.

MaxGhenis (approved):

MaxGhenis:
Could you also run `make format` and if `linecheck` isn't working add blank newlines to the end of files? Might be easiest to identify them in GitHub.

openfisca_us/parameters/gov/states/in/tax/income/deductions/unemployment_compensation/agi_reduction.yaml
MattHJensen: @MaxGhenis, a note in the instructions clarifies this: 

> If you were married but filing separately, and you lived with your spouse at any time during the year, enter -0- on line 3 of the worksheet. However, if you were MFS and lived apart from your spouse the entire year, enter $12,000 on line 3. 

My instinct is to leave the parameter as is ($12k) while adding a comment about this, but let me know if you think we should set to $0 by default for MFS or follow some other path.

openfisca_us/variables/gov/states/in/tax/income/deductions/in_unemployment_compensation_deduction.py
MattHJensen: It is in 6-3-2-10 (a) "For purposes of this section, "excess adjusted gross income" means the greater of zero (0) or one-half (1/2) of:"

openfisca_us/tests/policy/baseline/gov/states/in/tax/income/deductions/in_deductions.yaml
```diff
@@ -0,0 +1,16 @@
+- name: All IN deductions
```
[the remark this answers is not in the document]
MattHJensen: I've caught typos w/ these so am going to leave it unless you say otherwise.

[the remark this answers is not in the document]
MaxGhenis: `test_variables.py` checks that every variable computes without breaking. Does that provide enough info to check for typos without the yaml? I'd prefer to enhance its error messages than add unit tests to every `sum_of_variables` (for consistency).

openfisca_us/tests/policy/baseline/gov/states/in/tax/income/deductions/in_renters.yaml
```diff
@@ -0,0 +1,22 @@
+- name: Homeowner's renters' deduction, joint, not maxed out
+  period: 2021
+  input:
+    state_code: IN
+    people:
```
MaxGhenis, line 5: Variables get broadcasted across entities, so this works as long as you don't need to specify the hierarchical structure (e.g. different rent amounts for different people).

This should fix the breaking test, since it looks for entities in all the input first-levels if any entity is detected there.
```suggestion
    rent: 2_000
```

MaxGhenis (approved):
I think this should be good to merge after these and updating the changelog.

openfisca_us/tests/policy/baseline/gov/states/in/tax/income/deductions/in_renters.yaml
MattHJensen: Thanks @MaxGhenis, that fixed the issue with the renters deduction. I'm having a similar but distinct problem now with the nonpublic school deduction tests. Any recommendations there?

nikhilwoodruff: OK, I understand why this is failing: it's because by default, people in tests are put into their own group entities. I think we need to add the following to the unit test:
```yaml
people: ...
tax_units:
  tax_unit:
    members: [child1, child2]
```
Probably worth adding an adult head just to be realistic, though seems like the logic should work without it.

MaxGhenis:
This error is because we switched to a new changelog system. Instead of appending to changelog.yaml (which creates merge conflicts), please take that content and enter it in a new changelog_entry.yaml in the root folder. You'll also need to `git pull upstream master` to ensure it's checking against the latest version.

MattHJensen:
I believe this one is good to merge.

MaxGhenis:
@MattHJensen could you please resolve all comments? I think most are done but there are a couple minor ones left.

openfisca_us/tests/policy/baseline/gov/states/in/tax/income/deductions/in_deductions.yaml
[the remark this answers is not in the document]
MattHJensen: Yea, I replicated the bug I was remembering and test_variables.py does its job well. Removing in_deductions.yaml now.

openfisca_us/tests/policy/baseline/gov/states/in/tax/income/deductions/in_renters.yaml
MattHJensen: Thanks, Nikhil. That got this working.

MattHJensen:
Thanks for pointing those out @MaxGhenis. Good exercise to go through and resolved them all. Done now.

PR #1093, opened 2022-07-14
Add MD CDCC and fix federal CDCC

### LINKED ISSUE #941: Add Maryland CDCC
Matches the federal CDCC at a declining percentage depending on AGI and filing status
https://www.marylandtaxes.gov/tax-credits.php
@ngpsu22

### LINKED ISSUE #1095: Phase out CDCC in a stepped way
It's currently modeled as smooth, but in reality it falls by 1% for each $2,000 of AGI over the phase-out start(s). See [TPC](https://www.taxpolicycenter.org/briefing-book/how-does-tax-system-subsidize-child-care-expenses#:~:text=The%20child%20and%20dependent%20care%20tax%20credit%20(CDCTC)%20provides%20a,for%20two%20or%20more%20dependents).):
image

MaxGhenis:
* Fixes #941
* Fixes #1095
* Adds notebooks for federal and Maryland CDCCs


openfisca_us/variables/gov/states/md/tax/income/credits/md_cdcc.py
```diff
@@ -0,0 +1,35 @@
...
+        # Eligibility is based on AGI.
+        eligible = agi <= p.eligibility.agi_cap[filing_status]
+        # Maximum is a percent of federal.
+        max_cdcc = p.percent * tax_unit("cdcc", period)
+        # Phases out based on filing status.
+        phase_out_start = p.phase_out.start[filing_status]
+        excess = max_(0, agi - phase_out_start)
+        phase_out_increment = p.phase_out.increment[filing_status]
+        # TODO: Floor.
+        phase_out_increments = excess / phase_out_increment
```
nikhilwoodruff, line 26: I think this should be a staircase, rather than a ramp, right?

MaxGhenis: yes, added a floor statement

MaxGhenis:
This looks right for 2020/2022:
image

I'd like to dig a bit more to see if this cliff in 2021 is expected based on refundability or something:
image

@nikhilwoodruff and I also need to identify the sequence/treatment of multiple nonrefundable credits.

fedderw:
> This looks right for 2020/2022: image
> 
> I think this cliff in 2021 is also expected; it's the income limit for MD CDCC refundability: image
> 
> @nikhilwoodruff and I also need to identify the sequence/treatment of multiple nonrefundable credits.

Is [this](https://www.1040.com/pdf/1040/states/MD/2021/MD_MD-EICWK_EIC-worksheet_21.pdf) what you mean?

MaxGhenis:
@fedderw no, though that's helpful for the poverty level credit and we'll need to deal with that as well as another nonrefundable credit. Looking for how to limit all nonrefundable credits together to tax liability. The simplest is to do just that--sum nonrefundable credits and use that while keeping the credits themselves at their full uncapped values--but the forms or law may give more precise guidance.

MaxGhenis:
This is ready for review. I think we can tackle the multiple nonrefundable credits separately.

openfisca_us/variables/gov/states/md/tax/income/credits/md_cdcc.py
```diff
@@ -0,0 +1,37 @@
...
+    unit = USD
+    definition_period = YEAR
+    reference = "https://casetext.com/statute/code-of-maryland/article-tax-general/title-10-income-tax/subtitle-7-income-tax-credits/section-10-716-for-child-care-or-dependent-care"
+
+    def formula(tax_unit, period, parameters):
+        filing_status = tax_unit("filing_status", period)
+        agi = tax_unit("adjusted_gross_income", period)
+        p = parameters(period).gov.states.md.tax.income.credits.cdcc
+        # Eligibility is based on AGI.
+        in_md = tax_unit.household("state_code_str", period) == "MD"
```
fedderw, line 19: Why do we need to include the residency check here? Because of federal AGI is the input?

fedderw (approved):
lgtm (no comment on the fed/IN changes

openfisca_us/variables/gov/states/md/tax/income/credits/md_cdcc.py
MaxGhenis: Passing a MD AGI variable would still give >0 md_cdcc to people outside MD, since it includes zero-earners. We're not consistent around this yet (Nikhil's work on a filter will make it consistent) but I did it here so I could estimate the budgetary impact without zeroing out non-MD household weights.

nikhilwoodruff (approved):

Issue #867, opened 2022-06-04
Add another dividend variable and add missing dividend logic


martinholmer:
The 0.70.3 code defines three variables related to dividends: `qualified_dividend_income`, `ordinary_dividend_income`, and `dividend_income`.  With 0.70.3 logic, entering $100,000 in `qualified_dividend_income` produces zero income tax liability, and entering $100,000 in `ordinary_dividend_income` produces zero income tax liability.  Entering $100,000 in `dividend_income` produces an income tax liability equal to what the tax on non-qualified dividends would be.

The current version of OpenFisca-US does not seem to have dividend variables and logic specified in a useful manner.  Because qualified and non-qualified dividends are taxed in different ways (at different marginal rates, etc.), it is essential to allow model users a way to specify the difference between qualified and non-qualified dividends.

In [TAXSIM35](https://users.nber.org/~taxsim/taxsim35/), the `dividends` variable is qualified dividends and the `otherprop` variable is where users enter non-qualified dividends and other property income that does not get special tax treatment.

It seems to me that OpenFisca-US needs to add a `non_qualified_dividend_income` variable and remove the `dividend_income` variable.  And it needs to correctly define `ordinary_dividend_income` as the **sum** of `qualified_dividend_income` and `non_qualified_dividend_income`, as indicated in the 2019 Form 1040 instructions for line 3a.

<img width="259" alt="Screen Shot 2022-06-04 at 11 32 06 AM" src="https://user-images.githubusercontent.com/12170745/172015497-c967ef1d-0420-4a0d-b3bf-ddae7faa577b.png">

Once those changes in the basic variables are made (plus any needed changes to the code), the following two tests should pass.

First case 1 with qualified dividend income:

```
# OpenFisca-US test file created by hand
# Expected test output from NBER TAXSIM Model v35 (05/18/22) With TCJA

- name: Tax unit with taxsimid 1
  absolute_error_margin: 0.01
  period: 2019
  input:
    people:
      person1:
        is_tax_unit_head: 1
        is_tax_unit_spouse: 0
        is_tax_unit_dependent: 0
        age: 55
        qualified_dividend_income: 100000
    tax_units:
      tax_unit:
        members: [person1]
        premium_tax_credit: 0
    households:
      household:
        members: [person1]
        state_code: TX
  output:
    taxsim_tfica: 0.00
    income_tax: 7263.75
```

Then case 2 with non-qualified dividend income:

```
# OpenFisca-US test file created by hand
[...]

PR #1115, opened 2022-07-16
Add non-qualified dividends and fix ordinary dividends

nikhilwoodruff:
Fixes the first part of #867


nikhilwoodruff:
cc @martinholmer - this should fix that issue. Though at some point I would like to re-implement capital gains from the law to avoid the `dwks...` variables, which I find a bit opaque.

martinholmer:
@nikhilwoodruff said in PR #1115:
this should fix that issue [#867].  Though at some point I would like to re-implement capital gains from the law to avoid the `dwks...` variables, which I find a bit opaque.

That makes sense.  The Schedule D Worksheet variables make sense only if you are writing code that follows the tax forms, but in OpenFisca-US you are writing code that follows the tax legislation.

In any case, thanks for the fix in PR #1115.

MaxGhenis (approved):

PR #1124, opened 2022-07-18
Add Maryland CTC notebook

fedderw:
image

image


changelog_entry.yaml
```diff
@@ -0,0 +1,4 @@
+- bump: minor
```
MaxGhenis, line 1: Since it doesn't add new functionality to the package
```suggestion
- bump: patch
```

MaxGhenis (commented):
Please update the markdown cells in the notebook. It's not producing a budgetary impact because we don't currently impute disability status.

MaxGhenis:
Sorry I wasn't precise - could you keep the budgetary impact cell, and add an explanation for why it's zero?

PR #1135, opened 2022-07-19
Add MD income tax to State income tax

### LINKED ISSUE #1134: Adding `md_income_tax` to `state_income_tax` causes maximum recursion error
#1132 completes what we need to compute Maryland's state income tax. However, the final step of adding `md_income_tax` to `state_income_tax` (along with `ma_income_tax` and `wa_income_tax` currently) causes a maximum recursion error. For example, see errors in the [GH Action runs](https://github.com/PolicyEngine/openfisca-us/runs/7403664481?check_suite_focus=true) testing a commit to #1132 adding it.

Here's the error I get locally when running a notebook for `md_eitc` (notably this doesn't call `state_income_tax` directly or indirectly):
```
RecursionError                            Traceback (most recent call last)
~/opt/anaconda3/envs/py37/lib/python3.7/site-packages/openfisca_tools/hypothetical.py in calc(self, var, period, target, index, map_to, reform)
    280         try:
--> 281             result = self.simulation.calculate(var, period)
    282         except:

~/opt/anaconda3/envs/py37/lib/python3.7/site-packages/openfisca_core/simulations/simulation.py in calculate(self, variable_name, period)
     94         try:
---> 95             result = self._calculate(variable_name, period)
     96             self.tracer.record_calculation_result(result)

~/opt/anaconda3/envs/py37/lib/python3.7/site-packages/openfisca_core/simulations/simulation.py in _calculate(self, variable_name, period)
    124             self._check_for_cycle(variable.name, period)
--> 125             array = self._run_formula(variable, population, period)
    126 

~/opt/anaconda3/envs/py37/lib/python3.7/site-packages/openfisca_core/simulations/simulation.py in _run_formula(self, variable, population, period)
    231         else:
--> 232             array = formula(population, period, parameters_at)
    233 

~/opt/anaconda3/envs/py37/lib/python3.7/site-packages/openfisca_tools/model_api/model_api.py in sum_of_variables(entity, period, parameters)
    331             variable_names = variables
--> 332         return add(entity, period, variable_names)
...
---> 85             return self.simulation.calculate(variable_name, period)
     86 
     87     # Helpers

RecursionError: maximum recursion depth exceeded while calling a Python object
```
[This StackOverflow answer](https://stackoverflow.com/a/15241424/1840471) suggests increasing the recursion limit, but (a) I don't know where we would do that, (b) if we're hitting this, it makes me wonder if something is going wrong more generally, e.g. with the standard deduction, or if not, how we're going to scale to 50 states.

nikhilwoodruff:
Fixes #1134


changelog_entry.yaml
```diff
@@ -0,0 +1,4 @@
+- bump: mino
```
MaxGhenis, line 1:
```suggestion
- bump: minor
```

MaxGhenis (approved):

nikhilwoodruff:
Exclude QBID from MD taxes.

MaxGhenis:
@nikhilwoodruff the failing EITC test passes locally for me. Could you PTAL?

nikhilwoodruff:
@MaxGhenis this is weird. The test fails locally for me if run through `make test`, but `openfisca-us test [path]` passes.

PR #1148, opened 2022-07-21
MD Reduce CTC by federal CTC

### LINKED ISSUE #1140: Reduce MD CTC by the federal CTC
https://casetext.com/statute/code-of-maryland/article-tax-general/title-10-income-tax/subtitle-7-income-tax-credits/section-10-751-effective-until-712026-tax-credit-for-qualified-child

The [MD CTC code](https://github.com/PolicyEngine/openfisca-us/blob/master/openfisca_us/variables/gov/states/md/tax/income/credits/ctc/md_ctc.py) currently omits this logic:
>(c) The amount of the credit allowed under subsection (b) of this section for a qualified child shall be reduced, but not below zero, by the amount of any federal child tax credit claimed against the federal income tax for the qualified child under § 24 of the Internal Revenue Code.

fedderw:
fixes #1140


openfisca_us/variables/gov/states/md/tax/income/credits/ctc/md_ctc_without_federal.py
```diff
@@ -0,0 +1,26 @@
+from openfisca_us.model_api import *
+
+
+class md_ctc_without_federal(Variable):
+    value_type = float
+    entity = TaxUnit
+    label = (
+        "MD CTC without ssubtracting the value of the federal child tax credit"
```
MaxGhenis, line 8:
```suggestion
        "MD CTC without subtracting the value of the federal child tax credit"
```

MaxGhenis (commented):
Could you add a reference to a tax form? I think we need to know what it's specifically pointing to in terms of how the nonrefundable CTC is counted.

MaxGhenis:
The yaml tests pass, but we're getting a recursion error in the taxsim tests. Not sure why yet.

nikhilwoodruff:
I'd try something like: 

openfisca-us test taxsim_test.yaml -v > output.txt

This should print the verbose output to a text file (so it doesn't get clipped by the terminal history size), and the verbose output should at some point print out the computation tree, which should identify the loop.

MaxGhenis:
@nikhilwoodruff it's failing on `openfisca_us/tests/microsimulation/test_against_taxsim.py`. Is there a straightforward way to extract a failing test case from that?

nikhilwoodruff:
Hmm. Not sure, but one approach I'd try is a binary search-esque method: try running the microsim to the net income variable used in the test. If that fails, guess a variable roughly halfway down the computation tree and try that, refining until you can find the variable that breaks it. In future we should figure out a way to do this more straightforwardly.

MaxGhenis:
We've now sorted this in other PRs

PR #1149, opened 2022-07-21
Test misnamed class

MattHJensen:
This differs from #1084 only by way of a typo in the class name of [..._us/variables/gov/states/in/tax/income/deductions/in_homeowners_property_tax_deduction.py](https://github.com/PolicyEngine/openfisca-us/pull/1149/files#diff-3904d7b18e1f5d84bc4f64a6029bf8a3c0cbc728856e3ce0382a36ea6cb6913b)


MaxGhenis:
Why not make the change in #1084?

PR #1166, opened 2022-07-24
IN agi tax

MattHJensen:


openfisca_us/tests/policy/baseline/gov/states/in/tax/income/taxes/in_agi_tax.yaml
```diff
@@ -0,0 +1,15 @@
...
+  output:
+    in_agi_tax: 1_000 * 0.0323
+
+- name: 10,000 AGI
+  period: 2021
+  input:
+    state_code: IN
+    in_agi: 10_000
+  output:
+    in_agi_tax: 1_000 * 0.0323
```
MaxGhenis:
```suggestion
    in_agi_tax: 10_000 * 0.0323
```

MaxGhenis (approved):

MattHJensen:
This one is ready for review. Thanks @MaxGhenis and @nikhilwoodruff.

openfisca_us/parameters/gov/states/in/tax/income/taxes/agi/rates.yaml
```diff
@@ -0,0 +1,11 @@
+description: IN adjusted gross income tax rate.
+values:
+  2021-01-01: 0.0323
+metadata:
+  period: year
+  unit: /1
+  name: in_agi_tax_rate
+  label: IN agi tax rate
```
MaxGhenis, line 8:
```suggestion
  label: IN AGI tax rate
```

openfisca_us/parameters/gov/states/in/tax/income/taxes/agi/rates.yaml
```diff
@@ -0,0 +1,11 @@
+description: IN adjusted gross income tax rate.
+values:
+  2021-01-01: 0.0323
```
MaxGhenis, line 3: May as well since it's in the legal code
```suggestion
  2014-01-01: 0.034 # "For taxable years beginning before January 1, 2015"
  2015-01-01: 0.033
  2017-01-01: 0.0323
```

openfisca_us/parameters/gov/states/in/tax/income/taxes/agi/rates.yaml
```diff
@@ -0,0 +1,11 @@
+description: IN adjusted gross income tax rate.
```
MaxGhenis, line 1: Nit but I think we usually go with `rate` instead of `rates` when it's a single rate, could you rename?

MaxGhenis (approved):

openfisca_us/parameters/gov/states/in/tax/income/taxes/agi/rates.yaml
MattHJensen: I saw that there were other "rates" where the rate only varied across years. I'm just focussing on 2021 right now but I expect Indiana's rate has changed in the past will change in the future. Ok for me to use `rates` now so we don't have to change the filename when that stuff is added?

MattHJensen: Also just saw your commit adding the previous year. Should I look back for the other `rates` I found and rename it to `rate` since it only varies across years?

MattHJensen: Will change.

PR #1198, opened 2022-07-29
Add new york taxable income

Aleida24:
This PR adds a taxable income variable for new york state tax.

- [ ] `make format && make documentation` has been run.

# New variable

- [ ] Label field added
- [ ] Documentation field added
- [ ] Unit field added
- [ ] Default value field added if relevant
- [ ] Variable name follows conventions
- [ ] Unit test(s) added
- [ ] Integration test(s) added if relevant
- [ ] Issues this PR fixes linked

## What's changed

Description of the changes here.

# Bug fix

- [ ] Regression test added
- [ ] Regression test passing

Description of how this fix works goes here. Link any issues this PR fixes.


changelog_entry.yaml
```diff
@@ -0,0 +1,4 @@
+- bump: minor
```
nikhilwoodruff, line 1: @Aleida24 just found the issue: the changelog entry just needs to be a list item, rather than an object.

PR #1236, opened 2022-08-09
Fix qualified dividends bug

### LINKED ISSUE #1220: US tax difference for case with qualified_dividend_income
The case described below has been generated using the **r21** **US** sample and adding random `qualified_dividend_income` amounts.  The **r21** sample (without `qualified_dividend_income` amounts) produces no differences between OpenFisca-US and TAXSIM35 (and the US Tax Analyzer).

But in the case below, when adding $20,000 in `qualified_dividend_income`, we get an US income tax difference of $7400.   First the YAML test file containing the case, then the test results, and finally a discussion of the income tax difference.

Here is the test case:
```
# OpenFisca-US test file derived from f21.its.csv and f21.ota.csv files
- name: Tax unit with taxsimid 5832 from f21.its.csv
  absolute_error_margin: 0.01
  period: 2021
  input:
    people:
      person1:
        is_tax_unit_head: 1
        is_tax_unit_spouse: 0
        is_tax_unit_dependent: 0
        age: 53
        employment_income: 159000
        qualified_dividend_income: 20000
        taxable_interest_income: 3000
        social_security: 20000
        real_estate_taxes: 8000
        interest_expense: 2000
        ssi: 0
        state_supplement: 0
        wic: 0
    spm_units:
      spm_unit:
        members: [person1]
        snap: 0
        tanf: 0
    tax_units:
      tax_unit:
        members: [person1]
        premium_tax_credit: 0
    households:
      household:
        members: [person1]
        state_code: TX
  output:
    taxsim_tfica: 11159.10
    income_tax: 37091.00
```

Here are the results for the above test:
```
=============================== FAILURES ===================================
_____________________________ test session _________________________________
/Users/mrh/work/OpenFisca-US/US/f21-id5832.test.yaml:
  Test 'Tax unit with taxsimid 5832 from f21.its.csv':
    income_tax@2021: [44491.] differs from 37091.0
      with an absolute margin [7400.] > 0.01
======================= short test summary info ============================
FAILED f21-id5832.test.yaml::
========================== 1 failed in 1.10s ===============================
Trace output is in file: f21-id5832.test.trace
```

The models agree on AGI ($199,000) and taxable income ($186,450).
But yet the models disagree on income taxes before credits with OpenFisca-US getting $44,491 and TAXSIM35 (and USI Tax Analyzer) getting $37,091.  This $7400 difference also shows up for income tax because both models agree that this case gets no tax credits.

The trace output shows that OpenFisca-US is calculating a `capital_gain_tax` of $4000, which differs from the $3000 calculated by the other two models.  The other two models think the taxable income of $186,450 is split into two parts: (a) $166,450 that is taxed at regular rates (yielding a liability of $34,091) and (b) $20,000 that is taxed at 15 percent (yielding a liability of $3000).  The other two models combine those two liabilities to get a total US income tax liability of $37,091.   @nikhilwoodruff, do you think the other two models have made a mistake in calculating taxes in this test case?   If so, can you explain exactly what is wrong with the tax logic in TAXSIM35?  Perhaps filling out 2021 Form 1040 Schedule D for this test case would reveal where the problem is.

nikhilwoodruff:
Fixes #1220 . The issue was a bug in which `net_capital_gain` had multiple formulas, and the last one was incorrect. Removed.

cc @martinholmer (I've added your test case, which now passes).

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    fixed:
+    - A bug causing qualified dividends to not be counted as 'net capital gain'.
--- a/openfisca_us/tests/policy/baseline/taxsim/misc/qualified_dividends.yaml
+++ b/openfisca_us/tests/policy/baseline/taxsim/misc/qualified_dividends.yaml
@@ -1,0 +1,36 @@
+# OpenFisca-US test file derived from f21.its.csv and f21.ota.csv files
+- name: Tax unit with taxsimid 5832 from f21.its.csv
+  absolute_error_margin: 0.01
+  period: 2021
+  input:
+    people:
+      person1:
+        is_tax_unit_head: 1
+        is_tax_unit_spouse: 0
+        is_tax_unit_dependent: 0
+        age: 53
+        employment_income: 159000
+        qualified_dividend_income: 20000
+        taxable_interest_income: 3000
+        social_security: 20000
+        real_estate_taxes: 8000
+        interest_expense: 2000
+        ssi: 0
+        state_supplement: 0
+        wic: 0
+    spm_units:
+      spm_unit:
+        members: [person1]
+        snap: 0
+        tanf: 0
+    tax_units:
+      tax_unit:
+        members: [person1]
+        premium_tax_credit: 0
+    households:
+      household:
+        members: [person1]
+        state_code: TX
+  output:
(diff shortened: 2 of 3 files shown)


martinholmer:
@nikhilwoodruff,  Thanks for the fix in #1236.  I'm glad it turned out to be easy.

PR #1269, opened 2022-08-17
2022 value of EITC joint bonus for couples without children

### LINKED ISSUE #1267: Missing 2022 US EITC parameter
The [`joint_bonus.yaml`](https://github.com/PolicyEngine/openfisca-us/blob/master/openfisca_us/parameters/gov/irs/credits/eitc/phase_out/joint_bonus.yaml) EITC parameter file is missing the 2022 value for couples without children.   The referenced IRS document says the missing value should be $6130.

MaxGhenis:
Fixes #1267

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    added:
+      - 2022 value of EITC joint bonus for couples without children.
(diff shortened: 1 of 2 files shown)


martinholmer:
@MaxGhenis,  Thanks for the quick response to issue #1267.

PR #1276, opened 2022-08-20
General refactor

### LINKED ISSUE #1176: Use `default` arg to `select` instead of a dummy `True` condition
In some instances, we use `select` with a dummy `True` condition list item, to capture the default. Let's use the optional `default` arg to [`numpy.select`](https://numpy.org/doc/stable/reference/generated/numpy.select.html) instead.

Example:
https://github.com/PolicyEngine/openfisca-us/blob/aaf2616aeee5187266f9ff0a8d2b4156291711f9/openfisca_us/variables/household/demographic/person/postpartum/count_days_postpartum.py#L14-L25

Replacement:
```python
        return select(
            [under_60_days, under_12_months],
            [0, 60],
            default=np.inf
        )
```

[Time tests](https://colab.research.google.com/drive/11dJtJSLPxskoEow0QEsiMI7xaCyYOQrw#scrollTo=RnWjLYrWxTr5) show this improves performance of the select operation by about 10%.

MaxGhenis:
* Fixes #1176 by using `default` arg to `select`
* Add `defined_for` to some variables for performance
* Rename .py files to match variable name
* Move taxcalc variable aliasing to yaml mapping file


MaxGhenis:
Not sure what's driving this error, seems like it may have something to do with `maximum_state_supplement` and the `defined_for` filter.

MaxGhenis:
I think a recent failure was due to variable aliases not treating masks correctly, so I removed EITC aliases.

MaxGhenis:
Seems to still be some issues with the masks. Going to pause on this PR until @nikhilwoodruff returns next week.

openfisca_us/variables/gov/states/md/tax/income/credits/eitc/federal_eitc_without_age_minimum.py
```diff
@@ -9,6 +9,7 @@ class federal_eitc_without_age_minimum(Variable):
     documentation = "The federal EITC with the minimum age condition ignored."
     definition_period = YEAR
     reference = "https://casetext.com/statute/code-of-maryland/article-tax-general/title-10-income-tax/subtitle-7-income-tax-credits/section-10-704-effective-until-6302023-for-earned-income"
+    defined_for = StateCode.MD
```
nikhilwoodruff, line 12: Yeah it's an unconventional edge case I didn't think about when writing the original logic. Probably a fix in openfisca-tools (filed in https://github.com/PolicyEngine/openfisca-tools/issues/74) which might get a bit complex: in-built support for simulation-editing formulas in Core would reduce the complexity by a lot. But in the meantime a quick fix is to avoid using `defined_for` for variables whose formulas create a new simulation.

MaxGhenis:
I just tried to merge this with master, and it seems like too many changes with the renames and all to work. @nikhilwoodruff any ideas on reviving this, or should I close and pull out changes into a new PR?

PR #1315, opened 2022-09-01
Fix breakdown in BI parameters

### LINKED ISSUE #1314: Fix breakdown in BI parameters
Not provided as a list which makes them display wrong in PE

MaxGhenis:
Fixes #1314

### DIFF
--- a/openfisca_us/parameters/contrib/ubi_center/basic_income/agi_limit/amount.yaml
+++ b/openfisca_us/parameters/contrib/ubi_center/basic_income/agi_limit/amount.yaml
@@ -12,6 +12,7 @@ WIDOW:
 metadata:
   unit: currency-USD
   period: year
-  name: bi_agi_limit
+  name: bi_agi_limit_amount
   label: Basic income AGI limit
-  breakdown: filing_status
+  breakdown:
+    - filing_status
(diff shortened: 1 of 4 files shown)


openfisca_us/parameters/contrib/ubi_center/basic_income/agi_limit/amount.yaml
```diff
@@ -12,6 +12,7 @@ WIDOW:
 metadata:
   unit: currency-USD
   period: year
-  name: bi_agi_limit
+  name: bi_agi_limit_amount
```
MaxGhenis, line 15: I had to change this because it broke in PolicyEngine to have `bi_agi_limit` and `bi_agi_limit_in_effect`. Activating the parameter menu crashed the screen. @nikhilwoodruff Is there some `_in_effect` suffix that gets used for something that it's clobbering?

nikhilwoodruff: I think it's because of the fact that we convert `.` to `_`. This name would be ambigous: what if there were a structure named `bi_age_limit/in_effect.yaml`, etc. I'd favour something like `bi_age_limit_amount` here.

PR #1317, opened 2022-09-01
Add NY supplemental income tax

nikhilwoodruff:
Got the legislation decoded and working for single filers in 2022, with the rest of the parameters added as placeholders.


nikhilwoodruff:
OK, final parameters added.

openfisca_us/parameters/gov/states/ny/tax/income/supplemental/recapture_base/head_of_household.yaml
```diff
@@ -0,0 +1,37 @@
...
+      2022-01-01: 5_000_000
+    amount:
+      2022-01-01: 48_337
+  - threshold:
+      2022-01-01: 25_000_000
+    amount:
+      2022-01-01: 80_837
+
+metadata:
+  threshold_unit: currency-USD
```
MaxGhenis, line 29: do you need `type: single_amount`?

openfisca_us/parameters/gov/states/ny/tax/income/supplemental/agi_limit/max.yaml
```diff
@@ -0,0 +1,11 @@
+description: Filers with AGI over this limit pay a flat rate on taxable income.
+SINGLE:
```
MaxGhenis, line 2: Is this intended over `values`? If so it needs a `breakdown`.

openfisca_us/parameters/gov/states/ny/tax/income/supplemental/agi_limit/max.yaml
```diff
@@ -0,0 +1,11 @@
...
+  2022-01-01: 25_000_000
```
MaxGhenis, line 3: I don't see this value in the linked reference, could you point to a search term I can use?

openfisca_us/parameters/gov/states/ny/tax/income/supplemental/recapture_base/head_of_household.yaml
nikhilwoodruff: It's automatically inferred by OpenFisca because of the `amount` usage, but I agree explicit is better so added.

openfisca_us/parameters/gov/states/ny/tax/income/supplemental/agi_limit/max.yaml
nikhilwoodruff: Ah it was the wrong one - should point to 601 instead. Fixed.

PR #1366, opened 2022-09-14
Mo tax testing

nmrodelo:
Adding example notebooks, TAXSIM testing, and any changes to variables as necessary.


MaxGhenis (approved):

MaxGhenis:
Codecov is lower because there's some property tax credit code in there that's not being run, pending some more information from MO DOR. We'll hopefully get that detail soon to flesh out the full credit, otherwise we'll remove it to keep codecov clean. Merging for now.

PR #1380, opened 2022-09-18
Add contrib tax credit for "filling in" the gap of the EITC+CTC

### LINKED ISSUE #1377: Add contrib tax credit for "filling in" the gap of the EITC+CTC
From Matt Bruenig of People's Policy Project

https://www.peoplespolicyproject.org/projects/extending-child-benefits/
image

MaxGhenis:
Fixes #1377

@nikhilwoodruff any ideas on making it just fill in the left side? Basically want a threshold based on whether their income exceeds the income at which EITC+CTC == EITCmax+CTCmax.
image


nikhilwoodruff:
I'd subtract `eitc_reduction` and `ctc_reduction` in this credit's formula.

MaxGhenis:
@nikhilwoodruff I did that, but realized this makes it only an EITC fill-in, since CTC is defined as `ctc_maximum - ctc_reduction`. I think the CTC nonrefundability is going to make this more complicated than expected.

nikhilwoodruff:
OK think I solved the issue, LMK if it makes sense

MaxGhenis:
Thanks @nikhilwoodruff, that gets us closer but I think we need some more logic to get `max(eitc+ctc)`, which is less than `max(eitc) + max(ctc)`, e.g. this example from a family of four.
image

I don't think there's a closed-form solution to this; we'll need to run through their household's earnings to determine it. Is this possible?

@bruenig could you confirm that's your intent with the policy? And could you confirm how you envision states dealing with the non-refundable CTC? Currently we're basing the "fill-in credit" on the non-refundable CTC capped at taxes after all other non-refundable credits (plus the refundable CTC). Thanks.

nikhilwoodruff:
OK makes sense. Though I think one way we could do it is:
1. Create a branched simulation
2. Nullify intermediate and output CTC variables
3. Reform parameters to eliminate phase-ins
4. Calculate the difference

MaxGhenis:
Closing this for now as Matt Bruenig hasn't specified a closed-form definition of the proposal (one that doesn't require looping through all earnings levels for the filer)

MaxGhenis:
For posterity: Bruenig published this idea at https://www.peoplespolicyproject.org/projects/extending-child-benefits/

PR #1418, opened 2022-10-06
fix md non_refundable_eitc childless max parameter

fedderw:
Bug fix where the wrong parameter name was used
image


MaxGhenis:
Could you please add a `md_non_refundable_eitc.yaml` unit test file to https://github.com/PolicyEngine/openfisca-us/tree/master/openfisca_us/tests/policy/baseline/gov/states/md/tax/income/credits/eitc?

fedderw:
> d_non_refundable_eitc.yaml

Are they not covered here? I think this might be a parameter name issue, rather than what I thought originally
image
image

MaxGhenis:
I don't see a unit test for the `md_non_refundable_eitc` variable (which would have caught this bug initially).

PR #1429, opened 2022-10-12
person level tax calculations for MO

nmrodelo:
update tax variables to calculate at person level to account for MFJ calculation in MO


MaxGhenis:
Closing in favor of #1441, since some of these commits affected other files and we haven't found how to easily revert them

PR #1441, opened 2022-10-15
port work from mo_married_filing_jointly

nmrodelo:
copy of mangled pull request "person level tax calculations for MO"


openfisca_us/variables/gov/states/mo/tax/income/adjusted_gross_income/mo_adjusted_gross_income.py
```diff
@@ -13,9 +13,7 @@ class mo_adjusted_gross_income(Variable):
     )
     defined_for = StateCode.MO
 
-    def formula(tax_unit, period, parameters):
-        agi = tax_unit("adjusted_gross_income", period)
-        subtractions = tax_unit(
-            "mo_qualified_health_insurance_premiums", period
-        )
+    def formula(person, period, parameters):
+        agi = person("irs_gross_income", period)
```
MaxGhenis, line 17: Could you also list a reference for this and leave a comment confirming that the result is capped at zero? Federal AGI can be negative.

```suggestion
        gross_income = person("irs_gross_income", period)
```

MaxGhenis (approved):
Made some small changes and left one final request, then good to merge. Thanks!

openfisca_us/variables/gov/states/mo/tax/income/adjusted_gross_income/mo_adjusted_gross_income.py
nmrodelo: that should probably be removed then, good catch, thanks

PR #1485, opened 2022-11-06
Make formula for CTC consistent across years (including 2021 with ARPA)

### LINKED ISSUE #638: Use consistent formula across years for CTC
Let's replace this with a single formula and setting the increase to zero pre- and post-2021. This will enable simulating reforms with the young child bonus in the current year (including adding the parameter to PolicyEngine), avoid confusion when looking at the parameter file, and simplify the code.

https://github.com/PolicyEngine/openfisca-us/blob/c293b6e16f9c45ca64696d9522ead93e9c05d4d8/openfisca_us/variables/irs/credits/ctc/maximum/individual/child.py#L17-L31

https://github.com/PolicyEngine/openfisca-us/blob/c293b6e16f9c45ca64696d9522ead93e9c05d4d8/openfisca_us/parameters/irs/credits/ctc/child/young/increase.yaml#L1-L3

MaxGhenis:
* Fixes #638 by (a) adding flags for whether the CTC is fully refundable and the ARPA phase-out is in effect, triggering different logic, and (b) parameterizing some other conditionals by setting values to zero before/after ARPA
* Cleans up some other CTC and EITC stuff
* Removes some newly failing MO taxsim tests - these seemed a bit brittle before, so it's possible they relied on old logic
* Removes a divide by zero warning in the RRC


policyengine_us/parameters/gov/irs/credits/ctc/adult_dependent_amount.yaml
```diff
@@ -1,11 +1,14 @@
...
+  0000-01-01: 0
+  2018-01-01: 500
+  2026-01-01: 0
 metadata:
   unit: currency-USD
   period: year
   name: ctc_adult_dependent
   label: Child tax credit for adult dependents
   reference:
+    # TCJA only.
```
nikhilwoodruff, line 12: Should we add a "comment" field in Core?

nikhilwoodruff (approved):

policyengine_us/parameters/gov/irs/credits/ctc/adult_dependent_amount.yaml
MaxGhenis: Yes, and/or a start/end period? This would duplicate the functionality of providing references for a specific period, but reduce code duplication when there are breakdowns. Alternatively, flipping the order of dates and breakdowns...

Issue #1506, opened 2022-11-15
Need to reassess CPS-to-PolicyEngineUS income variable mapping


martinholmer:
In the [cps.py](https://github.com/PolicyEngine/policyengine-us/blob/master/policyengine_us/data/datasets/cps/cps.py) file, the following code assigns CPS income variables to PolicyEngineUS variables:
```
def add_personal_income_variables(cps: h5py.File, person: DataFrame):
    """Add income variables.

    Args:
        cps (h5py.File): The CPS dataset file.
        person (DataFrame): The CPS person table.
    """
    cps["employment_income"] = person.WSAL_VAL
    cps["interest_income"] = person.INT_VAL
    cps["self_employment_income"] = person.SEMP_VAL
    cps["farm_income"] = person.FRSE_VAL
    cps["dividend_income"] = person.DIV_VAL
    cps["rental_income"] = person.RNT_VAL
    cps["social_security"] = person.SS_VAL
    cps["unemployment_compensation"] = person.UC_VAL
    cps["pension_income"] = person.PNSN_VAL + person.ANN_VAL
    cps["alimony_income"] = (person.OI_OFF == 20) * person.OI_VAL
    cps["tanf_reported"] = person.PAW_VAL
    cps["ssi_reported"] = person.SSI_VAL
    cps["pension_contributions"] = person.RETCB_VAL
    cps[
        "long_term_capital_gains"
    ] = person.CAP_VAL  # Assume all CPS capital gains are long-term
    cps["receives_wic"] = person.WICYN == 1
```

The above code raises several questions:
* How should interest be split between taxable and tax-exempt?
* Should all self-employment income be **not** qualified business income?
* How should dividends be split between qualified and non-qualified?
* How should pensions be split between taxable and tax-exempt?
* How should capital gains be split between short-term and long-term?

MaxGhenis:
In the near term, we can use averages from administrative totals like we do for dividends.

Longer term, we plan to allocate using our synthimpute package for random forest-based quantile regression, pulling data from the PUF.

martinholmer:
@MaxGhenis said:
> In the near term, we can use averages from administrative totals like we do for dividends.

OK, but what about the **immediate** term?   

How about fixing the errors by making some simple assumptions just like is currently being done with capital gains?  This would mean making the following assumptions:

* All CPS interest is taxable
* All CPS dividends are qualified
* All CPS pensions are taxable
* All CPS capital gains are long-term (as assumed now)

[...]

PR #1502, opened 2022-11-15
Simplify calculation of dividend income from qualified and non-qualified dividends

martinholmer:
Fixes #1221.

PR #1507, opened 2022-11-16
Eliminate errors in CPS add_personal_income_variables function

martinholmer:
This pull request is a first step in resolving issue #1506.  It eliminates existing errors (such as no CPS interest income being included in AGI or taxable income) and anticipates further development of the CPS data preparation as discussed in issue #1506.  

This pull request passes all the tests (`make test`) and has the desired code style (`make format`).

This pull request should be merged before merging #1502.


martinholmer:
@MaxGhenis,  I've moved the four `*_fraction` parameters from the `cps.py` file to the `income_parameters.yaml` file, which is located in same directory as the `cps.py` file.

MaxGhenis:
Per our conversation, since this resolves inaccuracies in the current model, I'm merging this as-is. We can return to it using the standard parameters instead.

Issue #1538, opened 2022-12-04
Resolve US discrepancies against TAXSIM35 for 2021


martinholmer:
In the table below, the issues blocking progress are:
#1571 (US) and a TAXSIM35 QBID bug that has been reported.

```
Number of tax units (out of random samples of 100,000) with
an income tax difference between PolicyEngineUS and TAXSIM35
[sample assumptions are described below this table]
[what a 'difference' means is explained below this table]
--------------------------------------------------------------
: MODEL VERSIONS :   :  MODEL DIFFERENCES  :   : DIFF CAUSES :
PolEngUS  TAXSIM35   REGION SAMPLE NUM.DIFFS   RELEVANT.ISSUES
--------------------------------------------------------------
0.188.4   12/08/22       US    a21     15709   #1571
0.188.4   12/08/22       US    b21     10566   #1571
0.188.4   12/08/22       US    c21     21865   #1571, others
0.188.4   12/08/22       US    d21         0   NONE
0.188.4   12/08/22       US    e21         0   NONE
0.188.4   12/08/22       US    f21     13106   TAXSIM35 bug
--------------------------------------------------------------
```

_Sample assumptions:_

**a21**: 2021 tax units consisting of married couples and single individuals having from zero to four dependents of various ages; with tax units having various types of expenses [rent, local property taxes, mortgage interest, and child care]; and with tax units having various types of incomes [wages, qualified dividends, non-qualified dividends, taxable interest, short-term capital gains (not losses), long-term capital gains (not losses), taxable pension benefits, and social security benefits].

**b21**: same as the **a21** sample except adds self-employment income from a non-qualified business.

**c21**: same as the **a21** sample except adds self-employment income from a qualified business with some being a SSTB and others not being a SSTB.

**d21**: same as the **a21** sample except there are no child care expenses.

**e21**: same as the **b21** sample except there are no child care expenses.

**f21**: same as the **c21** sample except there are no child care expenses.

_Difference definition:_

A difference is generally defined to mean taxes from the two models being more than one cent apart.  The following tests use a definition of difference that is larger than one cent:

**b21** : three cents (PolicyEngine single-precision rounding error)
**c21** : three cents (PolicyEngine single-precision rounding error)
**e21** : three cents (PolicyEngine single-precision rounding error)
**f21** : three cents (PolicyEngine single-precision rounding error)

martinholmer:
[...]

PR #1540, opened 2022-12-04
Remove complex pass-through reform

### LINKED ISSUE #1539: Presence of a complex reform causes US income tax inaccuracies
One of the guiding principles of the PolicyEngine-US project is to avoid adding complex reforms to the code base.  Including complex reforms makes the code more complicated, and hence, more difficult to maintain.  The preferred approach is to use the [powerful capability for specifying complex reforms](https://openfisca.org/doc/coding-the-legislation/reforms.html) that is part of the PolicyEngine-Core framework.  This approach keeps legislative reforms (which come and go) separate from the code base that describes legislative history.

The validation testing against TAXSIM35 using the c21 sample that is described in issue #1538, highlights the pitfalls of including complex reforms in the code base.  The c21 sample has more than 90% of its tax units having a difference with TAXSIM35, with the differences being as large as one thousand dollars.   The complex reform included in the code base that causes all these differences involves parameters and code to allow reforms that tax pass-through income using different tax bracket rates and thresholds.

The following simple test case illustrates these tax differences:
```
- name: Single taxpayer with both wage and self-employment income
  absolute_error_margin: 0.01
  period: 2021
  input:
    people:
      person1:
        is_tax_unit_head: 1
        age: 50
        employment_income: 90_000
        self_employment_income: 100_000
        business_is_qualified: false
        business_is_sstb: false
        ssi: 0  # not in TAXSIM35
        state_supplement: 0  # not in TAXSIM35
        wic: 0  # not in TAXSIM35
    spm_units:
      spm_unit:
        members: [person1]
        snap: 0  # not in TAXSIM35
        tanf: 0  # not in TAXSIM35
    tax_units:
      tax_unit:
        members: [person1]
        premium_tax_credit: 0  # not in TAXSIM35
    households:
      household:
        members: [person1]
        state_code: TX  # no state income tax
  output:  # expected results from online TAXSIM35 12/02/22 version
    taxsim_tfica: 16_110.35
    adjusted_gross_income: 185_387.33
    taxable_income: 172_837.33
    income_tax: 36_134.94
```
Here are the test results for this case:
```
================================= FAILURES ===================================
_______________________________ test session _________________________________
/Users/mrh/work/Policy-Engine-US/US/pt-bug.yaml:
  Test 'Single taxpayer with both wage and self-employment income':
    income_tax@2021: [36514.445] differs from 36134.94140625
                     with an absolute margin [379.5039] > 0.01
========================= short test summary info ============================
FAILED pt-bug.yaml::
============================ 1 failed in 0.92s ===============================
```

martinholmer:
Fixes #1539.

The first commit adds a test that fails.  The added test case is an example of the many c21 test failures described in issue #1538.

Subsequent commits fix the problem that causes the test failure, and then removes the complex pass-through reform (that is the cause of the test failures) and simplifies the regular income tax calculation logic.

The changes in this pull request make all the c21 test failures in #1538 disappear.


martinholmer:
The number of lines not covered by tests is unchanged by pull request #1540, but because the pull request simplifies the code (eliminating 31 lines on net), the percent of lines covered by the tests drops by a tiny amount.

MaxGhenis (approved):
Thanks Martin! Just replaced the big number with `.inf` and merging.

martinholmer:
@MaxGhenis,  Thanks for pointing out the YAML accepts `.inf` as a number.
It does not accept `3e33` as a number.  I guess you need to add a decimal point and maybe even a plus after the `e`, as in `3.3e+33`.

PR #1547, opened 2022-12-12
Metadata additions

nikhilwoodruff:


martinholmer:
@nikhilwoodruff,  Why did you merge PR #1547 without running the tests?

Skipping the tests is at variance with the the contributors guide and pushes your problems on to others (as described in issue #1549 which is blocking pull request #1548).

MaxGhenis:
@martinholmer the tests passed in CI because it fetched the latest policyengine-core. It also would have passed on Nikhil's local machine. They just didn't on your machine because the version wasn't pinned. I'm not sure if pinning to the patch version number is overkill for this, but we'll try it for now.

nikhilwoodruff:
@martinholmer thanks for identifying in #1549. Usually I think it's good practice to pin dependencies at minor versions, so that patches can  be pulled in automatically. And the tests did pass at the time here, but clearly the patch update to Core that happened in between this PR being merged and you finding the error should have been a minor update. I think hard pinning to the patch version of Core should prevent this from happening again.

PR #1566, opened 2022-12-15
Remove unneeded taxsim35_emulation statements from tests

### LINKED ISSUE #1565: Need to remove unneeded use of taxsim35_emulation
The tests have a number of unnecessary uses of the following input statement:
```
    gov.contrib.nber.taxsim35_emulation: true
```

Most of these have been added by me, but I see now that unnecessary setting of the `emulation` parameter to `true` creates an incorrect impression that the test will fail without `emulation`.  Leaving that kind of impression is misleading and is likely to create confusion.

martinholmer:
Fixes #1565.


MaxGhenis (approved):
Love it, thanks Martin

PR #1572, opened 2022-12-17
Align cliff and MTR computations

### LINKED ISSUE #1564: Align MTR and cliff computations
Currently we compute MTRs with a delta of $1,000 (hard-coded), and compute cliffs with a delta of $2,000 (parameterized). The adult parameters are also both set to 1. Can we align these to use the same parameters?

MaxGhenis:
Fixes #1564


martinholmer:
@MaxGhenis,  Give recent changes to the `marginal_tax_rate` and `cliff_gap` variables, where does the very old, but still open, PR #1572 stand?

PR #1576, opened 2022-12-18
Added Ohio tax rates

### LINKED ISSUE #1500: Ohio tax bracket
image
https://tax.ohio.gov/static/forms/ohio_individual/individual/2021/pit-it1040-booklet.pdf#page=31
https://codes.ohio.gov/ohio-revised-code/section-5747.02 A3

dtlawhon:
Fixes #1500


policyengine_us/parameters/gov/states/oh/tax/income/rates/agi.yaml
```diff
@@ -0,0 +1,43 @@
...
+      0000-01-01: 110_651
+    base_amount: 
+      2021-01-01: 3_123.05
+    rate: 
+      2021-01-01: 0.0399
+metadata:
+  threshold_unit: currency_USD
+  threshold_period: year
+  type: marginal_rate
+  name: oh_agi_income_tax_rate
```
MaxGhenis, line 37: Since it's levied on AGI minus exemptions - in most states this would be called taxable income, so maybe the form clarifies
```suggestion
  name: oh_income_tax_rate
```

policyengine_us/parameters/gov/states/oh/tax/income/rates/agi.yaml
```diff
@@ -0,0 +1,43 @@
...
+  label: OH AGI Income Tax Rate
+  reference:
```
MaxGhenis, line 39: Could you add the latest tax form as a second reference?

policyengine_us/parameters/gov/states/oh/tax/income/rates/agi.yaml
```diff
@@ -0,0 +1,43 @@
...
+  - threshold:
+      0000-01-01: 0
+    base_amount: 
+      0
+    rate:
+      2021-01-01: 0
+  - threshold: 
+      0000-01-01: 25_001
+    base_amount:
+      2021-01-01: 346.16
```
MaxGhenis, line 12: We don't need `base_amount` since it's computed mechanically from the rate structure. You can remove all those values.

policyengine_us/parameters/gov/states/oh/tax/income/rates/agi.yaml
```diff
@@ -0,0 +1,43 @@
...
+    rate:
+      2021-01-01: 0.02765
```
MaxGhenis, line 14: Is it 2021? I can't tell from the legal code, which AFAICT doesn't specify a year and only says it was last updated in 2022. Tax form should clarify.

policyengine_us/parameters/gov/states/oh/tax/income/rates/agi.yaml
```diff
@@ -0,0 +1,43 @@
+description: OH individual tax on nonbusiness income
```
MaxGhenis, line 1: Since these are main rates, let's make this file `income/rates.yaml`

MaxGhenis (commented):
Thank you for contributing to PolicyEngine, @dtlawhon! Ohio is the most populous state we don't model yet, so we really appreciate you kicking off their tax code.

In addition to these comments, could you please populate `changelog_entry.yaml`? You can see a recently merged PR for an example of how that works.

This PR will then be good to merge, though FYI to fully resolve #1500 we'd also want to add three more files:
1. `variables/gov/states/oh/tax/income/oh_taxable_income.py`, a variable representing taxable income (or perhaps AGI less exemptions if that's what they call it) with no formula (we could fill that in in another PR)
2. `variables/gov/states/oh/tax/income/oh_income_tax_before_credits.py`, a variable that runs `parameters(period).gov.states.oh.tax.income.rates.calc(oh_taxable_income)`
3. `tests/policy/baseline/gov/states/oh/tax/income/oh_income_tax_before_credits.yaml`, a file representing a couple unit tests

Feel free to ask questions here or on [Slack](https://join.slack.com/t/policyengine-group/shared_invite/zt-1i9jpin5f-JLe~ob__MAK4OoycHk7GCQ).

dtlawhon:
> 

No problem @MaxGhenis! Will get all of the feedback incorporated asap.

policyengine_us/parameters/gov/states/oh/tax/income/rates/agi.yaml
dtlawhon: Good question. I based the year on the tax form screenshot in the issue itself #1500 but I'm going to find an actual link to the form to double check, and then will include that as a second reference per your comment above.

dtlawhon: As best as I can find here, 2021 is the latest IT form posted - https://tax.ohio.gov/individual/get-a-form

policyengine_us/parameters/gov/states/oh/tax/income/rates/agi.yaml
dtlawhon: After double checking this, I believe the base amount applies independently of the rate structure.

If you go to page 32 of the tax form ([link to the specific page here](https://tax.ohio.gov/static/forms/ohio_individual/individual/2021/pit-it1040-booklet.pdf#page=32)) it gives a table, and you can see that at $25001 it starts with $347 as a base amount, and then increases in line with the rate on top of that. The table directly above that on page 31 seems clear that it's base amount + rate as well. At each new threshold, the stated base kicks in + the rate above as well. 

Let me know if I'm missing something or if there's a better way to represent this.

policyengine_us/parameters/gov/states/oh/tax/income/rates/agi.yaml
dtlawhon: Confirmed that the most recent published tax form is 2021, though it does note it was updated 1/22.

policyengine_us/parameters/gov/states/oh/tax/income/rates/agi.yaml
dtlawhon: Clarification: the brackets above the second one appear to be computable from the rate structure. But the initial $346.16 is not.

policyengine_us/parameters/gov/states/oh/tax/income/rates.yaml
```diff
@@ -0,0 +1,39 @@
+description: OH individual tax on nonbusiness income
+brackets:
+  - threshold:
+      0000-01-01: 0
+    base_amount:
```
MaxGhenis, line 5: remove this and another below

policyengine_us/parameters/gov/states/oh/tax/income/rates.yaml
```diff
@@ -0,0 +1,39 @@
+description: OH individual tax on nonbusiness income
+brackets:
+  - threshold:
+      0000-01-01: 0
```
MaxGhenis, line 4: let's set the thresholds to 2021 as well

setup.py
```diff
@@ -4,7 +4,7 @@
 
 setup(
     name="policyengine-us",
-    version="0.188.6",
+    version="0.191.3",
```
MaxGhenis, line 7: no need to change this - our bots will bump the version automatically

MaxGhenis (approved):
could you make these changes and run `make format`? good to merge from there!

dtlawhon:
Thanks @MaxGhenis for wrapping up those last items you mentioned above! Anything else required on this?

MaxGhenis:
Sure, the tests just aren't passing. Could you confirm those are the intended values?

dtlawhon:
Ah, I think the issue is in removing the initial base amount. All of the others are calculable from the rates applied to the previous bracket's income, but the first one for some bizarre reason is not. The test values are taken directly from the tax document example tables and double checked by hand calculation.

MaxGhenis:
Oh, Ohio has a cliff! Interesting, I think this is the first we've seen this among states.

In that case, one option could be adding a 1.385% rate for that first bracket ($346.16 / $25,000) and then adding an `income_tax_exempt` variable which is true for income below $25,000 (add as a separate parameter or grab from the parameter node - @nikhilwoodruff could you advise on the latter?) and then use that in determining the tax.

MaxGhenis:
I accidentally started a new PR from this branch, so went ahead and made the needed changes. Thanks again @dtlawhon, please let me know if I can help with other Ohio taxes or other programs you'd like to contribute. Merging.

PR #1600, opened 2022-12-27
Add normalised poverty variables

nikhilwoodruff:


MaxGhenis (approved):

MaxGhenis:
Optionally, tests would be nice and it may be cleaner to use aliases either on the app or country package side, but this is good to merge.

PR #1605, opened 2022-12-28
Remove duplicate formula components from `iitax` variable

### LINKED ISSUE #1597: `iitax` variable has both `adds`/`subtracts` and a `formula`
@nikhilwoodruff what takes precedence in these cases?

https://github.com/PolicyEngine/policyengine-us/blob/master/policyengine_us/variables/gov/irs/tax/federal_income/income_tax.py

MaxGhenis:
Fixes #1597

Translates the formula (which takes precedence) to `adds`/`subtracts` syntax


MaxGhenis:
This got handled in another PR

PR #1631, opened 2023-01-01
Pin and alphabetize dependencies, and pin Python 3.9.

### LINKED ISSUE #1628: Pin all package versions
For improved stability, following on doing this in https://github.com/PolicyEngine/policyengine-core/pull/74

MaxGhenis:
Fixes #1628

Also bumps policyengine-core


setup.py
```diff
@@ -27,41 +27,41 @@
...
-        "pytest",
-        "pytest-dependency",
-        "pyyaml",
-        "requests",
-        "synthimpute",
-        "tables",
-        "tqdm",
-        "click>=8.0.0",
-        "tabulate",
+        "click==8.1.3",
```
nikhilwoodruff, line 30: I'd favour pinning `<NEXT MAJOR` or `<NEXT MINOR`, given that `pip` will automatically find these versions?

MaxGhenis: What's the downside of pinning to patch versions? We had one issue from not pinning to the patch in policyengine-core and it seems like in general more specific environments will improve stability. @ibushong

ibushong: I'm on the road so it's tough for me to see the whole context here. But yeah generally you want to use inequality operators to specify and upper or lower limits of versions to prevent any incompatibilities. Note that this is for setup.py.

I may have miscommunicated my thoughts on this, since the use cases are different between setup.py and requirements.txt

MaxGhenis: Per our conversation, we'll pin everything and check in in a couple weeks to see how much we'd change

PR #1633, opened 2023-01-01
Permanent taxsim35-emulation branch (not for merging)

martinholmer:


MaxGhenis:
Thanks for kicking off this interesting approach, which could set the stage for modeling other reforms. How do you envision this facilitating ongoing integration testing against taxsim though?

martinholmer:
@MaxGhenis said in PR #1633:
> Thanks for kicking off this interesting approach, which could set the stage for modeling other reforms. How do you envision this facilitating ongoing integration testing against TAXSIM35 though?

The plan is to make a package from the code on the `taxsim35-emulation` branch and then use that package to conduct the validation testing against TAXSIM35 outside of the `policyengine-us` source code tree.   This is an experimental approach.  We'll see how it works.

Note that integration testing against TAXSIM35 results can still be conducted on the `master` branch if the test does not require the `taxsim35_emulation` parameter.

martinholmer:
@MaxGhenis, This branch was never to be merged into the master branch.

MaxGhenis:
I didn't merge this, I merged #1671, without realizing it was branched from this (it didn't say `(not for merging)`). @nikhilwoodruff is looking into reverting this, but to avoid this issue we're going to make a structural reform. There is only one outstanding taxsim deviation (PA income tax) so it's not worth trying to maintain separate branches and worrying about merging them.

policyengine_us/variables/gov/states/pa/tax/income/forgiveness/pa_tax_forgiveness_rate.py
```diff
@@ -33,9 +33,6 @@ def formula(tax_unit, period, parameters):
         )
         excess = eligibility_income - eligibility_income_increment
         forgiveness_increment = p.rate_increment
-        if parameters(period).gov.contrib.nber.taxsim35_emulation:
-            increments = excess / forgiveness_increment
-        else:
-            increments = np.ceil(excess / forgiveness_increment)
+        increments = np.ceil(excess / forgiveness_increment)
```
nikhilwoodruff, line 36: @martinholmer can I just check this is correct: if this branch was emulating taxsim, why did it change the code to act as in the non-emulation parameter scenario?

martinholmer:
@MaxGhenis said:
>  @nikhilwoodruff is looking into reverting this [i.e., PR #1633], but to avoid this issue we're going to make a structural reform.

Please do revert the merge of #1633.   It was marked as a DRAFT and github claims that means it will not be merged, so I don't understand why it did do the merge.

@nikhilwoodruff,  To find **all** the old `taxsim35_emulation` code, you'll have to go back to see what was eliminated on the master branch in PR #1632.

MaxGhenis:
> Please do revert the merge of #1633. It was marked as a DRAFT and github claims that means it will not be merged, so I don't understand why it did do the merge.

It's because I merged https://github.com/PolicyEngine/policyengine-us/pull/1671, which was not marked as draft.

No worries - we're in a simpler spot now with the structural taxsim35-emulation reform.

PR #1641, opened 2023-01-03
Widen Python to avoid breaking GCP instances

nikhilwoodruff:


nikhilwoodruff:
@MaxGhenis I'm going to merge this because right now the pros are:
* PolicyEngine Canada without needing to dig through GCP deployment configuration debugging
and the cons are:
* Nothing outright breaks

Issue #1571, opened 2022-12-16
Child/dependent care credit should be refundable in 2021


martinholmer:
According to the [2021 IRS Publication 503](https://www.irs.gov/pub/irs-pdf/p503.pdf), the CDCC was refundable in 2021.  It does not seem as if PolicyEngineUS incorporates this one-year feature.   But because TAXSIM35 does make the CDCC refundable in 2021, the following test case has a $5000 difference in the US income tax liability.

```
- name: Tax unit with taxsimid 80649 in r21.its.csv and r21.ots.csv
  absolute_error_margin: 0.01
  period: 2021
  input:
    people:
      person1:
        is_tax_unit_head: true
        age: 58
        employment_income: 21_010
        taxable_interest_income: 1_010
        social_security: 23_000
        real_estate_taxes: 14_000
        interest_expense: 22_000
        ssi: 0  # not in TAXSIM35
        state_supplement: 0  # not in TAXSIM35
        wic: 0  # not in TAXSIM35
      person2:
        age: 11
        ssi: 0  # not in TAXSIM35
        state_supplement: 0  # not in TAXSIM35
        wic: 0  # not in TAXSIM35
      person3:
        age: 11
        ssi: 0  # not in TAXSIM35
        state_supplement: 0  # not in TAXSIM35
        wic: 0  # not in TAXSIM35
    spm_units:
      spm_unit:
        members: [person1, person2, person3]
        snap: 0  # not in TAXSIM35
        tanf: 0  # not in TAXSIM35
    tax_units:
      tax_unit:
        members: [person1, person2, person3]
        tax_unit_childcare_expenses: 10_000
        premium_tax_credit: 0  # not in TAXSIM35
    households:
      household:
        members: [person1, person2, person3]
        state_code: TX  # no state income tax
  output:  # expected results from TAXSIM35 12/08/22 version
    taxsim_tfica: 1_607.26
    income_tax: -19_756.34
```

Here is the test result:
```
================================= FAILURES ===================================
_______________________________ test session _________________________________
/Users/mrh/work/Policy-Engine-US/US/r21-id80649.yaml:
  Test 'Tax unit with taxsimid 80649 in r21.its.csv and r21.ots.csv':
    income_tax@2021: [-14756.344] differs from -19756.33984375 
                                  with an absolute margin [4999.996] > 0.01
========================= short test summary info ============================
FAILED r21-id80649.yaml::
============================ 1 failed in 1.53s ===============================
```

[...]

PR #821, opened 2022-05-25
Re-implement CDCC from the law

nikhilwoodruff:
Fixes #820 (the test case now runs without error).

cc @martinholmer

PR #1644, opened 2023-01-04
Add test to prevent inadvertently making 2021 CDCC non-refundable

martinholmer:
The added test is from issue #1571.


MaxGhenis:
Thanks, could you just rename the file to integration.yaml? We reserve naming the tests with the name of the variable for unit tests.

martinholmer:
@MaxGhenis said:
> Thanks, could you just rename the file to integration.yaml? We reserve naming the tests with the name of the variable for unit tests.

There is already a unit test in the `cdcc.yaml` file, so I changed the new test to be a unit test (rather than rename the file).   Should be ready to merge.

MaxGhenis:
We'd still consider this an integration test. Since the [`cdcc` formula](https://github.com/PolicyEngine/policyengine-us/blob/master/policyengine_us/variables/gov/irs/credits/cdcc/cdcc.py) only has two inputs--`cdcc_relevant_expenses` and `cdcc_rate`--a unit test would only have those inputs as well.

martinholmer:
@MaxGhenis said:
> We'd still consider this an integration test. Since the [cdcc formula](https://github.com/PolicyEngine/policyengine-us/blob/master/policyengine_us/variables/gov/irs/credits/cdcc/cdcc.py) only has two inputs--cdcc_relevant_expenses and cdcc_rate -- a unit test would only have those inputs as well.

Max, it's good to have standards, but it seems to me you are setting a double standard in this case.   When I look at the old test in the `cdcc.yaml` file, I see that it is not a unit test by the above definition.  Yet this old test was developed by @nikhilwoodruff and reviewed and approved by you as part of PR #821.  So, why was that test in the `cdcc.yaml` file?

MaxGhenis:
I'd missed that the current test is an integration test, you're right. We might have refactored the CDCC code and not mirrored that to the test, or we may have been still developing our policy in May. Could you rename the full file to integration.yaml?

I also filed https://github.com/PolicyEngine/policyengine-us/issues/1646 to catch up our code to this standard.

MaxGhenis (approved):

PR #1648, opened 2023-01-04
Complete logic of MD care expense AGI subtraction

### LINKED ISSUE #1647: Incomplete MD AGI subtraction for child/dependent care expenses
The 2021 MD Form 502 instructions say this:

Screenshot 2023-01-04 at 3 58 18 PM

The code is missing part (b) of the Line 9 instructions.

martinholmer:
Fixes #1647.


MaxGhenis:
Thanks Martin. Rather than introduce new parameters that don't appear in the MD legal code, could you apply a branched simulation to get the parameters as of 2019? 2020 didn't affect this subtraction, but Maryland appears to have decoupled for both 2020 and 2021, so I think a parameter specifying the decoupling year, set to 2019 in 2020-21 and null or 0 otherwise, would probably keep us closest to the law. @nikhilwoodruff do you agree?

The [NY CTC code](https://github.com/PolicyEngine/policyengine-us/blob/master/policyengine_us/variables/gov/states/ny/tax/income/credits/ctc/ny_ctc.py) offers an example.

martinholmer:
@MaxGhenis said in PR #1648:
>  I think a parameter specifying the decoupling year, set to 2019 in 2020-21 and null or 0 otherwise, would probably keep us closest to the law.

I've implemented a parameter that is logically equivalent to your suggestion, but that is much easier to use in the variable formula.

MaxGhenis:
Fair enough. This approach won't capture the effect of federal CDCC reform on the state subtraction, but Maryland's automatic decoupling makes this ambiguous for any implementation. Merging.

PR #1672, opened 2023-01-14
Add taxsim-emulating structural reform

nikhilwoodruff:
@martinholmer I looked through your taxsim emulation branch but only found one substantive change, which I've implemented here (also left a comment on it in your PR).

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    added:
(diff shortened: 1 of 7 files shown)


MaxGhenis:
Will we specify the taxsim reform in taxsim tests in a separate PR?

martinholmer:
@nikhilwoodruff, it seems as if PR #1672 is correct.

PR #1742, opened 2023-02-02
NYC taxable income

### LINKED ISSUE #1728: NYC taxable income
@leogoldman is adding this as an empty variable in #1727, now needs a formula.

https://www.tax.ny.gov/pdf/2022/printable-pdfs/inc/it201i-2022.pdf#page=16

leogoldman:
Fixes #1728

Without contributions to the NY Charitable Gifts Trust Fund accounts and itemized deductions for those contributions, NYC taxable income is equal to NY AGI - exemptions for dependents.


MaxGhenis:
>NYC taxable income is equal to NY AGI - exemptions for dependents.

Might this come from NYS law? From the instruction booklet:

>The value of each New York State dependent exemption is $1,000.

Could you check how the NYC legal code defines taxable income?

leogoldman:
https://codelibrary.amlegal.com/codes/newyorkcity/latest/NYCadmin/0-0-0-13686
From § 11-1711, city taxable income = city AGI - ( city deductions + city exemptions )
From § 11-1712, city AGI = federal AGI with modifications under this section

None of the modifications in § 11-1712 stand out to me but the deduction amounts in § 11-1714 are not the same as the federal deduction amounts, suggesting that there must be some difference?

Form I-201 seems to give the impression that NYC AGI = NY AGI

MaxGhenis:
Thanks, you're right, we should define the NYC dependent exemption as a separate parameter from the NYS one, since the NYC law defines it.

This section will require some more advanced coding, since IIUC one can itemize in NYC independently of their federal itemization. So we may need to branch the logic and compute taxes twice (at least federally, it's not as simple as whether itemized deductions exceed the standard deduction).
>If federal taxable income of a city resident individual is determined by itemizing deductions or claiming the federal standard deduction from his or her federal adjusted gross income, such resident individual may elect to deduct his or her city itemized deduction or claim his or her city standard deduction. 

For now, I'd suggest referencing an empty `nyc_taxable_income_deductions` variable and we'll save the standard/itemized logic for another PR.

Other than that, this would just need a unit test.

PR #1799, opened 2023-02-16
Add DC EITC

### LINKED ISSUE #1714: DC EITC

cmfung:
Fixes #1714


MaxGhenis:
@nikhilwoodruff could you let us know how we can make `defined_for = StateCode.DC` work? This is an identical test and variable to the MA EITC so I think that's the only reason the [`Variable: DC does not exist` error is arising](https://github.com/PolicyEngine/policyengine-us/actions/runs/4197309937/jobs/7279460081#step:5:543).

policyengine_us/parameters/gov/states/dc/tax/income/credits/etic/match.yaml
```diff
@@ -0,0 +1,12 @@
+description: DC matches this percentage of the federal earned income tax credit.
```
MaxGhenis, line 1: @cmfung if you can fix the typo in the folder `etic` this should work

PR #1810, opened 2023-02-20
Create CARE Discount Variable, FERA Program

KathrynPanger:
fixes #1798

PR #1813, opened 2023-02-21
KathrynPanger/issue1812

KathrynPanger:
Updates CARE and FERA references. Also adds FERA parameter which was previously missing, percent of poverty line for eligibility.


MaxGhenis:
Since these are relatively minor changes and this branch has parts of the #1810 branch, let's put the changes in there. That PR can close both issues.

PR #1817, opened 2023-02-22
Add IL to the state income tax list

martinholmer:
Thanks to the development effort of @yvvijay121, the Illinois income tax has been working for some time.  But the `il_income_tax` variable has never been added to the list of working state income taxes in the `state_income_tax` variable formula.  This pull request adds the `il_income_tax` variable to that list and alphabetizes the list.  This addition causes no test changes.

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,5 @@
+- bump: patch
+  changes:
+    fixed:
+      - Incomplete list of state income tax variables by adding IL.
+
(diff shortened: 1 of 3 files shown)


MaxGhenis:
Thanks @martinholmer - while you're in here could you add it to https://github.com/PolicyEngine/policyengine-us/blob/master/policyengine_us/modelled_policies.yaml also? We model IL TANF as well.

martinholmer:
@MaxGhenis,  Why isn't MO in the `modelled_policies.yaml` file?

MaxGhenis:
Please add MO as well. MO state income tax doesn't count toward the SALT deduction due to the circular reference, but we're displaying its values so we should include it in the dialog. Thanks.

Issue #1857, opened 2023-03-04
People who are too high-income are eligible for California CARE via categorical eligibility


MaxGhenis:
I recorded this video to show the issue. I don't know why it's happening, but an integration test could help, e.g. checking that this family of four with $79k income shouldn't be eligible.

https://www.loom.com/share/75dd7276d16a4353ab77b90581af67cd

MaxGhenis:
For posterity: #3431 didn't actually fix this. We discovered that it had to do with Medicaid, which covers children up to 266% of FPL in California; the code has been working as intended.

PR #1865, opened 2023-03-06
Integration Test of CARE/FERA Shows SNAP Issue

KathrynPanger:
This PR contains a failed integration test which demonstrates that the reason CARE and FERA variables aren't working (issue #1857) is because the SNAP variable is assigning money to people who do not qualify for SNAP.


PavelMakarchuk:
Closing since there are many changes which did not seem to be intended

PR #2128, opened 2023-04-13
Fix bug in recent revision of CalEITC logic

martinholmer:
Pull request #2124 had the laudable goal of implementing CalEITC without cumbersome branching logic.  But in the course of implementing the new `ca_eitc` variable formula, a mistake was made.  Fortunately, the mistake was identified by the failure of two CalEITC integration tests.  Unfortunately, the warning provided by the failure of existing tests was ignored.  Instead of looking for the problem, the error tolerances of the two failing tests were increased enough to make the tests pass, and then the pull request was merged without any review.

This kind of development behavior is akin to firefighters at a fire station hearing the fire alarm, and then responding by turning off the alarm rather than jumping into the fire truck to go put out the fire.

The changes in this pull request restore the error tolerances to what they were before #2124, and then fixes the bug in the `ca_eitc` formula and removes an unused CalEITC parameter.


nikhilwoodruff:
Ah sorry Martin, my bad developer habit to blame here! Will avoid doing that in future.

policyengine_us/variables/gov/states/ca/tax/income/credits/earned_income/ca_eitc.py
```diff
@@ -24,13 +24,16 @@ def formula(tax_unit, period, parameters):
         phased_in_amount = phase_in_income * phase_in_rate
 
         phase_out_min_income = p.phase_out.start.calc(child_count)
-        phase_out_rate = p.phase_out.rate.calc(child_count)
+        phase_out_rate = (
```
nikhilwoodruff, line 27: Ah, makes sense.

nikhilwoodruff (approved):

PR #2131, opened 2023-04-14
Fix README.md display on PyPI

westurner:

- The README.md render on PyPI: https://pypi.org/project/policyengine-us/
- https://packaging.python.org/en/latest/guides/making-a-pypi-friendly-readme/#including-your-readme-in-your-package-s-metadata
- https://github.com/pypa/readme_renderer/blob/main/readme_renderer/__main__.py

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,5 +1,5 @@
 - bump: patch
   changes:
     changed:
-    - DOC,BLD: setup.py: set `long_description_content_type='text/markdown'` for pypi
+    - Fixed README.md display on PyPI.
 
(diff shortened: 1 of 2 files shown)


MaxGhenis:
Thank you @westurner! Could you please also
* Run `make format` (we use `black` which requires double quotes)
* Populate `changelog_entry.yaml` as per other recently merged PRs (this can be a `patch`)

westurner:
Should be ready to merge

changelog_entry.yaml
```diff
@@ -0,0 +1,5 @@
+- bump: patch
+  changes:
+    changed:
+    - DOC,BLD: setup.py: set `long_description_content_type='text/markdown'` for pypi
```
nikhilwoodruff, line 4:
```suggestion
    - Fixed README.md display on PyPI.
```

westurner:
Thanks!

PR #2222, opened 2023-05-01
New Jersey TANF earned income disregard

### LINKED ISSUE #2217: New Jersey TANF earned income disregard
Thanks for filing an issue! Please remove any top-level headers that do not apply.

# Variable to add: name

# Variable to amend: name

# Parameter to add: name

# Parameter to amend: name

xudonglai0426:
Fixes #2217

Co-authored-by: VioletChen0916 <VioletChen0916@users.noreply.github.com>

copilot:all

### DIFF
--- a/policyengine_us/variables/gov/states/nj/njdhs/tanf/income/nj_tanf_countable_earned_income.py
+++ b/policyengine_us/variables/gov/states/nj/njdhs/tanf/income/nj_tanf_countable_earned_income.py
@@ -1,0 +1,35 @@
+from policyengine_us.model_api import *
+
+
+class nj_tanf_countable_earned_income(Variable):
+    value_type = float
+    entity = SPMUnit
+    label = "New Jersey TANF countable earned income"
+    unit = USD
+    definition_period = YEAR
+    defined_for = StateCode.NJ
+
+    def formula(spm_unit, period, parameters):
+        # Get gross earned income.
+        person = spm_unit.members
+        gross_earned_income = spm_unit("nj_tanf_gross_earned_income", period)
+        # Multiply by 100% minus the EID.
+        p = parameters(
+            period
+        ).gov.states.nj.njdhs.tanf.income.earned_income_deduction
+        months_enrolled_in_tanf = person("months_enrolled_in_tanf", period)
+        weekly_hours_worked = person("weekly_hours_worked", period)
+        # New Jerset Admin Code 10:90-3.8(b)
+        return where(
+            months_enrolled_in_tanf > 1,
+            where(
+                months_enrolled_in_tanf > 7,
+                gross_earned_income * (1 - p.additional_percent),
+                where(
+                    weekly_hours_worked >= p.work_hours_threshold,
+                    gross_earned_income * (1 - p.consecutive_month_percent),
+                    gross_earned_income * (1 - p.additional_percent),
+                ),
+            ),
+            gross_earned_income * (1 - p.first_month_percent),
+        )
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: minor
+  changes:
+    added:
+    - New Jeresy TANF earned income disregards.
--- a/policyengine_us/parameters/gov/states/nj/njdhs/tanf/income/earned_income_deduction/additional_percent.yaml
+++ b/policyengine_us/parameters/gov/states/nj/njdhs/tanf/income/earned_income_deduction/additional_percent.yaml
@@ -1,0 +1,11 @@
+description: New Jersey excludes this percentage of earned income for household enrolled in TANF for each additional month.
+
+metadata:
+  unit: /1
+  label: New Jersey TANF earned income deduction
+  reference:
+    - title: New Jersey Administrative Code(N.J.A.C 10:90-3.8) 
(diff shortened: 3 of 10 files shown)


policyengine_us/variables/gov/states/nj/njdhs/tanf/income/nj_tanf_countable_earned_income.py
```diff
@@ -19,16 +19,41 @@ def formula(spm_unit, period, parameters):
...
+            months_enrolled_in_tanf <= p.consecutive_month_threshold
+        ) and (months_enrolled_in_tanf > p.first_month_threshold)
+        person_enrolled_in_tanf_for_additional_months_with_work_hours_over_20 = (
+            months_enrolled_in_tanf > p.consecutive_month_threshold
+            and person_meet_higher_work_hours_threshold
+        )
+        person_enrolled_in_tanf_for_additional_months_with_Work_hours_below_20 = (
+            months_enrolled_in_tanf > p.first_month_threshold
+        )
+        if person_meet_higher_work_hours_threshold:
```
MaxGhenis, line 38: Since `if` statements aren't vectorized, we have to use `where` and `select` statements instead

xudonglai0426: OK. Is there an example shows how to use "select" statement or how to use "where" and "select" to achieve the function of "if" there? Originally, we are using a nested where function, it is not so readable although it passes all the test. That's why we are trying to use "if" function.

policyengine_us/variables/gov/states/nj/njdhs/tanf/income/nj_tanf_countable_earned_income.py
```diff
@@ -0,0 +1,75 @@
...
+            gross_earned_income * (1 - p.higher_work_hours.additional_percent),
+            result,
+        )
+        # If applicant is enrolled in TANF for additional months with work hours below 20, the earned income deduction is 50%.
+        result = where(
+            person_enrolled_in_tanf_for_additional_months_with_work_hours_below_20,
+            gross_earned_income * (1 - p.lower_work_hours.additional_percent),
+            result,
+        )
+        return result
```
leogoldman, line 75: Here's an example of how to use `select()`:
https://github.com/PolicyEngine/policyengine-us/blob/83b0d38b209f07dbe2b7d593c3e7dce7e26778b6/policyengine_us/variables/gov/states/nj/tax/income/nj_main_income_tax.py#L4

The function's first argument is a condition list, the second argument is the list of values to return if that condition is true, and the (optional) third argument is the value to return if none of them are true (by default this is zero). The function goes through the condition list and returns the corresponding value. For example, if the third condition is true, the third value in the return list would be returned.
Here, you would want something like
`select( [ enrolled_in_tanf_first_month, meets_hours_threshold, ... ],
               [ gross_earned_income * ( 1-p.higher_work_hours.first_month_percent),
                 gross_earned_income
            * (1 - p.higher_work_hours.consecutive_month_percent),
            ... ]
)
`
You don't need to specify a default return value since you want zero anyway.

policyengine_us/variables/gov/states/nj/njdhs/tanf/income/nj_tanf_countable_earned_income.py
```diff
@@ -0,0 +1,75 @@
...
+            and person_meet_higher_work_hours_threshold
+        )
+        person_enrolled_in_tanf_for_additional_months_with_work_hours_over_20 = (
+            months_enrolled_in_tanf > p.consecutive_month_threshold
+            and person_meet_higher_work_hours_threshold
+        )
+        person_enrolled_in_tanf_for_additional_months_with_work_hours_below_20 = (
+            months_enrolled_in_tanf > p.first_month_threshold
+            and person_meet_lower_work_hours_threshold
+        )
```
leogoldman, line 43: Since you want element-wise boolean operations, make sure you use `&` instead of `and`.

policyengine_us/variables/gov/states/nj/njdhs/tanf/income/nj_tanf_countable_earned_income.py
```diff
@@ -0,0 +1,75 @@
...
+        ).gov.states.nj.njdhs.tanf.income.earned_income_deduction
+        months_enrolled_in_tanf = person("months_enrolled_in_tanf", period)
+        weekly_hours_worked = person("weekly_hours_worked", period)
+        # New Jerset Admin Code 10:90-3.8(b)
+        person_meet_higher_work_hours_threshold = (
+            weekly_hours_worked >= p.work_hours_threshold
+        )
+        person_meet_lower_work_hours_threshold = (
+            weekly_hours_worked < p.work_hours_threshold
+        )
```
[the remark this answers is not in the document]
leogoldman: No need for a variable that is the opposite of the first variable. If you need to check if the number of hours is less than the threshold, you can use `~meets_hours_threshold`, flipping all the booleans

leogoldman (commented):
Try to use variable names that aren't quite as long (but still descriptive) to make the code a bit easier to read. Also, avoid creating variables that are the opposite of existing variables (use the boolean inverse instead)

policyengine_us/variables/gov/states/nj/njdhs/tanf/income/nj_tanf_countable_earned_income.py
leogoldman: @xudonglai0426 see my comment in the review below

PR #2450, opened 2023-06-12
[DRAFT] initial changes for HUD User data.

shahzorkhan123:
copilot:all


policyengine_us/data/datasets/huduser/huduser.py
```diff
@@ -0,0 +1,124 @@
...
+from policyengine_core.data import PublicDataset
+import h5py
+from policyengine_us.data.datasets.huduser.raw_acs import RawHUDUSER
+from policyengine_us.data.storage import policyengine_us_MICRODATA_FOLDER
+from pandas import DataFrame
+
+
+class HUDUSER(PublicDataset):
+    name = "huduser"
+    is_openfisca_compatible = True
```
shahzorkhan123, line 11: @nikhilwoodruff  does this mean all the columns need to match variable names?

policyengine_us/variables/gov/hud/hud_free_market_rent.py
nikhilwoodruff, line 1: We could combine these into one variable `hud_free_market_rent`

nikhilwoodruff (commented):
So, the full process:

* Take a household
* Look up min(num bedrooms, num people) in the FMR table for the local area FIPS
...

policyengine_us/data/datasets/huduser/huduser.py
shahzorkhan123: discussed, will change to false

policyengine_us/data/datasets/huduser/huduser.py
```diff
@@ -0,0 +1,124 @@
...
+
+        # Prepare raw HUDUSER tables
+        year = int(year)
+        if year in self.years:
+            self.remove(year)
+        if year not in RawHUDUSER.years:
+            RawHUDUSER.generate(year)
+
+        raw_data = RawHUDUSER.load(year)
+        huduser_data = h5py.File(HUDUSER.file(year), mode="w")
```
shahzorkhan123, line 30: Discussed, no need to convert to H5, use csv/xls instead.

PR #2563, opened 2023-07-03
Eliminate divide-by-zero warnings in Income decile calculations

### LINKED ISSUE #2562: Divide-by-zero warnings when testing income_decile variables
When performing the variable tests for the `household_income_decile` and the `spm_unit_income_decile` variables, divide-by-zero warning arise in the `microdf` package code because all the weights are zero in the PolicyEngineUS `test_variables` DEFAULT_SITUATION.

There is nothing wrong with the `microdf` code, it is just inappropriate to call the `household_income_decile` and `spm_unit_income_decile` formulas when the Microsimulation object uses a Dataset that contains no weights.

So, the solution is to add the `household_income_decile` and `spm_unit_income_decile` variables to `test_variables` `EXEMPTIONS` so that these two variables are not tested as part of the `test_variables` logic, and adding them to `test_microsim_runs_cps` where the Microsimulation Dataset (the CPS) does include weights.

martinholmer:
Fixes #2562.

copilot:all

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,4 +1,4 @@
 - bump: patch
   changes:
     fixed:
-    - Array divide-by-zero warnings in household/spmunit income decile variable tests by skipping the tests.
+    - Array divide-by-zero warnings in household/spmunit income decile variable tests by moving the tests.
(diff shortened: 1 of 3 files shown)


nikhilwoodruff (approved):
Thanks @martinholmer - apologies for my delay.

PR #2755, opened 2023-08-02
Equivalised household net income is SPM-unit based, not household-based

### LINKED ISSUE #2754: Equivalised household net income is SPM-unit based, not household-based
We should use the household net income income definition in equivalised household income, rather than summing equivalised SPM unit incomes.

nikhilwoodruff:
Fixes #2754

copilot:all


MaxGhenis (commented):
could you add a test

PR #3046, opened 2023-09-19
DC single-joint threshold ratio switch

### LINKED ISSUE #3045: Add a reform parameter to change the ratio between single and joint tax brackets in DC

nikhilwoodruff:
Fixes #3045

copilot:all

### DIFF
--- a/policyengine_us/reforms/dc_tax_threshold_joint_ratio.py
+++ b/policyengine_us/reforms/dc_tax_threshold_joint_ratio.py
@@ -1,0 +1,34 @@
+from policyengine_us.model_api import *
+
+
+def create_dc_tax_threshold_joint_ratio_reform(parameters, period, bypass=False):
+    joint_ratio = parameters(period).gov.contrib.dc_tax_threshold_joint_ratio
+
+    class dc_income_tax_before_credits_joint(Variable):
+        value_type = float
+        entity = TaxUnit
+        label = "DC income tax before credits when married couples file jointly"
+        unit = USD
+        definition_period = YEAR
+        reference = (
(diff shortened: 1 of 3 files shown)


nikhilwoodruff:
Needs tests

policyengine_us/reforms/dc_tax_threshold_joint_ratio.py
```diff
@@ -0,0 +1,52 @@
...
+        unit = USD
+        definition_period = YEAR
+        reference = (
+            "https://otr.cfo.dc.gov/sites/default/files/dc/sites/otr/publication/attachments/52926_D-40_12.21.21_Final_Rev011122.pdf#page=36"
+            "https://otr.cfo.dc.gov/sites/default/files/dc/sites/otr/publication/attachments/2022_D-40_Booklet_Final_blk_01_23_23_Ordc.pdf#page=34"
+        )
+        defined_for = StateCode.DC
+
+        def formula(tax_unit, period, parameters):
+            taxinc = max_(0, tax_unit("dc_taxable_income_joint", period))
```
MaxGhenis, line 29: Will this work in conjunction with repealing married filing separately on same return? Or do we need to base it on filing status?

MaxGhenis (approved):

policyengine_us/reforms/dc_tax_threshold_joint_ratio.py
MaxGhenis: I forgot that married filing separately on same return is a normal parameter now, not a structural reform. So I think it should work but it'll be easier to test in the app once merged.

PR #3319, opened 2023-11-28
Fix definition of income used to calculate Maine fairness credits

### LINKED ISSUE #3248: Fix definition of income used to calculate Maine fairness credits
The 0.520.0 version, includes the following specification of income used to calculate the Maine fairness credits:

Screenshot 2023-11-08 at 11 31 57 AM

But the first two items in the above list are incorrect.   The [2021 form](https://www.maine.gov/revenue/sites/maine.gov.revenue/files/inline-files/21_1040me_sched_pstfc_dwnldff.pdf) makes it clear that the first list item is not federal AGI, but rather federal total income (before above-the-line adjustments), and that the second item is not total social security benefits, but rather only social security benefits not included in federal total income.   Including all social security benefits overstates fairness income in cases where some social security was taxed at the federal level.  This overstatement of fairness income causes an underestimate of the size of the fairness credits in many cases.  And this leads to many differences with TAXSIM35.

XZ-mabel:
Fixes #3248

copilot:all

### DIFF
--- a/policyengine_us/parameters/gov/states/me/tax/income/credits/fairness/income_sources.yaml
+++ b/policyengine_us/parameters/gov/states/me/tax/income/credits/fairness/income_sources.yaml
@@ -1,7 +1,7 @@
 description: Maine accounts for the following income sources for the sales and property tax credits.
 values:
   2021-01-01:
-    - adjusted_gross_income
+    - irs_gross_income
     - social_security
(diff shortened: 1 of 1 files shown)


MaxGhenis (approved):
Thanks - confirmed that our irs_gross_income concept matches the total income concept from the 1040, which excludes tax-exempt ss

PR #3483, opened 2023-12-17
Fix Alabama federal tax deduction calculations

### LINKED ISSUE #3478: Fix Alabama federal tax deduction calculations
There are at least two problems with the AL federal tax deduction calculations in the 0.582.2 code.  Here is the 2022 worksheet:

Screenshot 2023-12-16 at 1 16 41 PM

The first problem is that on line 1 the phrase "line 22 on [IRS] 2022 Form 1040" seems like it should be interpreted as `income_tax_before_refundable_credits` (rather than `income_tax_before_credits`, which ignores non-refundable federal credits).

The second problem is that on line 6 the phase "if amount is negative enter zero" is ignored in the code, which causes negative deductions in many cases.

PavelMakarchuk:
Fixes #3478

copilot:all

### DIFF
--- a/policyengine_us/variables/gov/states/al/tax/income/deductions/federal_income_tax/al_federal_income_tax_deduction.py
+++ b/policyengine_us/variables/gov/states/al/tax/income/deductions/federal_income_tax/al_federal_income_tax_deduction.py
@@ -10,6 +10,9 @@ class al_federal_income_tax_deduction(Variable):
     reference = "https://www.revenue.alabama.gov/ultraviewer/viewer/basic_viewer/index.html?form=2023/01/22f40abk.pdf#page=20"
     defined_for = StateCode.AL
 
-    adds = "gov.states.al.tax.income.deductions.federal_tax.countable_sources"
-
-    subtracts = "gov.states.al.tax.income.deductions.federal_tax.credits"
+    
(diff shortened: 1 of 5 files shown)


policyengine_us/variables/gov/states/al/tax/income/deductions/federal_income_tax/al_federal_income_tax_deduction.py
```diff
@@ -10,6 +10,9 @@ class al_federal_income_tax_deduction(Variable):
     reference = "https://www.revenue.alabama.gov/ultraviewer/viewer/basic_viewer/index.html?form=2023/01/22f40abk.pdf#page=20"
     defined_for = StateCode.AL
 
-    adds = "gov.states.al.tax.income.deductions.federal_tax.countable_sources"
-
-    subtracts = "gov.states.al.tax.income.deductions.federal_tax.credits"
+    
+    def formula(tax_unit, period, parameters):
+        p = parameters(period).gov.states.al.tax.income.deductions.federal_tax
+        income_sources = p.countable_sources
```
MaxGhenis, line 16: and below
```suggestion
        income_sources = add(tax_unit, period, p.countable_sources)
```

MaxGhenis (approved):

PR #3502, opened 2023-12-20
Update South Carolina net_capital gain deduction formula

### LINKED ISSUE #3452: Question about South Carolina net capital gains deduction
In the 0.578.0 model version, the formula for the `sc_net_capital_gain_deduction` calculates the deduction as 44% of the `net_capital_gain` variable (which **includes** qualified dividends) rather than 44% of the `net_capital_gains` variable (which **excludes** qualified dividends).  The cited reference provides a detailed example, but the example does not include any qualified dividends.   Also, for whatever it is worth, TAXSIM35 interpretes this SC deduction as having nothing to do with qualified dividends.

So, the question is whether there is some specific tax-legislation or form-instruction language that can be pointed to in order to rationalize the use of the `net_capital_gain` variable rather the `net_capital_gains` variable.

hua7450:
Fixes #3452

copilot:all


PavelMakarchuk (approved):

policyengine_us/variables/gov/states/sc/tax/income/deductions/net_capital_gain/sc_net_capital_gain_deduction.py
```diff
@@ -18,5 +18,11 @@ def formula(tax_unit, period, parameters):
...
             period
         ).gov.states.sc.tax.income.deductions.net_capital_gain
-        capital_gains = tax_unit("net_capital_gain", period)
-        return capital_gains * p.rate
+        qualified_dividends = add(
+            tax_unit, period, ["qualified_dividend_income"]
+        )
+        ltcg = add(tax_unit, period, ["long_term_capital_gains"])
+        stcg = add(tax_unit, period, ["short_term_capital_gains"])
+        net_capital_gain = max_(0, ltcg + stcg + qualified_dividends)
```
MaxGhenis, line 26: replacing the above, though more importantly this doesn't seem to align with https://github.com/PolicyEngine/policyengine-us/issues/3452#issuecomment-1857948935 am i wrong?
```suggestion
        net_capital_gain = max_(0, add(tax_unit, period, ["qualified_dividend_income", "long_term_capital_gains", "short_term_capital_gains"]))
```

MaxGhenis (changes requested):

policyengine_us/variables/gov/states/sc/tax/income/deductions/net_capital_gain/sc_net_capital_gain_deduction.py
hua7450: I think we should include the `qualified_dividend_income` as part of the base. For the `short_term_capital_gains`, the example in issue 3452 shows that `short_term_capital_gains` can be negative, so I simply add them up.

PavelMakarchuk: removed the dividend income and added the integration test from the form

hua7450: I have a quick question, why do we set the short term capital gains higher than 0, isn't the example shows that it can be negative?

MaxGhenis (approved):

PR #3547, opened 2023-12-29
Removed the redundant reassignments of adds and subtracts

### LINKED ISSUE #3533: Remove extraneous logic in income_tax.py
We have two adds/subtracts pairs, plus a formula. Seems like we should remove adds/subtracts:

https://github.com/PolicyEngine/policyengine-us/blob/1aa383dc8d0c62904fc83b4b8c8b909e0c3911a4/policyengine_us/variables/gov/irs/tax/federal_income/income_tax.py#L11-L30

See also https://github.com/PolicyEngine/policyengine-core/issues/139

shashank524:
Reduced the redundancy in the `income_tax.py` file by removing duplicate `adds` and `subtracts` assignments. (Fixes #3533 )

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,4 +1,4 @@
 - bump: patch
   changes:
     fixed:
-    - Removed extraneous add/subtract pairs in income_tax calculation logic
+    - Removed extraneous add/subtract pairs in income_tax calculation logic.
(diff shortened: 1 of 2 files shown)


MaxGhenis (changes requested):
Let's remove both adds/subtracts. Variables should only have either a formula or a single adds/subtracts pair.

MaxGhenis:
`git pull upstream master` will fix the version check

PR #3560, opened 2024-01-03
Fix South Carolina taxable income calculation

### LINKED ISSUE #3552: Fix South Carolina taxable income calculation
In the 0.601.0 version, the calculation of the `sc_taxable_income` variable does not correspond to the SC income tax form or instructions.   The metadata in the additions and subtractions parameter files are correct.  They say:
```
description: South Carolina adds these sources from federal taxable income 
to determine state taxable income. 
```
and
```
description: South Carolina subtracts these sources from federal taxable income
to determine state taxable income.
```
But the formula for the `sc_taxable_income` does not do that calculation: it adds and subtracts these amounts to the federal AGI not to the federal taxable income.   As as result, there are over 94,000 tax differences in the `j21` sample (out of 100,000 test cases) as described in the integration test table in issue 993.

Eliminate this bug by making the variable formulas correspond to how the instructions say to fill out the SC income tax form.

PavelMakarchuk:
Fixes #3552


policyengine_us/variables/gov/states/sc/tax/income/sc_taxable_income.py
```diff
@@ -9,5 +9,5 @@ class sc_taxable_income(Variable):
     unit = USD
     definition_period = YEAR
```
MaxGhenis, line 10: could you add a reference? eg to indicate that values can be negative

MaxGhenis (changes requested):

policyengine_us/variables/gov/states/sc/tax/income/sc_taxable_income.py
PavelMakarchuk: add the reference and changed to a max_ function

MaxGhenis (approved):

PR #3596, opened 2024-01-12
Add Louisiana federal tax deduction logic for 2021

### LINKED ISSUE #3589: Add Louisiana deduction logic for 2021
The 0.615.0 version implements 2021 LA deductions incorrectly.  There is at least one problem.   The 2021 deduction of federal income taxes, for which every taxpayer was eligible, is not implemented.  The 2021 tax documents are referenced, but the Python code does not implement this feature of the LA income tax.

Also, note that the handling of LA itemized deductions changed significantly beginning in 2022.

PavelMakarchuk:
Fixes #3589


MaxGhenis (approved):
could you replace the current separate `+ `calls with `add()`? otherwise good to merge

PR #3622, opened 2024-01-14
Fix West Virginia family tax credit

### LINKED ISSUE #3603: Fix West Virginia family tax credit
In the 0.615.0 version, there are several problems with the representation of the WV family tax credit.

First, the name `wv_low_income_family_tax_credit` is confusing because there is no credit with that name on the WV income tax form or in its instructions.   Plus, the current variable name is verbose: what does "low_income" add?   Most credits are restricted to taxpayers with lower income.  For clarification, I would suggest naming this variable `wv_family_tax_credit`.

Second, the credit formula uses state AGI (rather than federal AGI plus WV additions and tax-exempt interest), which leads to differences between results produced by Policyengine-US and TAXSIM35.

Third, the credit formula ignores the fact that paying the federal AMT makes you ineligible for the WV family tax credit.

PavelMakarchuk:
Fixes #3603

### DIFF
--- a/policyengine_us/variables/gov/states/wv/tax/income/credits/liftc/wv_low_income_family_tax_credit_eligible.py
+++ b/policyengine_us/variables/gov/states/wv/tax/income/credits/liftc/wv_low_income_family_tax_credit_eligible.py
@@ -1,0 +1,14 @@
+from policyengine_us.model_api import *
+
+
+class wv_low_income_family_tax_credit_eligible(Variable):
+    value_type = float
+    entity = TaxUnit
+    label = "West Virginia low-income family tax credit"
+    unit = USD
+    definition_period = YEAR
+    defined_for = StateCode.WV
+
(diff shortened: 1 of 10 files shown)


MaxGhenis (approved):

policyengine_us/variables/gov/states/wv/tax/income/credits/liftc/wv_low_income_family_tax_credit_eligible.py
```diff
@@ -0,0 +1,15 @@
+from policyengine_us.model_api import *
+
+
+class wv_low_income_family_tax_credit_eligible(Variable):
+    value_type = float
```
MaxGhenis, line 5: also change unit
```suggestion
    value_type = bool
```

MaxGhenis (approved):
can merge after this

PR #3636, opened 2024-01-17
Remove CPS person weights in favour of households

nikhilwoodruff:


policyengine_us/data/datasets/cps/cps.py
```diff
@@ -100,7 +100,6 @@ def add_id_variables(
 
     # Add weights
     # Weights are multiplied by 100 to avoid decimals
-    cps["person_weight"] = person.A_FNLWGT / 1e2
     cps["family_weight"] = family.FSUP_WGT / 1e2
```
MaxGhenis, line 103: should we also remove these while we're here, and use only household weights?

MaxGhenis:
should we still do this @nikhilwoodruff?

PavelMakarchuk:
I believe this was superseded by changes in `us-data` @nikhilwoodruff feel free to reopen if will relevant

PR #3660, opened 2024-01-21
Cap SNAP excess shelter deduction after applying utility allowance

### LINKED ISSUE #3652: Cap SNAP excess shelter deduction after applying utility allowance
From @CalebPena:

For SNAP, we think that there is a problem with [snap_excess_shelter_expense_deduction](https://github.com/PolicyEngine/policyengine-us/blob/52fcfe5be6e3fe4aad84bd069883a057d25629e7/policyengine_us/variables/gov/usda/snap/income/deductions/shelter/snap_excess_shelter_expense_deduction.py#L33-L35) where the cap is being applied before the utility allowance is applied. Kathleen's thoughts:

> I think that the utility allowance is part of housing expenses and therefore subject to the excess shelter cap. When I run the profile in row 2 of the [sheet](https://docs.google.com/spreadsheets/d/1UEER3_aEvEfFK8anTS1C7iR0c6bq_AFER3MvW_Zq2ik/edit#gid=0) through SNAP Screener, I get results that are consistent with my math (see attached). My math uses the $672 cap for both rent and utilities.
> As per Section 4.407.3 of the [regs](https://www.sos.state.co.us/CCR/GenerateRulePdf.do?ruleVersionId=11117&fileName=10%20CCR%202506-1) (p. 107), HHs get a deduction for allowable monthly shelter costs in excess of 50% of the HH's income after all other deductions. A shelter deduction cap - currently $672 - applies. As per Section 4.407.31 (p. 109), a mandatory standard utility allowance deduction is used "in determining a HH's excess shelter deduction." Practically, it makes sense that utilities would be part of shelter expenses. Also note that Section 4.407.E. (p. 106) lists the applicable deductions and doesn't list utilities as a separate deduction; utilities are included under 4.407.3 (excess shelter deduction) as a subpart - 4.407.31; and 4.407.3E.4 and 5. list heating, phone, etc. as part of shelter costs.
> So I think the gist is that rent/mortgage + the applicable utility allowance = shelter costs. And note that the current utility allowances are on pp. 109 and 110 of the regs - looks like PE needs to make updates.

PavelMakarchuk:
Fixes #3652

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    fixed:
+    - Cap SNAP excess shelter deduction after applying utility allowance.
--- a/policyengine_us/tests/policy/baseline/gov/usda/snap/income/deductions/snap_excess_shelter_expense_deduction.yaml
+++ b/policyengine_us/tests/policy/baseline/gov/usda/snap/income/deductions/snap_excess_shelter_expense_deduction.yaml
@@ -44,3 +44,16 @@
     housing_cost: 1
   output:
     snap_excess_shelter_expense_deduction: 159.73
+
+- name: Caps at $701 per month in GU with housing costs and utility allowance added.
+  period: 2022-01
+  input:
+    snap_net_income_pre_shelter: 0
+    snap_region_str: GU
+    housing_cost: 4_800 # 400 monthly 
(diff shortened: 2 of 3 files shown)


MaxGhenis (approved):

CalebPena (approved):
Looks good, thanks for making this a priority.

PR #3695, opened 2024-01-26
Adjust the LA General relief computation to include withholding rules

### LINKED ISSUE #3694: Adjust the LA General relief computation to include withholding rules
Currently the `la_general_relief_net_income` computation includes federal and state deduction which should be adjusted to use a newly created `paycheck_withholding` variable

PavelMakarchuk:
Fixes #3694

### DIFF
--- a/policyengine_us/variables/household/income/spm_unit/spm_unit_paycheck_withholdings.py
+++ b/policyengine_us/variables/household/income/spm_unit/spm_unit_paycheck_withholdings.py
@@ -1,0 +1,15 @@
+from policyengine_us.model_api import *
+
+
+class spm_unit_paycheck_withholdings(Variable):
+    value_type = float
+    entity = SPMUnit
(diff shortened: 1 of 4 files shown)


policyengine_us/variables/household/income/spm_unit/spm_unit_paycheck_withholdings.py
```diff
@@ -0,0 +1,15 @@
...
+class spm_unit_paycheck_withholdings(Variable):
+    value_type = float
+    entity = SPMUnit
+    label = "Paycheck withholdings of the SPM unit"
+    unit = USD
+    definition_period = YEAR
+
+    adds = [
+        "employee_payroll_tax",
+        "state_income_tax_before_refundable_credits",
```
MaxGhenis, line 13: This is OK for now but let's move to before all credits in the future

MaxGhenis (approved):

PR #2577, opened 2023-07-05
South Carolina State Tax Addback

hua7450:
Fixes #2575

### Summary
📱🛠️🌐

Add a new policy parameter `child_allowance` to the US tax and benefit system. Update the `household_net_income` function to include the child allowance in the calculation. Add tests and documentation for the new parameter and function.

> _`is_valid` checks_
> _refactored for clarity_
> _autumn leaves falling_

### Walkthrough
*  Add a new parameter `child_allowance.amount` to the `parameters.yaml` file, which sets the monthly amount of the child allowance benefit per child ()

PR #3745, opened 2024-01-31
Fix South Carolina state tax addback formula

### LINKED ISSUE #3703: Fix South Carolina state tax addback formula
In the 0.632.0 version, the value of the `sc_state_tax_addback` variable is not always correct.  There are [many tests](https://github.com/PolicyEngine/policyengine-us/blob/master/policyengine_us/tests/policy/baseline/gov/states/sc/tax/income/sc_state_tax_addback.yaml) of this variable, which is admirable.   But when I look at the expected output in the tests, I don't understand how the expected results for Line 3 (see the SC addback worksheet below) were calculated.  The expected results in the test make all the tests pass, but it seems to me that the expected Line 3 results are at variance to what you would get filling out the worksheet.   

Perhaps I'm confused and all the tests are accurate.  But if not, this situation illustrates a common pitfall in testing: expected test results are computed using the same misunderstanding of the tax logic as is incorporated in the code.  The pull request review process should catch these kinds of problems.

Here is the 2021 worksheet I'm looking at:

Screenshot 2024-01-26 at 3 05 29 PM

Shinuing:
Fix #3703

### DIFF
--- a/policyengine_us/variables/gov/states/sc/tax/income/additions/sc_state_tax_addback.py
+++ b/policyengine_us/variables/gov/states/sc/tax/income/additions/sc_state_tax_addback.py
@@ -16,18 +16,13 @@ class sc_state_tax_addback(Variable):
 
     def formula(tax_unit, period, parameters):
         p_us = parameters(period).gov.irs.deductions
-        itm_deds = [
-            deduction
-            for deduction in p_us.itemized_deductions
-            if deduction not in ["salt_deduction"]
-        ]
-        us_itemizing = tax_unit("tax_unit_itemizes", period)
+
         standard_deduction = tax_unit("standard_deduction", period)
         filing_status = tax_unit("filing_status", period)
         eligible = filing_status != filing_status.possible_values.SEPARATE
         # line 1
-        federal_itemized_deduction = (
-            add(tax_unit, period, itm_deds) * us_itemizing
+        federal_itemized_deduction = add(
+            tax_unit, period, p_us.itemized_deductions
         )
         # line 2
         federal_standard_deduction = standard_deduction * eligible
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    fixed:
+    - South Carolina state tax addback.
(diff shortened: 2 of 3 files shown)


PavelMakarchuk:
> @martinholmer @PavelMakarchuk Two concerns for the[ old model ](https://github.com/PolicyEngine/policyengine-us/pull/2577/files#diff-0d692520904f4c152eb2ab1ffb9875f611581a0191b0b7114e5ff5742cf24fbf)
> 
> 1. Based on the worksheet line 1, Schedule A line 17 includes all deductions instead of `itemized_deductions_less_salt`
>    image
> 2. The old structure used `tax_unit_itemizes` to check for itemized deductions, which was sweet. However, some test cases provided standard deduction > itemized deduction (line 3), while `tax_unit_itemizes = true` conditioned on itemized deduction > federal deduction.
>    Not sure if we need the  `tax_unit_itemizes` check here, I removed temporarily

Hey Kingsley, the `tax_unit_itemizes` variable checks whether the tax unit has itemized their deductions on the federal return. Usually states either allow a filer to pick the larger of the standard or itemized deduction (in which case we dont use the `tax_unit_itemizes` and just use '>') or some states require the tax unit to apply the same deduction on the state return as they have done on the federal return. E.g. if I itemized my deductions federally I have to itemize on my state return even if on the state return my standard deduction would be more beneficial. In this case we use `tax_unit_itemizes` - let me know if this makes sense @Shinuing

Shinuing:
Yes, thanks for clarification

PavelMakarchuk (approved):

policyengine_us/variables/gov/states/sc/tax/income/additions/sc_state_tax_addback.py
```diff
@@ -16,18 +16,14 @@ class sc_state_tax_addback(Variable):
...
-        ]
+
         us_itemizing = tax_unit("tax_unit_itemizes", period)
         standard_deduction = tax_unit("standard_deduction", period)
         filing_status = tax_unit("filing_status", period)
         eligible = filing_status != filing_status.possible_values.SEPARATE
         # line 1
         federal_itemized_deduction = (
-            add(tax_unit, period, itm_deds) * us_itemizing
+            add(tax_unit, period, p_us.itemized_deductions) * us_itemizing
```
MaxGhenis, line 26:
```suggestion
            tax_unit("itemized_taxable_income_deductions", period) * us_itemizing
```

MaxGhenis (approved):
looks good after this change (please also update the unit test accordingly)

PR #3784, opened 2024-02-06
Fix Oregon Federal Tax Liability Subtraction Calculation

Shinuing:
Testing against the TAXSIM35 model until we find the issue in either model


PavelMakarchuk:
@martinholmer the PR for the integration test mentioned via email - I believe the issue lies with TAXSIM if we are reading your files correctly

martinholmer:
@PavelMakarchuk said in #3784:
> [For] the PR for the integration test mentioned via email, I believe the issue lies with TAXSIM if we are reading your files correctly.

OK.  Can you give a detailed explanation of the reasons you think that?

PavelMakarchuk:
> @PavelMakarchuk said in #3784:
> 
> > [For] the PR for the integration test mentioned via email, I believe the issue lies with TAXSIM if we are reading your files correctly.
> 
> OK. Can you give a detailed explanation of the reasons you think that?

If I am reading this correctly then TAXSIM and PE align on the oregon AGI amount: 
or_agi: $139,803.51
As well as the oregon standard and itemized deductions:
or_standard_deduction: $4,980.00
or_itemized_deductions: 0

This leads us to the following calculation from the [tax form](https://www.oregon.gov/dor/forms/FormsPubs/form-or-40_101-040_2021.pdf#page=3):
image

The calculation that i see then:
$139,803.51 - $4,980.00 = $134,823.51

PavelMakarchuk:
> @PavelMakarchuk, Thanks so much for the detailed explanation in pull request #3784.
> 
> So, summarizing what you found, the two model differ on OR taxable income by $7050.
> 
> Is it possible that PEUS has missed the federal tax subtraction for this case? (If so, the model difference should have shown up in OR AGI, but TAXSIM35 logic is a little loose here and considers the OR federal tax subtraction as a deduction.)
> 
> What do your hand calculations with the OR tax forms/worksheets show as the OR federal tax subtraction for this case? When I look at the .trace file I sent you, it is zero. I don't understand why this case would not get a positive federal tax subtraction. In fact, when I look at the instructions worksheet, I see the OR federal tax subtraction is limited to $7050 for this case. Don't you think it might be more than a coincidence that the maximum OR federal tax subtraction amount and the taxable income difference are the exact same number?

I see, this makes perfect sense I believe we are not considering the [2021 logic](https://www.oregon.gov/dor/forms/FormsPubs/form-or-40-inst_101-040-1_2021.pdf#page=14) for the federal tax subtraction in this case. I dont see anything in the code related to the ctc - will fix, I appreciate the explanation, the approach to debugging makes a lot of sense.

As a general question, would it be dangerous to apply the federal tax subtraction as a deduction as opposed to a subtraction. This might cause differing amounts in the `or_agi` variable which then can cause discrepancies in the calculations like the `or_ctc` which takes `or_agi` as an input.

martinholmer:
@PavelMakarchuk said:
> As a general question, would it be dangerous to apply the federal tax subtraction as a deduction as opposed to a subtraction. This might cause differing amounts in the `or_agi` variable which then can cause discrepancies in the calculations like the `or_ctc` which takes `or_agi` as an input.

Yes, this is an astute observation.  
But it has not come up in our testing because there was no `or_ctc` in 2021.

Shinuing:
Fix #3797

Shinuing:
@PavelMakarchuk Not very sure about  _6. Other taxes (see instructions)_, any variable for this?

Line 6: Enter the total of any other taxes, plus any additions to tax that include only income tax, such as tax credit recaptures, from Form 1040, Schedule 2, lines 8, 10,** 16, and 17. Include any tax on non-effectively connected income from Form 1040-NR, line 23a. Don’t include any additions to tax from Schedule 2, line 17 that are penalties, interest, excise tax, or other amounts that aren’t income tax. If you have no other taxes, enter $0.

PavelMakarchuk:
> @PavelMakarchuk Not very sure about _6. Other taxes (see instructions)_, any variable for this?
> 
> Line 6: Enter the total of any other taxes, plus any additions to tax that include only income tax, such as tax credit recaptures, from Form 1040, Schedule 2, lines 8, 10,** 16, and 17. Include any tax on non-effectively connected income from Form 1040-NR, line 23a. Don’t include any additions to tax from Schedule 2, line 17 that are penalties, interest, excise tax, or other amounts that aren’t income tax. If you have no other taxes, enter $0.

looking at [Schedule 2](https://www.irs.gov/pub/irs-pdf/f1040s2.pdf):
image

I believe we have:
```
alternative_minimum_tax
self_employment_tax
net_investment_income_tax
```

@MaxGhenis please let me know if I missed anything or if we have a variable that sums the form up

MaxGhenis:
Those and `additional_medicare_tax`

Shinuing:
> I believe we have:
> 
> ```
> alternative_minimum_tax
> self_employment_tax
> net_investment_income_tax
> ```

Hi @PavelMakarchuk , I believe in this program we only need a few lines and we don't seem to have those variables
image

policyengine_us/tests/policy/baseline/gov/states/or/tax/income/integration.yaml
```diff
@@ -36,3 +36,45 @@
     or_exemption_credit: 426.00
     or_retirement_credit: 900.00
     or_income_tax: 23.75  # = 1349.75 - 426.00 - 900.00
+
```
Shinuing, line 39: uncomment

PavelMakarchuk:
> You don't need the others

we dont need any of the ones mentioned above right?

PavelMakarchuk:
> @PavelMakarchuk `or_agi` failed in the integration test, others passed
> 
> `Test 'Tax unit with taxsimid 99570 in j21.its.csv and j21.ots.csv': or_agi@2021: [132753.5] differs from 139803.515625 with an absolute margin [7050.0156] > 0.1`

I think the issue is that TAXSIM uses the subtractions from state and not federal AGI, I think we should merge the change and see if we fail integration tests from there. I think it shouldnt be an issue

MaxGhenis (approved):

PR #3813, opened 2024-02-08
Add pre-subsidy childcare care and electricity expenses and use for childcare subsidy rules

### LINKED ISSUE #3765: Add pre-subsidy childcare expenses and use for childcare subsidy rules
Currently we're using `childcare_expenses` or `care_expenses` both to calculate childcare subsidies (Colorado CCAP and CalWORKs Childcare) and also to calculate deductions and credits (CDCC, SNAP dependent care deduction, state tax deductions and credits). The latter are based on post-subsidy childcare expenses, so we should separate out a pre-subsidy childcare expense and make that the input, then use calculated post-subsidy expense for the latter type.

This request mirrors #3729 for rent

### LINKED ISSUE #3766: Add pre-subsidy electricity expenses
Currently we're using electricity_expense both to calculate electricity subsidies (CARE and FERA in California) and also to calculate utility deductions in programs like SNAP. The latter are based on post-subsidy electricity expenses, so we should separate out a pre-subsidy electricity expense and make that the input, then use calculated post-subsidy expense for the latter type.

This request mirrors https://github.com/PolicyEngine/policyengine-us/issues/3729 for rent and #3765 for childcare.

### LINKED ISSUE #3967: DC disabled exclusion subtraction shouldn't be based on all benefits
It says total income but I think that probably means cash benefits, if any. For now let's use IRS gross income in the household, then check with DC on what it includes.

https://github.com/PolicyEngine/policyengine-us/blob/2c0afdbf4f53278a55060f0e2ca8bb7d8d4b4344/policyengine_us/variables/gov/states/dc/tax/income/subtractions/dc_disabled_exclusion_subtraction.py#L25

PavelMakarchuk:
Fixes #3765
Fixes #3766
Fixes #3967

And some other minor fixes to get tests passing


PavelMakarchuk:
@MaxGhenis do we want to add 
CARE 
FERA 
as the relevant subsidies for the utilities? If so that will cause circular reference issues -> same for CO CCAP programs as the subsidy for the childcare expense programs as the computation of CARE, FERA, CCAP and CalWorks require snap and / or WIC as income inputs while those programs depend on now post-subsidy inputs

alll88 (approved):
LGTM! 🚀

MaxGhenis (approved):

PR #3850, opened 2024-02-12
Louisiana integration test comparing against TAXSIM

PavelMakarchuk:
Currently the purpose of this PR is to provide an integration test to find the changes in the LA tax model bteween TAXSIM and PE


PavelMakarchuk:
Fixed in #3978

PR #3866, opened 2024-02-13
Update California SNAP parameters

### LINKED ISSUE #3429: Update California SNAP parameters
We should update all, but start with California for SBN. For example, https://www.snapscreener.com/?p=table shows that California's SUA is $596, but we have $487.

PavelMakarchuk:
Fixes #3429


MaxGhenis (approved):
good to merge after changing effective date to 2023-10-01

PR #3886, opened 2024-02-14
Allocate the Montana deductions and exemptions optimally to the head and spouse if filing separately

### LINKED ISSUE #3885: Allocate the Montana deductions and exemptions optimally to the head and spouse if filing separately

PavelMakarchuk:
Fixes #3885


MaxGhenis (changes requested):
Add two more variables:
* `mt_tax_unit_deductions_exemptions_indiv`: Total deductions and exemptions to be allocated between head and spouse under individual filing
* `mt_head_deductions_exemptions_indiv`: Dollar value of deductions and exemptions allocated to the head under individual filing

Then this main formula does simple subtraction based on if the person is head or spouse.

policyengine_us/variables/gov/states/mt/tax/income/mt_taxable_income_indiv.py
```diff
@@ -15,47 +15,14 @@ class mt_taxable_income_indiv(Variable):
...
-        reduced_agi = (
-            mt_agi
-            - difference_applied_to_larger_agi
-            - halved_deductions_allocated_to_head_or_spouse
+        head_deductions = person.tax_unit(
+            "mt_head_deductions_exemptions_indiv", period
         )
-        return max_(reduced_agi, 0)
+        spouse_deductions = deductions_and_exemptions - head_deductions
+        capped_head_agi = max_(mt_agi - head_deductions, 0)
```
MaxGhenis, line 26: Just a note that I think this approach correctly handles cases where taxable income for one spouse gets capped at zero, since it will appropriately move more to the other.

MaxGhenis (approved):

policyengine_us/variables/gov/states/mt/tax/income/mt_taxable_income_indiv.py
PavelMakarchuk: I believe in cases where the total deductions and exemptions exceed the amount of AGI of both people the taxable income will be below 0 if this is not in place.
See test number 5 in the `mt_taxable_income_indiv.yaml` which fails without the max_

PR #3901, opened 2024-02-15
Reduce the `adjusted_net_capital_gain` by the qualified dividend income in the `vt_capital_gains_exclusion` calculation

### LINKED ISSUE #3900: Reduce the `adjusted_net_capital_gain` by the qualified dividend income in the `vt_capital_gains_exclusion` calcualtion

PavelMakarchuk:
Fixes #3900


MaxGhenis (commented):
Can you give more detail on this change? `adjusted_net_capital_gain` has logic with respect to `qualified_dividend_income` that isn't just adding it so without looking at the form it's not obvious to me this is captures the logic.

PavelMakarchuk:
> Can you give more detail on this change? `adjusted_net_capital_gain` has logic with respect to `qualified_dividend_income` that isn't just adding it so without looking at the form it's not obvious to me this is captures the logic.

image
We are trying to follow the following definition from the VT [Form](https://tax.vermont.gov/sites/tax/files/documents/IN-153%20Instr-2021.pdf) which includes adjusted net capital gains but does not include dividend income. 
Is there a better variable to use that would represent IRC 1 (h)

MaxGhenis:
Right I'm wondering if you should rebuild `adjusted_net_capital_gain` without dividends. See [the formula](https://github.com/PolicyEngine/policyengine-us/blob/master/policyengine_us/variables/gov/irs/tax/federal_income/capital_gains/adjusted_net_capital_gain.py).

PavelMakarchuk:
> Right I'm wondering if you should rebuild `adjusted_net_capital_gain` without qualified dividends. See [the formula](https://github.com/PolicyEngine/policyengine-us/blob/master/policyengine_us/variables/gov/irs/tax/federal_income/capital_gains/adjusted_net_capital_gain.py).

I can recreate `adjusted_net_capital_gain` without adding back the dividend income in the return function, however the comment in the code of adjusted_net_capital_gain suggests that having a `net_capital_gain` variable without dividend income would be more accurate. It also mentions that it will get more nuanced as to which dividends we are not including in this case, not sure how much we want to split the original logic

policyengine_us/variables/gov/states/vt/tax/income/adjusted_gross_income/subtractions/vt_capital_gain_exclusion/vt_percentage_capital_gains_exclusion.py
```diff
@@ -20,10 +20,16 @@ def formula(tax_unit, period, parameters):
         adjusted_net_capital_gain = tax_unit(
             "adjusted_net_capital_gain", period
         )
+        dividend_income = add(tax_unit, period, ["qualified_dividend_income"])
```
MaxGhenis, line 23:
```suggestion
        qualified_dividend_income = add(tax_unit, period, ["qualified_dividend_income"])
```

MaxGhenis (approved):
Good to merge after this and fixing the description in variable `captial`

For posterity: We do **not** remove qualified dividend income from the `net_capital_gain` and `adjusted_net_capital_gain` formulas, where they interact in several places. Instead we follow the VT law with the simpler reduction:

>(ii) with respect to adjusted net capital gain income as defined in 26 U.S.C. § 1(h) **reduced by the total amount of any qualified dividend income**: either the first $5,000.00 of such adjusted net capital gain income or 40 percent of adjusted net capital gain income from the sale of assets held by the taxpayer for more than three years, except not adjusted net capital gain income from:

MaxGhenis (approved):

PR #3940, opened 2024-02-18
Disabled the cdcc cap increase in 2021 for the `id_household_and_dependent_care_expense_deduction`

### LINKED ISSUE #3937: Disabled the cdcc cap increase in 2021 for the `id_household_and_dependent_care_expense_deduction`

PavelMakarchuk:
Fixes #3937

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: minor
+  changes:
+    added:
+    - Disabled the cdcc cap increase in 2021 for the id_household_and_dependent_care_expense_deduction.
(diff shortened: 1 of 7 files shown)


MaxGhenis (changes requested):
can you add a reference? curious if they look forward to 2022 or look backward instead (as is more common)

PavelMakarchuk:
> can you add a reference? curious if they look forward to 2022 or look backward instead (as is more common)

The [legal code](https://law.justia.com/codes/idaho/2022/title-63/chapter-30/section-63-3022d/) does not seem to change in any year from 2018 to 2022, given that justia is accurate 
[Idaho](https://legislature.idaho.gov/statutesrules/idstat/title63/t63ch30/sect63-3022d/) does not allow to backdate 

The information comes from the [tax form](https://tax.idaho.gov/wp-content/uploads/forms/EIN00046/EIN00046_11-15-2021.pdf#page=31)

MaxGhenis:
Looks like they [conformed to the IRC as of 2021-01-01](https://tax.idaho.gov/governance/statutes/irc/), before Biden signed ARPA in March 2021. We can't exactly get that here but tying to 2020 would be closer than 2022.

MaxGhenis (approved):

PR #3951, opened 2024-02-19
Include pension income in the Mississippi AGI calculation

### LINKED ISSUE #3949: Pension income should be included in MS AGI
Pull request #3947, which "fixed" issue #3946, which itself has no description, removed taxable pension income from MS AGI.   Given the 2021 form, and the lack of any rationale for doing this, I'd say this is a bug.   Certainly, the testing results against TAXSIM35 have suffered a regression.

What was the rationale for doing this?

Here is the 2021 form which includes this type of income on line 45:

Screenshot 2024-02-19 at 3 44 04 PM

PavelMakarchuk:
Fixes #3949

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    fixed:
+    - Include pension income in the Mississippi AGI calculation.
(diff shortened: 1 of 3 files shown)


MaxGhenis (approved):
For posterity: this removes the integration test because it was based on a version of TAXSIM that didn't reflect the law correctly.

PR #3972, opened 2024-02-21
Use `adds` for `household_refundable_tax_credits` and `household_tax_before_refundable_credits`

### LINKED ISSUE #3851: Use `adds` for `household_refundable_tax_credits` and `household_tax_before_refundable_credits`
We removed these to avoid confusing/error-prone combinations of `adds` and a `formula`. But it was a bit of a Chersterton's Fence moment: we did that so they would break down in the app. Now we only show total tax and credits. [Example](https://policyengine.org/us/household?focus=householdOutput.netIncome&household=41227):

image

I think we can get around this by applying reforms, e.g. for [`household_refundable_tax_credits`](https://github.com/PolicyEngine/policyengine-us/blob/master/policyengine_us/variables/household/income/household/household_refundable_tax_credits.py) to make a reform for `abolish_federal_income_tax`. I'm not sure how to treat `reported_state_income_tax`, which also shows up in [`household_tax_before_refundable_credits`](https://github.com/PolicyEngine/policyengine-us/blob/master/policyengine_us/variables/household/income/household/household_tax_before_refundable_credits.py)--@nikhilwoodruff ideas?

PavelMakarchuk:
Fixes #3851

### DIFF
--- a/policyengine_us/reforms/federal/reported_state_income_tax.py
+++ b/policyengine_us/reforms/federal/reported_state_income_tax.py
@@ -1,0 +1,68 @@
+from policyengine_us.model_api import *
+
+
+def create_reported_state_income_tax() -> Reform:
+    class household_tax_before_refundable_credits(Variable):
+        value_type = float
+        entity = Household
+        label = "total tax before refundable credits"
+        documentation = "Total tax liability before refundable credits."
+        unit = USD
+        definition_period = YEAR
+
+        def formula(household, period, parameters):
+            p = parameters(period)
+            added_components = p.gov.household_tax_before_refundable_credits
+            added_components = [
+                "employee_payroll_tax",
+                "self_employment_tax",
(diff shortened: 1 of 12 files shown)


MaxGhenis (approved):

policyengine_us/reforms/federal/reported_state_income_tax.py
```diff
@@ -0,0 +1,54 @@
...
+        entity = Household
+        label = "total tax before refundable credits"
+        documentation = "Total tax liability before refundable credits."
+        unit = USD
+        definition_period = YEAR
+
+        adds = [
+            "income_tax_before_refundable_credits",
+            "self_employment_tax",
+            "income_tax_before_refundable_credits",
```
nikhilwoodruff, line 16: @MaxGhenis @PavelMakarchuk - should one of these not be `employee_payroll_tax`?

PR #3989, opened 2024-02-23
Include social security in the HUD annual income variable

### LINKED ISSUE #3988: Include social security in the HUD annual income variable
This is a partial fix to #1053 to include more accurate income definitions but does not include every possible income type.

PavelMakarchuk:
Fixes #3988


MaxGhenis (approved):
Could you add SSI while you're here?

PR #4021, opened 2024-02-28
Uprate IRS parameters from CBO projections

### LINKED ISSUE #657: Uprate payroll tax cap to average wage growth
We're currently citing https://www.ssa.gov/benefits/retirement/planner/maxtax.html

This comes from [Section 230 of the Social Security Act](https://www.ssa.gov/OP_Home/ssact/title02/0230.htm):
>[(b)]() The amount of such contribution and benefit base shall (subject to subsection (c)) be the amount of the contribution and benefit base in effect in the year in which the determination is made or, if larger, the product of—

>[(1)]() $60,600, and

>[(2)]() the ratio of (A) the national average wage index (as defined in section 209(k)(1)) for the calendar year before the calendar year in which the determination under subsection (a) is made to (B) the national average wage index (as so defined) for 1992,

>with such product, if not a multiple of $300, being rounded to the next higher multiple of $300 where such product is a multiple of $150 but not of $300 and to the nearest multiple of $300 in any other case.

### LINKED ISSUE #4014: Uprate standard deduction

MaxGhenis:
Fixes #4014
Fixes #657

Plus other upratings, entering the full set of CBO tax parameter projections and adding some uprating metadata (though without rounding rules).


nikhilwoodruff (approved):

policyengine_us/parameters/gov/irs/income/bracket.yaml
```diff
@@ -342,6 +555,17 @@ thresholds:
         2022-01-01: 340_100
         2023-01-01: 364_200
         2024-01-01: 383_900
+        # CBO forecast.
+        2035-01-01: 393_650
```
MaxGhenis, line 559: No! Max! No!!!

PR #4052, opened 2024-03-01
Eliminate randomness in baseline tax results across model runs

### LINKED ISSUE #3579: Randomness causing small impact of null reforms
For example, [this reform](https://policyengine.org/us/policy?focus=policyOutput.policyBreakdown&reform=44027&region=us&timePeriod=2023&baseline=2) takes effect in 2024 and therefore should have no effect on 2023, yet we're seeing a small budgetary impact. Assume this is due to some household randomness.

image

martinholmer:
Fixes #3579.


MaxGhenis (commented):
Could we structure this as a reform instead? Example: https://github.com/PolicyEngine/policyengine-us/blob/master/policyengine_us/reforms/federal/abolish_payroll_tax.py

PR #4066, opened 2024-03-04
Replace `employment_income` with `irs_employment_income` in `earned_income`

### LINKED ISSUE #4007: Replace `employment_income` with `irs_employment_income` in `earned_income`
`earned_income` is used in IRS computations and should therefore be based on taxable wages and salaries, after pre-tax contributions.

The Maryland TANF also uses `earned_income`, so I think we should move that to use `employment_income` directly.

We may want to rename `earned_income` to `irs_earned_income` to clarify the consideration of pre-tax contributions.

I believe this is the final update to ensure that `employment_income` is only used for payroll taxes and benefits.

SirMalamute:
Fixes #4007

copilot:all

### DIFF
--- a/policyengine_us/variables/gov/irs/income/taxable_income/adjusted_gross_income/irs_gross_income/earned_income/earned_income.py
+++ b/policyengine_us/variables/gov/irs/income/taxable_income/adjusted_gross_income/irs_gross_income/earned_income/earned_income.py
@@ -9,4 +9,4 @@ class earned_income(Variable):
     documentation = "Income from wages or self-employment"
     definition_period = YEAR
 
-    adds = ["employment_income", "self_employment_income"]
+    adds = ["irs_employment_income", "self_employment_income"]


changelog_entry.yaml
PavelMakarchuk, line 1: Which issue do we address if we are changing the Maryland TANF files?

changelog_entry.yaml
```diff
@@ -0,0 +1,5 @@
+- bump: minor
+  changes:
+    changed:
+    - Replaced employment_income with irs_employment_income in earned_income.
+    - Edited Maryland TANF to use employment_income instead of earned_income
```
PavelMakarchuk, line 5:
```suggestion
```

MaxGhenis:
Per Pavel's question about Maryland TANF, I think we should rename it to `irs_earned_income` and use a separate `earned_income` concept for benefit programs that refer to earned income without subtracting pre-tax contributions.

PavelMakarchuk:
> Per Pavel's question about Maryland TANF, I think we should rename it to `irs_earned_income` and use a separate `earned_income` concept for benefit programs that refer to earned income without subtracting pre-tax contributions.

Does this relate to this PR or the MD TANF #4223

PR #4080, opened 2024-03-05
Fix error in `make install`...

### LINKED ISSUE #3712: Documentation failing
In CI

abhcs:
...which should fix #3712.

See https://github.com/PolicyEngine/policyengine-us/issues/3712#issuecomment-1977921069.


Makefile
PavelMakarchuk, line 1: could you add an entry to changelog_entry

MaxGhenis (approved):

PR #4139, opened 2024-03-14
Make the EZ save program household level and use pre-subsidy electricity expenses.

### LINKED ISSUE #4140: Make the EZ save program household level and use pre-subsidy electricity expenses.
Currently we use electricity expenses but the program itself is a subsidy.

PavelMakarchuk:
Fixes #4140


MaxGhenis (approved):
For posterity, `ca_la_ez_save_eligible` was already household-level

PR #4245, opened 2024-03-20
Represent TCJA changes as disabling programs rather than zeroing them out

### LINKED ISSUE #4070: Represent TCJA changes as disabling programs rather than zeroing them out
For instance, we currently represent TCJA's personal exemption repeal as [zeroing out the parameter](https://policyengine.org/us/policy?reform=2&focus=gov.irs.income.exemption.amount&region=us&timePeriod=2024&baseline=2). This doesn't let us apply our standard uprating machinery though. Instead if we disable exemptions, we can let the parameter continue to grow. This will make it easier to conduct TCJA analysis.

Kiki-67:
Fixes #4070


PavelMakarchuk:
Closed for now as we will revisit later

PR #4354, opened 2024-04-09
Verify that the Enhanced CPS doesn't contain PUF-identifiable records

### LINKED ISSUE #4353: Verify that the Enhanced CPS doesn't contain PUF-identifiable records
This should be via a Python script or notebook that checks each record in the ECPS to verify that it is not both:

* In the PUF
* Unique within the PUF

nikhilwoodruff:
Fixes #4353


MaxGhenis:
Works and we can merge but I think this would be more efficient:
1. Identify unique PUF records (limiting to columns we use)
2. Restrict the CPS to records from the PUF (if we happen to recreate unique PUF records from the CPS that's OK)
3. Join tables 1 and 2 and assert the result is empty

PR #4360, opened 2024-04-10
Remove the adds function from the `reported_state_income_tax` reform file

### LINKED ISSUE #4359: Remove the adds function from the `reported_state_income_tax` reform file
Currently we have an adds function as well as a formula in the `create_reported_state_income_tax` reform. The adds function should be deleted

PavelMakarchuk:
Fixes #4359


PavelMakarchuk:
> are you sure this works? the formula uses the elements of the adds

I dont think it does, it uses the list parameter

MaxGhenis (approved):

PR #4365, opened 2024-04-10
Fix Delaware itemized deductions logic

### LINKED ISSUE #4362: Fix Delaware itemized deductions logic
The formula for the `de_deduction_indv` variable does not correspond to what the form instructions say (namely that married filing separately must make the same itemization decision):

https://github.com/PolicyEngine/policyengine-us/blob/db2a86c58739e44fd58d086bfe480085ab53df55/policyengine_us/variables/gov/states/de/tax/income/deductions/de_deduction_indv.py#L4-L16

PavelMakarchuk:
Fixes #4362

### DIFF
--- a/policyengine_us/variables/gov/states/de/tax/income/deductions/de_tax_unit_itemizes.py
+++ b/policyengine_us/variables/gov/states/de/tax/income/deductions/de_tax_unit_itemizes.py
@@ -6,9 +6,9 @@ class de_tax_unit_itemizes(Variable):
     entity = TaxUnit
     label = "Whether the tax unit in Delaware itemizes the deductions when married filing separately"
     definition_period = YEAR
-    defined_for = StateCode.MT
+    defined_for = StateCode.DE
 
     def formula(tax_unit, period, parameters):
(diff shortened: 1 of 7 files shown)


policyengine_us/variables/gov/states/de/tax/income/deductions/de_tax_unit_itemizes.py
```diff
@@ -0,0 +1,14 @@
+from policyengine_us.model_api import *
+
+
+class de_tax_unit_itemizes(Variable):
+    value_type = bool
+    entity = TaxUnit
+    label = "Whether the tax unit in Delaware itemizes the deductions when married filing separately"
```
MaxGhenis, line 7: does this only apply to separate on same return? is there another variable for the choice in other filing statuses?

PavelMakarchuk: we use the same variable in all cases, the label is misleading, changed it

MaxGhenis (approved):

PR #4403, opened 2024-04-15
SSI Couples Countable Income When Both Are Eligible

### LINKED ISSUE #1900: SSI Not Including Both Incomes for Married and Both Eligible
When 2 people are married and both eligible for SSI, only one of their incomes is used when calculating SSI amount. It should calculate both of their countable incomes and then subtract those from the married grant standard.

Example: https://colab.research.google.com/drive/1nayyALu9Y9A-By1Aq-1CTuym1__qazbn#scrollTo=YDeKreCARbZq

CalebPena:
fixes #1900


MaxGhenis:
Michigan home heating credit integration tests are failing because that credit is based on an income measure that includes SSI. Looks like the integration test initially defined a household that (wrongly) didn’t previously qualify for SSI but now does with your fix. I'd suggest setting `ssi: 0` to each person entity in `home_heating/integration.yaml` to get back to the original expected outputs and pass the tests again.

changelog_entry.yaml
```diff
@@ -0,0 +1,4 @@
+- bump: minor
+  changes:
+    fixed:
+    - Countable income now includes the spouses income when both the head and spouse are eligible.
```
MaxGhenis, line 4:
```suggestion
    - Countable income now includes the spouse's income when both the head and spouse are eligible.
```

policyengine_us/variables/gov/ssa/ssi/eligibility/status/is_ssi_ineligible_spouse.py
```diff
@@ -11,4 +11,5 @@ class is_ssi_ineligible_spouse(Variable):
 
     def formula(person, period, parameters):
         spouse = person("is_tax_unit_spouse", period)
-        return spouse & ~person("is_ssi_eligible_spouse", period)
+        eligible = person("is_ssi_aged_blind_disabled", period)
+        return spouse & ~person("is_ssi_eligible_spouse", period) & ~eligible
```
MaxGhenis, line 15: Just for stylistic consistency within the formula
```suggestion
        eligible_spouse = person("is_ssi_eligible_spouse", period)
        return spouse & ~eligible_spouse & ~eligible
```

MaxGhenis (changes requested):

CalebPena:
> Michigan home heating credit integration tests are failing because that credit is based on an income measure that includes SSI. Looks like the integration test initially defined a household that (wrongly) didn’t previously qualify for SSI but now does with your fix. I'd suggest setting `ssi: 0` to each person entity in `home_heating/integration.yaml` to get back to the original expected outputs and pass the tests again.

I think that the couple in this test was eligible for SSI previously, but the spouse's income was not considered, so they had a different SSI amount. Setting SSI to 0 did not work.

MaxGhenis:
Can you set the output to the result for the failing integration tests after setting it to zero? We could calculate it by hand but since they previously worked and your PR doesn't change those programs we can just keep them working.

MaxGhenis (approved):

PR #4210, opened 2024-03-17
Add `economy: false` metadata to SSI asset limit parameters

MaxGhenis:
Fixes #4208

PR #4430, opened 2024-04-21
Add `economy: false` metadata to SSI asset limit parameters

### LINKED ISSUE #4208: Add `economy: false` metadata to SSI asset limit parameters
As we don't currently have assets in the microdata

PavelMakarchuk:
Fixes #4208

Extracted from the PR #4210 trying to isolate individual commits


MaxGhenis:
Interesting, so this commit is the culprit, or at least one of them

PavelMakarchuk:
> Interesting, so this commit is the culprit, or at least one of them

This was the first commit of the PR, not entirely sure how we would debug this besides removing one added file at a time but this won't eliminate the possibility that it is a combination of multiple index files added

PR #4434, opened 2024-04-22
Test the VA itemized deductions structure

PavelMakarchuk:
Using this to showcase local test outputs, not sure if there is an issue yet, will add after we come to a consensus.


PavelMakarchuk:
Not needed anymore

PR #4454, opened 2024-04-25
Reduce the Louisiana exempt income in 2021

### LINKED ISSUE #4453: Reduce the Louisiana exempt income
This only applied in 2021 based on the [tax form](https://revenue.louisiana.gov/TaxForms/IT540i(2021)%20Instructions.pdf#page=9)

PavelMakarchuk:
Fixes #4453


PavelMakarchuk:
I can not find the legal code reference for the reduction rate, as the rate was repealed in 2021 will address in #4455

policyengine_us/parameters/gov/states/la/tax/income/exempt_income/reduction/rate.yaml
MaxGhenis, line 1: rate.yaml confused me as a name - i thought it was a single rate. don't we normally label these marginal structures reduction.yaml? we could then skip the folder and just check if the top rate is nonzero. or leave as-is but rename reduction.yaml within the reduction folder?

MaxGhenis (changes requested):

policyengine_us/parameters/gov/states/la/tax/income/exempt_income/reduction/rate.yaml
PavelMakarchuk: The idea was to have a folder dedicated to everything that is related to the reduction -> one file for the rate and one file for in-effect. I think we have done it similarly in a couple of places

PavelMakarchuk: usually we name the file just `reduction.yaml` when it is the only file related to the reduction aspect of the given program. If there are multiple pieces we sometimes revert to reduction/rate.yaml etc.

MaxGhenis (approved):

PR #4478, opened 2024-05-05
Improve LSR performance and fix bugs

### LINKED ISSUE #4477: Improve performance in labor supply responses
Currently very slow, because we're running 7x as many simulations.

### LINKED ISSUE #4479: Labor supply responses calculate state taxes in baselines
This bug causes a big slowdown in federal simulations, because the baseline simulations we calculate MTRs in don't set state taxes to reported even if the reformed one does.

nikhilwoodruff:
Fixes #4477 
Fixes #4479


nikhilwoodruff:
Tests passing, merging.

PR #4526, opened 2024-05-14
Update core to 2.21.5

### LINKED ISSUE #4516: Fix policyengine-core so that policyengine-us does not crash with fatal error
I am using the 0.770.0 version of `policyengine-us` and the 2.21.1 version of `policyengine-core`, which `setup.py` says is all that is required:

https://github.com/PolicyEngine/policyengine-us/blob/9c5ec52c3817784bd55e550d76dd0305eada11fe/setup.py#L34-L50

.

But yet when I run the MA and OH state income tests (see issue #993), they **all** fail with this error message:
```
File "//Users/mrh/anaconda3/envs/policyengine-us/lib/python3.10/site-packages/policyengine_core/simulations/simulation.py", line 535, in _calculate
    cache_path = self._get_macro_cache(variable_name, str(period))
File "//Users/mrh/anaconda3/envs/policyengine-us/lib/python3.10/site-packages/policyengine_core/simulations/simulation.py", line 1347, in _get_macro_cache
    self.dataset.file_path.parent
AttributeError: 'str' object has no attribute 'parent'
```

PavelMakarchuk:
Fixes #4516

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,4 +1,4 @@
 - bump: minor
   changes:
     added:
-    - Update core to 2.21.1.
+    - Update core to 2.21.5.
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: minor
+  changes:
+    added:
+    - Update core to 2.21.1.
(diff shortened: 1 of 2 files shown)


MaxGhenis (changes requested):
could you update to 2.21.5? looks like we fixed a bug there per https://github.com/PolicyEngine/policyengine-core/blob/master/CHANGELOG.md

MaxGhenis (approved):

PR #4572, opened 2024-05-30
Remove total_income and net_income

### LINKED ISSUE #4548: Remove total_income and net_income
These are empty variables that should instead sum from components. total_income comes from the ACS, net_income from nowhere. Neither is used in any formulas.

### LINKED ISSUE #5106: The label for both `household_net_income` and `net_income` is "net income"
Confusingly, both are labeled as such, so on the front end, it's possible to pull up two different "net income" charts on "Varying your earnings", for example

Kevintowne:
Fixes #4548

Fix #5106


PavelMakarchuk (approved):

MaxGhenis:
@nikhilwoodruff  could you confirm these aren't used somewhere else?

PR #4589, opened 2024-06-01
Round Arkansas deduction allocation fraction to nearest whole percent

### LINKED ISSUE #4586: Round Arkansas deduction allocation fraction to nearest whole percent
In the 0.778.0 version, AR itemized deductions for a married couple are allocated in proportion to head and spouse AGI when they are filing separately,   PEUS does this but fails to round the allocation fraction to the nearest whole percent as required by the following instructions:

AR-allocation-rules

TAXSIM35 does do the required rounding, and therefore, there are a number of tax differences in the k21 sample.

PavelMakarchuk:
Fixes #4586


MaxGhenis (changes requested):
Could you add a test that would previously have failed?

MaxGhenis (approved):

PR #4620, opened 2024-06-09
Adjust the New Mexico refundable credits parameter files

### LINKED ISSUE #4619: Adjust the New Mexico refundable credits parameter files
E.g. the New Mexico CTC should only be applied in 2023

PavelMakarchuk:
Fixes #4619


MaxGhenis (approved):

MaxGhenis (changes requested):
Fix test

PavelMakarchuk:
> should we be just deleting those tests rather than adjusting them?

Generally we should definitely not just remove tests but I think here it is the best way to avoid misleading assumptions about the code structure.  

The tests were checking the structure of zeroing out parameters for the years that the program does not exist. Since this is not practice anymore those tests are obsolete - they would check exactly the same situation for e.g. 2021 and 2022 while there is nothing in the variables and parameters that would lead to the necessity of such tests.

MaxGhenis (approved):

PR #4658, opened 2024-06-18
Updated YAML files for SNAP income deductions for utility expenses.

eqberkovich:


policyengine_us/parameters/gov/usda/snap/income/deductions/utility/limited/allowance.yaml
```diff
@@ -1,184 +1,235 @@
-description: A household is eligible for the Limited Utility Allowance if it does not have separate heating and cooling costs, but does have at least two other utilities. Also known as the 'basic' utility allowance.
+description: A household is eligible for the Limited Utility Allowance if it does
+  not have separate heating and cooling costs, but does have at least two other utilities.
+  Also known as the 'basic' utility allowance.
 AK:
   2010-01-01: 0
   2021-10-01: 0
+  2023-10-01: 0.0
```
PavelMakarchuk, line 7: I dont think we need to values if they do not change as they are set from that first date forward,

In this case just having 
`2010-01-01: 0 ` would be sufficient

policyengine_us/parameters/gov/usda/snap/income/deductions/utility/limited/allowance.yaml
```diff
@@ -1,184 +1,235 @@
...
 AL:
   2010-01-01: 0
   2021-10-01: 389
+  2023-10-01: 434.0
```
PavelMakarchuk, line 11:
```suggestion
  2023-10-01: 434
```

PavelMakarchuk: Just some formatting standards

policyengine_us/parameters/gov/usda/snap/income/deductions/utility/limited/allowance.yaml
PavelMakarchuk: That approach makes sense, I have historically seem reference links used to identify the most recent publications used to determine the parameter value (e.g. in a state model where the tax rate is not adjusted). But I see the value for more complex programs such as snap

policyengine_us/parameters/gov/usda/snap/income/deductions/utility/limited/allowance.yaml
eqberkovich: OK. I'll format it as int (vs. float). To your previous point, this brings up the larger conversation about how PolicyEngine stores and uses parameter values. It may be time to move toward a more complex database or parameters.

PavelMakarchuk: Thank you @eqberkovich, more than happy to get together and discuss options in the near future

eqberkovich: Pavel -- How attached are you to the int vs. float format? It turns out some of the values are floats (there's an 83.96).
The units are currency-USD, so maybe should format as XXX.XX ?

MaxGhenis: Let's continue without the decimal for consistency with the rest of the repo, and discuss a potential switch in https://github.com/PolicyEngine/policyengine-us/discussions/4663

policyengine_us/parameters/gov/usda/snap/income/deductions/utility/limited/allowance.yaml
```diff
@@ -1,184 +1,235 @@
...
   reference:
-    - name: USDA | Food and Nutrition Service | SNAP | Standard Utility Allowances
-      href: https://www.fns.usda.gov/snap/eligibility/deduction/standard-utility-allowances
-    - name: SNAP Screener | SNAP Eligibility Parameters
-      href: https://www.snapscreener.com/?p=table
+  - name: USDA | Food and Nutrition Service | SNAP | Standard Utility Allowances
+    href: https://www.fns.usda.gov/snap/eligibility/deduction/standard-utility-allowances
+  - name: SNAP Screener | SNAP Eligibility Parameters
+    href: https://www.snapscreener.com/?p=table
+  - name: USDA utility allowances by state spreadsheet. Extracted by build_utility_allowances()
```
MaxGhenis, line 231: What's `build_utility_allowances`? Is it code we could include here? @nikhilwoodruff do we have a standard for including such code in the repo?

policyengine_us/parameters/gov/usda/snap/income/deductions/utility/limited/allowance.yaml
MaxGhenis: Yeah I think it makes sense to update values when they haven't changed, if the statutory/regulatory source does so. As Pavel said, many tax parameters don't change year to year because the law hasn't changed, so we don't need to clutter updates in those cases. I'm not sure which applies in this case: Is Alaska actively deciding each year to not have a LUA, or are there laws or regs that haven't been updated? Since the source we have doesn't specify (not a primary source) I think we can repeat as Efraim has done for now.

policyengine_us/parameters/gov/usda/snap/income/deductions/utility/limited/allowance.yaml
eqberkovich: OK. I'll round the floats to ints.

policyengine_us/parameters/gov/usda/snap/income/deductions/utility/limited/allowance.yaml
eqberkovich: build_utility_allowances() is the LLM-based scraper code I am working on.

MaxGhenis:
Please run git pull upstream master and link issue(s)

MaxGhenis:
What's left for this? Should we change our unit tests to monthly?

PavelMakarchuk:
> What's left for this? Should we change our unit tests to monthly?

Need to adjust tests

MaxGhenis:
We were uprating previous values with inflation in October. Now they're flat until 2023.

PavelMakarchuk:
> We were uprating previous values with inflation in October. Now they're flat until 2023.

I see, should we be uprating snap parameters with inflation then if this can happen?

MaxGhenis:
Yes we should. We expect them to rise. But unit tests should not be subject to those projections; we should make them monthly for monthly variables.

PavelMakarchuk:
> Yes we should. We expect them to rise. But unit tests should not be subject to those projections; we should make them monthly for monthly variables.

agreed

MaxGhenis (approved):

PR #4677, opened 2024-06-21
Update to add Python 3.11 support

pmberg:
Adds Python 3.11 support to PolicyEngine US. Consists of edits to setup.py and GitHub workflows, as there are no changes needed otherwise.


nikhilwoodruff (approved):
Thanks!

Issue #4582, opened 2024-05-31
`test_repeal_head_of_household` failing after new version of core


MaxGhenis:
https://github.com/PolicyEngine/policyengine-us/blob/master/policyengine_us/tests/microsimulation/reforms/test_repeal_head_of_household.py

https://github.com/PolicyEngine/policyengine-us/actions/runs/9317273856/job/25653860997?pr=4576#step:5:1673

```
FAILED policyengine_us/tests/microsimulation/reforms/test_repeal_head_of_household.py::test_repeal_head_of_household - AssertionError: 
Not equal to tolerance rtol=1e-07, atol=0

x and y nan location mismatch:
 x: array([ 32459.   ,  44800.367,  19105.   , ..., 157421.36 ,  77087.65 ,
        93635.97 ], dtype=float32)
 y: array([ 32459.   ,  44800.367,  19105.   , ..., 157421.36 ,  77087.65 ,
        93635.97 ], dtype=float32)
====== 1 failed, 2567 passed, 13 skipped, 5 warnings in 131.26s (0:02:11) ======
make: *** [Makefile:10: test] Error 1
```

MaxGhenis:
#4588 pinned core, keeping this open to unpin

pmberg:
Can confirm. I have this issue when I pin PolicyEngine Core 2.22.0 and 2.23.1 in installation, but not with 2.21.8.

PR #4686, opened 2024-06-25
Update to python 3.10

pmberg:
This PR also includes the changes necessary to update to Python 3.11 (which are minimal). These are exclusively to setup.py and the .github workflows folder.

Note that one of the tests fails, but this is a known issue (issue #4582) that this patch does not aim to fix.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,6 +18,8 @@ setup(
         "Operating System :: POSIX",
         "Programming Language :: Python",
         "Programming Language :: Python :: 3.9",
+        "Programming Language :: Python :: 3.10",
+        "Programming Language :: Python :: 3.11",
         "Topic :: Scientific/Engineering :: Information Analysis",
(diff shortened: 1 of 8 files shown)


anth-volk:
Because it's not possible to do any testing locally of the GitHub Actions, if you don't mind, I'm going to push a couple commits on top that I think will fix the deploy errors we're getting with this

setup.py
```diff
@@ -68,7 +70,7 @@
             "survey-enhance",
         ],
     },
-    # Windows CI requires Python 3.9.
+    # Windows CI requires Python 3.9 or above.
     python_requires=">=3.7",
```
MaxGhenis, line 74: should this be 3.9 given the above comment?

MaxGhenis:
Closing as we moved to 3.11 in #4677

PR #4698, opened 2024-06-27
Move DC minimum allotment code to dc folder

### LINKED ISSUE #4687: Move DC minimum allotment code to dc folder
Not a federal/USDA policy so let's put in states/dc instead

e.g.: https://github.com/PolicyEngine/policyengine-us/tree/master/policyengine_us/parameters/gov/usda/snap/temporary_local_benefit/dc

vrathi101:
Fixes #4687


policyengine_us/variables/gov/usda/snap/snap.py
```diff
@@ -23,9 +23,5 @@ def formula(spm_unit, period, parameters):
             return add(
                 spm_unit,
                 period,
-                [
```
PavelMakarchuk, line 26: we still want to keep the sum here as it will be part of final snap variable

PavelMakarchuk (approved):

MaxGhenis (approved):

PR #4724, opened 2024-07-07
Fix 2023 EITC joint bonus parameter value

### LINKED ISSUE #4723: Fix incorrect 2023 federal EITC parameter values
Some of the 2023 EITC parameter values seem to be incorrect.  They differ from the values in the IRS publication that is cited in the EITC parameter files.

Here is a simple test that illustrates the problems:

```
- name: Married filing jointly Texas couple with one child and modest wages
  absolute_error_margin: 0.01
  period: 2023
  input:
    people:
      person1:
        age: 30
        employment_income: 50_000
      person2:
        age: 30
      person3:
        age: 5
    spm_units:
      spm_unit:
        members: [person1, person2, person3]
    tax_units:
      tax_unit:
        members: [person1, person2, person3]
    households:
      household:
        members: [person1, person2, person3]
        state_code: TX  # no state income tax
  output:
    # expected output from IRS publication cited in PEUS EITC parameter files,
    # which is at this URL:
    # https://www.irs.gov/pub/irs-drop/rp-22-38.pdf#page=10
    eitc_maximum: 3_995
    eitc_phase_out_start: 28_120  # fails with PEUS returning 28_130

###############################################################################
# ABOVE TEST IS IN A FILE NAMED case23.yaml
# TEST ENVIRONMENT:
#   % pip list | grep policyengine
#   policyengine-core             2.21.5
#   policyengine-us               1.13.0
# RESULTS FROM RUNNING THIS TEST ARE AS FOLLOWS:
#
#=================================== FAILURES =================================
#_________________________________ test session _______________________________
#/Users/mrh/.../US/case23.yaml:
#  Test 'Married filing jointly Texas couple with one child and modest wages':
#    eitc_phase_out_start@2023: [28130.] differs from 28120.0
#       with an absolute margin [10.] > 0.01
#=========================== short test summary info ==========================
#FAILED case23.yaml::
#============================== 1 failed in 1.23s =============================
```

PavelMakarchuk:
Fix #4723


MaxGhenis (commented):
This seems to trade one accuracy for another (3-person for 0-person). Can we cite the IRC? We may have to restructure this to capture the rules; for instance, Tax-Calculator splits this parameter by the number of children, even though it's almost always the same: https://github.com/PSLmodels/Tax-Calculator/blob/bb2c566ee2e4679591a8f59128ef240c7524397f/taxcalc/policy_current_law.json#L18073

PavelMakarchuk:
> This seems to trade one accuracy for another (3-person for 0-person). Can we cite the IRC? We may have to restructure this to capture the rules; for instance, Tax-Calculator splits this parameter by the number of children, even though it's almost always the same: https://github.com/PSLmodels/Tax-Calculator/blob/bb2c566ee2e4679591a8f59128ef240c7524397f/taxcalc/policy_current_law.json#L18073

The [IRC](https://uscode.house.gov/view.xhtml?req=granuleid:USC-prelim-title26-section32&num=0&edition=prelim) defined the phaseout thresholds for 0, 1 and 2 <= children, with the 1 and 2<= brackets being identical. 

The joint adjustment is a flat increase defined irregardless of number of children. The issue is the inflation adjustments as it involves rounding to the nearest $10 after the joint adjustment, which can differ for each of the two values, so if we were to represent the joint adjustment as one value (as defined in the IRC) we might be off by $10 in some cases (like 2023 where the 0 children bracket is uprated but not the bracket with children). 

But I dont think we a trading accuracy with this adjustment, I think the amounts were flipped for the two brackets in the parameter file (unless I missed something) 

image


image

MaxGhenis:
The value was previously correct for 3 kids and would no longer be

PavelMakarchuk:
> The value was previously correct for 3 kids and would no longer be

In [2023](https://www.irs.gov/pub/irs-drop/rp-22-38.pdf#page=10)? Would it not be 28,120 - 21,560 ($6,560) for three or more kids as it is in this PR: 

image

MaxGhenis (approved):
I'm sorry I totally misread this parameter, didn't realize it already was a scale. Looks good thanks!

PR #4737, opened 2024-07-11
Update tables

### LINKED ISSUE #4273: Support tables 3.9
Currently `tables` 3.9 throws an error, so we pin to 3.8 in #4272

anth-volk:
Fixes #4273

PR updates `tables` to allow some external contributors who had install issues with `tables==3.8.0` to contribute. I haven't run into the error messages mentioned in the issue, which was opened in March.

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,4 +1,4 @@
 - bump: patch
   changes:
     changed:
-    - Updated version of tables
+    - Updated version of tables.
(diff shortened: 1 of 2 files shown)


changelog_entry.yaml
```diff
@@ -0,0 +1,4 @@
+- bump: patch
+  changes:
+    changed:
+    - Updated version of tables
```
PavelMakarchuk, line 4:
```suggestion
    - Updated version of tables.
```

anth-volk: My bad, thanks for catching that

PR #4779, opened 2024-07-24
Restructure Nebraska variables and tests

### LINKED ISSUE #4777: Restructure Nebraska variables
Should follow the structure for other states instead of having all variables under the state folder

vrathi101:
Fixes #4777


vrathi101:
@MaxGhenis this is minor but i noticed that for most/all states we have "{state}_additions.py" and "{state}_agi_subtractions.py", should we specify that "{state}_additions.py" is "{state}_agi_additions.py" or is there another reason for this?

MaxGhenis:
Yeah let's align those

vrathi101:
I think something got messed up when trying to accept incoming changes

PR #4780, opened 2024-07-24
Restructure Nebraska variables and tests

### LINKED ISSUE #4777: Restructure Nebraska variables

vrathi101:
Fixes #4777


PavelMakarchuk (approved):

MaxGhenis (commented):
in other state folders do we typically use `adjusted_gross_income` as the folder, or `agi`? i'd favor the latter unless we've mostly adopted the former

vrathi101:
> in other state folders do we typically use `adjusted_gross_income` as the folder, or `agi`? i'd favor the latter unless we've mostly adopted the former

more states use `adjusted_gross_income` then `agi`

MaxGhenis (approved):

PR #4849, opened 2024-08-09
Update to Python 3.12

### LINKED ISSUE #4848: Update to Python 3.12

vrathi101:
Fixes #4848


MaxGhenis:
You'll need to update scipy as well

setup.py
```diff
@@ -44,7 +45,7 @@
         "pytest-dependency",
         "pyyaml",
         "requests",
-        "scipy==1.10.1",
+        "scipy==1.14.0",
```
MaxGhenis, line 48: If the current works
```suggestion
        "scipy<2",
```

MaxGhenis:
Actually @anth-volk do you recall why you added scipy as a dep [here](https://github.com/PolicyEngine/policyengine-us/commit/0af066f8431a2532c64eb0afa960228b000b34f0)?

We do not currently import it anywhere.

anth-volk:
@MaxGhenis This was due to a bug from about a year ago (I believe Oct. 2023) on deployment. For some reason, the `scipy` version at the time had to be pinned to prevent an upstream dependency error. It may be removable at this point.

setup.py
```diff
@@ -44,7 +45,7 @@
         "pytest-dependency",
         "pyyaml",
         "requests",
-        "scipy==1.10.1",
+        "scipy<2",
```
MaxGhenis, line 48:
```suggestion
```

MaxGhenis (approved):

MaxGhenis:
OK I removed scipy

PR #4859, opened 2024-08-10
Simplify setup.py install_requires and extras_require lists

### LINKED ISSUE #4860: Remove `extras_require`s that are redundant with policyengine-core

martinholmer:
Removes unneeded packages from `setup.py` install_requires list.  The removed packages are required by the `policyengine-core` package, so their inclusion here is redundant.

And also fixes #4860.


MaxGhenis (approved):

setup.py
```diff
@@ -35,19 +34,8 @@
...
     install_requires=[
-        "click==8.1.3",
-        "h5py",
-        "microdf_python",
-        "pandas",
-        "pathlib",
         "policyengine-core",
-        "pytest",
-        "pytest-dependency",
         "pyyaml",
```
MaxGhenis, line 38: we can remove this too - while it's not in -core, it's not imported anywhere

setup.py
```diff
@@ -35,19 +34,8 @@
...
-        "requests",
-        "synthimpute",
-        "tables==3.9.2",
-        "tabulate",
         "tqdm",
     ],
     extras_require={
```
MaxGhenis, line 41: for posterity, i filed #4860 to remove redundant ones here as well. no need to put in this pr though.

MaxGhenis (changes requested):

changelog_entry.yaml
```diff
@@ -0,0 +1,4 @@
+- bump: patch
+  changes:
+    fixed:
+    - Inclusion of unneeded packages in setup.py install_requires list.
```
MaxGhenis, line 4:
```suggestion
      - Inclusion of unneeded packages in setup.py install_requires list.
```

setup.py
MaxGhenis, line 38:
```suggestion
```

setup.py
martinholmer: Making extras list changes in this PR.

setup.py
martinholmer: OK

setup.py
```diff
@@ -35,43 +34,17 @@
...
-        "click==8.1.3",
-        "h5py",
-        "microdf_python",
-        "pandas",
-        "pathlib",
         "policyengine-core",
-        "pytest",
-        "pytest-dependency",
+        "microdf-python",
         "pyyaml",
```
MaxGhenis, line 39:
```suggestion
```

martinholmer: `pyyaml` is apparently never used by policyengine-us.

PR #5001, opened 2024-09-03
Fix the repeal dependent exemption reform

### LINKED ISSUE #5000: Repealing the dependent exemption does not have any impacts
[Example](https://policyengine.org/us/policy?reform=65060&focus=policyOutput.policyBreakdown&region=enhanced_us&timePeriod=2026&baseline=2)

PavelMakarchuk:
Fixes #5000

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,4 +1,4 @@
 - bump: patch
   changes:
     fixed:
-    - Repeal dependent exemptions reform impact.
+    - Five year forward check in the Repeal dependent exemptions reform.
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    fixed:
+    - Repeal dependent exemptions reform impact.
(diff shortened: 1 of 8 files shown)


MaxGhenis (commented):
why didn't the earlier approach work?

PavelMakarchuk:
> why didn't the earlier approach work?

I am not entirely sure, it seems to be the way we compute `tax_unit_size`  but it seems to be fixed by swapping that variable out in `exemptions` for `head_spouse_count`

MaxGhenis:
Can you look more into the previous approach? It'd be much cleaner. Add a unit test to debug?

PavelMakarchuk:
> Can you look more into the previous approach? It'd be much cleaner. Add a unit test to debug?

Five year forward check was not applied, we get impacts if repealing starting 2024 and checking other years 
[Example](https://policyengine.org/us/policy?focus=policyOutput.policyBreakdown&reform=65061&region=enhanced_us&timePeriod=2026&baseline=2)

MaxGhenis (approved):

PR #5088, opened 2024-09-17
Add `defined_for` metadata for all state level variables

### LINKED ISSUE #5083: Calculation issue in various state-level taxes
While doing some testing on the Policy2Code back-end changes, I find that even when calculating outputs for a family from Alabama, I receive a value for both of these variables that does not equal the default value defined in the metadata. There may be some sort of calculation issue here, perhaps not checking to ensure the household lives in Illinois (unless Illinois collects universally, which I wouldn't put past them).

PavelMakarchuk:
Fixes #5083


MaxGhenis (approved):

MaxGhenis (changes requested):
tests failing

PR #5094, opened 2024-09-17
Always use the SUA for CO

### LINKED ISSUE #4246: Colorado SNAP updates
Later this year, the SNAP utility calculations are changing in a way that will increase a HH's utility allowance and thereby decrease countable income. Consistent with federal regulations, if a state agency elects to use a standard utility allowance, that allowance must be made available to HHs that received a payment under LEAP or an other similar energy assistance program.
Starting October 1, CDHS will issue EBT credits to HHs ($20 to help cover utilities + $1 for ATM costs) that didn't get LEAP in the prior 12 months. All eligible SNAP households will get this amount regardless of income, expenses, or HH size. The result is that all Colorado HHs will get the HCUA. The goal is to give HHs the highest possible utility deduction.
The rules will go into effect by August 1 with implementation October 1 (coinciding with when SNAP updates its income / benefits amounts). Details are in this [deck](https://docs.google.com/presentation/d/1LjJz-Ycx1-acefOcEJNtMxVRGfYMLEQgbRxGtEeyYgI/edit#slide=id.g2ab7e3e92af_0_134).

CalebPena:
fixes #4246

@PavelMakarchuk Reposting my questions here:

- Is the HCUA the same as the SUA?
- Do you want to implement this by updating the snap_utility_allowance_type variable?
- Does this apply to people who don't pay utilities?
- Do you have a better source for this?


policyengine_us/variables/gov/usda/snap/income/deductions/shelter/snap_always_standard_utility_allowance.py
```diff
@@ -0,0 +1,14 @@
+from policyengine_us.model_api import *
+
+
+class snap_always_standard_utility_allowance(Variable):
```
PavelMakarchuk, line 4:
```suggestion
class snap_state_using_standard_utility_allowance(Variable):
```

policyengine_us/variables/gov/usda/snap/income/deductions/shelter/snap_always_standard_utility_allowance.py
```diff
@@ -0,0 +1,14 @@
...
+    value_type = bool
+    entity = SPMUnit
+    definition_period = MONTH
+    label = "Always use the standard utility allowance"
```
PavelMakarchuk, line 8:
```suggestion
    label = "Whether a state always uses the standard utility allowance"
```

policyengine_us/variables/gov/usda/snap/income/deductions/shelter/snap_always_standard_allowance.py
PavelMakarchuk, line 1: could you remove this file?

changelog_entry.yaml
```diff
@@ -0,0 +1,4 @@
+- bump: patch
+  changes:
+    added:
+      - Always use the SUA for CO
```
PavelMakarchuk, line 4:
```suggestion
      - Always use the SUA for CO.
```

PavelMakarchuk (changes requested):
LGTM after these, could you also run a `make install` core update and an upstream command, should fix the test issues

policyengine_us/parameters/gov/usda/snap/income/deductions/utility/always_standard.yaml
```diff
@@ -0,0 +1,116 @@
...
+WI:
+  2024-10-01: false
+WV:
+  2024-10-01: false
+WY:
+  2024-10-01: false
+metadata:
+  unit: bool
+  period: month
+  label: States that always use the SUA
```
PavelMakarchuk, line 111:
```suggestion
  label: SNAP States using SUA
```

policyengine_us/parameters/gov/usda/snap/income/deductions/utility/always_standard.yaml
```diff
@@ -0,0 +1,116 @@
+description: These states always use the standard utility allowance
```
PavelMakarchuk, line 1:
```suggestion
description: The following states always use the standard utility allowance under the SNAP program.
```

PavelMakarchuk (changes requested):
TEst seems to fail - might be a missing state code?

CalebPena:
> TEst seems to fail - might be a missing state code?

The `ME` state code is included in [always_standard.yaml](https://github.com/PolicyEngine/policyengine-us/pull/5094/files#diff-87100aafe5c3fc623d06b074fc02a6a5dc7a2e14334d9e9dfc292957318d53fa). Is there another state code that you think is missing?

policyengine_us/parameters/gov/usda/snap/income/deductions/utility/always_standard.yaml
```diff
@@ -0,0 +1,116 @@
+description: The following states always use the standard utility allowance under the SNAP program.
+AK:
+  2024-10-01: false
```
MaxGhenis, line 3: I think the tests are failing because they're running on 2023 and the parameter values are unspecified. Please leave a comment indicating when you're confident the values are correct for.
```suggestion
  2023-01-01: false
```

CalebPena: I just backdated to 2020, although I only know about Colorado. Have y'all thought about adding the ability to add a default value for breakdown parameters, so we would not have to define every state, as `false`?

MaxGhenis (approved):

PR #5129, opened 2024-09-25
Apply the TCJA mortgage interest deduction limits

### LINKED ISSUE #5128: Apply the TCJA mortgage interest deduction limits
We currently only account for pre-TCJA limits 
[Law](https://www.law.cornell.edu/uscode/text/26/163)

PavelMakarchuk:
Fixes #5128


MaxGhenis (changes requested):
The US caps the mortgage value to which interest is applied, not the interest itself. It's also based on the year of purchase not tax year. 

We can't model this today so could we change the logic to deduct the full value of a new variable for deductible mortgage interest? Keep the policy parameter but don't apply it. Check us-data too.

MaxGhenis:
still need to update the description and the variable names - `mortgage_interest` implies total

PavelMakarchuk:
> still need to update the description and the variable names - `mortgage_interest` implies total

adjusted

policyengine_us/parameters/gov/irs/deductions/itemized/interest/mortgage/cap.yaml
```diff
@@ -1,21 +1,33 @@
-description: IRS limits the total home mortgage interest deduction to this amount, based on filing status.
+description: IRS limits the total home mortgage value to this amount, based on filing status under the mortgage interest deduction.
```
MaxGhenis, line 1:
```suggestion
description: The IRS allows an itemized deduction for interest paid on mortgages up to this amount, based on filing status.
```

MaxGhenis (approved):

PR #5137, opened 2024-09-25
Determine random numbers from entity IDs

### LINKED ISSUE #5135: Move all randomness to microdata
See also #828 #831 #3462 https://github.com/PolicyEngine/policyengine-us-data/issues/84

nikhilwoodruff:
Fixes #5135

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: minor
+  changes:
+    added:
+    - Deterministic takeup based on entity IDs.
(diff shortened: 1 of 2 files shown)


MaxGhenis (commented):
Why overwrite [`policyengine-core`'s `random`](https://github.com/PolicyEngine/policyengine-core/blob/2e177cc71c4cd7542fabaf47c5ce49be42db4fe8/policyengine_core/commons/formulas.py#L303)? Seems opaque and error-prone. Can we make the change in core instead?

nikhilwoodruff:
OK moving to Core.

PR #5147, opened 2024-09-26
Alaska Permanent Fund Dividend

### LINKED ISSUE #5132: Alaska Permanent Fund Dividend
The IRS requires recipients to enter this on line 8g of the Form 1040: https://www.irs.gov/pub/irs-pdf/f1040s1.pdf

It's probably not in the CPS: https://forum.ipums.org/t/alaska-permanent-fund-dividend-recipients-in-cps-asec-which-variable-would-give-the-info/2500

Parts are classified as energy rebates and thus not taxable; other parts are taxable: https://pfd.alaska.gov/payments/tax-information

MaxGhenis:
Fixes #5132

### DIFF
--- a/policyengine_us/parameters/gov/household/household_benefits.yaml
+++ b/policyengine_us/parameters/gov/household/household_benefits.yaml
@@ -14,6 +14,10 @@ values:
     - unemployment_compensation
     - head_start
     - early_head_start
+    # One-time energy relief payments.
+    # Paid at the same time as the Alaska Permanent Fund Dividend,
+    # which is part of IRS gross income.
+    - ak_energy_relief
     # Contributed.
     - basic_income
     - spm_unit_capped_housing_subsidy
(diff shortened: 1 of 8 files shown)


PavelMakarchuk (approved):
LGTM after test fixes

policyengine_us/parameters/gov/household/household_benefits.yaml
```diff
@@ -14,6 +14,10 @@ values:
     - unemployment_compensation
     - head_start
     - early_head_start
+    # One-time energy relief payments.
```
PavelMakarchuk, line 17: actually, I think we should add it to the 2022 benefit list 

And I believe the entire 2023 list can be backdated to 2022

policyengine_us/parameters/gov/irs/gross_income/sources.yaml
```diff
@@ -21,6 +21,8 @@ values:
     - illicit_income
     - taxable_retirement_distributions
     - miscellaneous_income
+    # Listed separately on 1040 line 8g.
+    - ak_permanent_fund_dividend
 metadata:
   unit: currency-USD
```
PavelMakarchuk, line 27:
```suggestion
  unit: list
  period: year
```

While you are in here - and changing the list date, same as above

PR #5159, opened 2024-09-29
Add `md_local_income_tax_before_credits` to `local_income_tax`

### LINKED ISSUE #5157: Add `md_local_income_tax_before_credits` to `local_income_tax`
So it goes into SALT. Currently local_income_tax is only nyc.

vrathi101:
Fixes #5157

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: minor
+  changes:
+    added:
+    - Add md_local_income_tax_before_credits to local_income_tax.
--- a/policyengine_us/variables/gov/local/tax/income/local_income_tax.py
+++ b/policyengine_us/variables/gov/local/tax/income/local_income_tax.py
@@ -8,4 +8,4 @@ class local_income_tax(Variable):
     label = "Local income tax"
     unit = USD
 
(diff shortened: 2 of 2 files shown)


MaxGhenis:
We'll need to estimate it as we do for other withheld state income tax numbers

vrathi101:
> We'll need to estimate it as we do for other withheld state income tax numbers

there is currently logic in the Maryland local variable, is anything else needed to be included in the definition?

MaxGhenis:
I mean something like this https://github.com/PolicyEngine/policyengine-us/blob/master/policyengine_us/variables/gov/states/md/tax/income/md_withheld_income_tax.py

PavelMakarchuk:
> I mean something like this https://github.com/PolicyEngine/policyengine-us/blob/master/policyengine_us/variables/gov/states/md/tax/income/md_withheld_income_tax.py

@MaxGhenis if I am understanding correctly, we need to adjust `md_local_income_tax_before_credits` to not use `md_taxable_income` but to apply a new estimation, something like (agi_person - SD (single)) * rates ?

MaxGhenis:
Not adjust the existing variable, but create a new withheld variable used for the SALT deduction as we do for state income taxes

PR #5162, opened 2024-09-30
Fix abolitions

### LINKED ISSUE #5161: Enabling an abolition and running simulation causes app to break
This may only be the case if the abolition is enabled, but nothing else; I vaguely remember running an abolition alongside a policy and not having an issue.

If this issue emanates from another package, it will be transferred there.

anth-volk:
Fixes #5161.

Previously, the abolitions folder was correctly added to the US model, but when activating an abolition parameter, the application would crash. This PR moves up the adding of the abolitions parameters such that they're added before the application of a reform. This also adds a basic test for abolishing the `adjusted_gross_income` variable, but more can be added if desired.

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    changed:
+    - Moved loading of abolitions parameters earlier in initialization process
--- a/policyengine_us/system.py
+++ b/policyengine_us/system.py
@@ -48,6 +48,7 @@ class CountryTaxBenefitSystem(TaxBenefitSystem):
     def __init__(self, reform=None):
         super().__init__(entities, reform=reform)
         self.load_parameters(COUNTRY_DIR / "parameters")
+        self.add_abolition_parameters()
         if reform:
             self.apply_reform_set(reform)
         self.parameters = set_irs_uprating_parameter(self.parameters)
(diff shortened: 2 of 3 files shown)


MaxGhenis:
Thanks for the fix. Still would very much like to move to a list of neutralized variables in the API to avoid this and align the API with the Python package.

anth-volk:
@nikhilwoodruff mentioned that getting this to work was a high priority due to the fact that it blocked an external user, which is why I kept this fix limited. 

Could you describe what you mean by a list of neutralized variables? As in, manually defining every single abolish-able variable? 

And would this be maintaining the current front-end interface, or also creating a simplified component that would allow a user to select from that list? I find that the interface and search box are both far too cluttered with "Abolish X"-es.

MaxGhenis:
Python users use `neutralize_variable` in their reform objects, rather than toggling parameters that then apply a neutralization. I'd suggest we move to a similar paradigm in the API, where the reform is an object that can include parameter modifications and/or neutralizations, applied as a list of variables.

anth-volk:
In that case, I'd need @nikhilwoodruff to weigh in on how to proceed on this. As I said, he indicated that an external user uses our existing "abolish X" parameters for policy simulation.

I see the logic in what's described above, but I question how feasible it would be without pretty significant alterations to the API. This would also require the user somehow being able to interact with the system's variables, which are never exposed to them, and if they were, I think the system's already steep learning curve would grow rapidly.

MaxGhenis:
I'm not suggesting we tackle it in this PR, which we should merge upon Nikhil's approval.

nikhilwoodruff (approved):

anth-volk:
> I'm not suggesting we tackle it in this PR, which we should merge upon Nikhil's approval.

Gotcha, I wasn't sure

PR #5196, opened 2024-10-09
Add compatibility settings for editable installs

### LINKED ISSUE #5195: Add compatibility settings for editable installs
pip install -e ".[dev]"  --config-settings editable_mode=compat

anth-volk:
Fixes #5195.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ format:
 	black . -l 79
 	linecheck . --fix
 install:
-	pip install -e .[dev]
+	pip install -e ".[dev]" --config-settings editable_mode=compat
 test:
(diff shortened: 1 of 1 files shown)


MaxGhenis:
Do we still want to do this?

anth-volk:
@nikhilwoodruff Could comment more, but I'd imagine so. If we ever switch to a `pyproject.toml` or upgrade to `pip` higher than 24, local editable installs require these changes.

PavelMakarchuk:
> @PavelMakarchuk Do you ever run this package in editable? If so, are you running into any problems when running it side-by-side with another, using one shared environment? If you are, then I think this is valuable, otherwise no

Wym by editable? But I don't think so.

PR #5201, opened 2024-10-10
Fix NJ EITC bug

### LINKED ISSUE #5200: NJ EITC doesn't correctly calculate federal EITC entitlement
The NJ EITC has an EITC 'match' for filers who are only not entitled to the federal EITC because of their age (they still need to meet the income tests). But the formula currently gives this to anyone not entitled where their age was *one of* the reasons why.

nikhilwoodruff:
Fixes #5200


MaxGhenis:
tests failing

MaxGhenis (approved):

PR #5236, opened 2024-10-15
Refactor the Missouri income tax parameter structure

### LINKED ISSUE #5211: Refactor the Missouri income tax parameter structure
We currently use the state code in the parameter names which should be consolidated to match other state income tax models

leehengpan:
Fix #5211


PavelMakarchuk (changes requested):
@leehengpan you have created this from a branch - can we you re-create this PR from the master - we dont want the reform changes to also be applied here 

Also could you add a change log and run a formatting command

PR #5260, opened 2024-10-22
Adjust the separate value in the Harris capital gains reform

### LINKED ISSUE #5239: Adjust the separate value in the Harris capital gains reform
Currently set to 1,000,000 but should be 500k

PavelMakarchuk:
Fixes #5239


MaxGhenis (changes requested):
Add reference

PR #5276, opened 2024-10-28
Split SSI couple amount for OAP

### LINKED ISSUE #5275: Fix OAP for married people who both qualify for SSI.
Currently PE assigns SSI for married couples who both qualify for SSI to one of the members. This results in OAP being 0 for one member and full for the other. Instead, we should use half of the couple's SSI amount for each member of the marital unit.

CalebPena:
fixes #5275

### DIFF
--- a/policyengine_us/variables/gov/states/co/ssa/oap/co_oap.py
+++ b/policyengine_us/variables/gov/states/co/ssa/oap/co_oap.py
@@ -10,8 +10,11 @@ class co_oap(Variable):
 
     def formula(person, period, parameters):
         income = person("ssi_countable_income", period)
+        is_joint = person("ssi_claim_is_joint", period)
         ssi = person("ssi", period)
-        total_countable_income = ssi + income
+        # The SSI variable assigns SSI to one member of the marital unit if both are eligible
+        marital_ssi = where(is_joint, person.marital_unit.sum(ssi) / 2, ssi)
+        total_countable_income = marital_ssi + income
         p = parameters(period).gov.states.co.ssa.oap
         grant_standard = p.grant_standard * MONTHS_IN_YEAR
         return max_(0, grant_standard - total_countable_income)
--- a/policyengine_us/tests/policy/baseline/gov/states/co/ssa/oap/co_oap.yaml
+++ b/policyengine_us/tests/policy/baseline/gov/states/co/ssa/oap/co_oap.yaml
@@ -4,6 +4,7 @@
     co_oap_eligible: true
     ssi_countable_income: 1_000
     ssi: 0
+    ssi_claim_is_joint: false
   output:
     co_oap: 10_424
 
@@ -13,6 +14,7 @@
     co_oap_eligible: true
     ssi_countable_income: 0
     ssi: 2_000
+    ssi_claim_is_joint: false
   output:
     co_oap: 9_424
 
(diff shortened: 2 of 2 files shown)


policyengine_us/variables/gov/states/co/ssa/oap/co_oap.py
```diff
@@ -10,8 +10,11 @@ class co_oap(Variable):
 
     def formula(person, period, parameters):
         income = person("ssi_countable_income", period)
+        is_joint = person("ssi_claim_is_joint", period)
         ssi = person("ssi", period)
-        total_countable_income = ssi + income
+        # The SSI variable assigns SSI to one member of the marital unit if both are eligible
+        marital_ssi = where(is_joint, person.marital_unit.sum(ssi) / 2, ssi)
```
PavelMakarchuk, line 16:
```suggestion
        marital_ssi = where(is_joint, person.household.sum(ssi) / 2, ssi)
```
I think we want to sum it by household

CalebPena: SSI is a per member benefit except for the marital unit. So, if the child qualifies for SSI, we would not want that to be counted.

MaxGhenis: Could you add a source for this? AFAIUI, spouses could get different portions of the marital unit's total SSI, based on the personal income of each spouse; it's not automatically split 50/50.

CalebPena: My understanding is that if both spouses are eligible for SSI, they would only receive one check.

FYI, I just realized that I used the wrong variable. Just changed it to use `ssi_marital_both_eligible`

CalebPena:
@MaxGhenis Can you review this pr?

policyengine_us/variables/gov/states/co/ssa/oap/co_oap.py
MaxGhenis: Could you share a source on that?

MaxGhenis: What part of that? This seems related but doesn't back this up AFAICT:
>Deduct an amount to meet the needs of each dependent child living in the household of
the spouse equal to half the maximum SSI grant benefit standard

Could you please provide a full justification for this change in #5275? We prefer to document rationale in issues especially for nontrivial changes here. I'm not yet convinced so please include legal sources and pointers to our current code that needs correction.

CalebPena: @MaxGhenis Sorry for the delay on this; we had to do some more research. I was wrong about couples receiving one check. According to the [handbook](https://www.ssa.gov/OP_Home/handbook/handbook.01/handbook-0131.html), couples receive separate checks. We also compiled some more detailed notes [here](https://docs.google.com/document/d/1D-pKsKqNIRhu7ksXIcAXMHBH9aX-g_-Nz9dgx2T8pmo/edit?usp=sharing).

PavelMakarchuk (approved):

policyengine_us/variables/gov/states/co/ssa/oap/co_oap.py
MaxGhenis: Please justify the change in #5275

CalebPena:
@PavelMakarchuk @MaxGhenis I realized that AND-CS (CO state supplement) has the same issue as OAP, so I added that change to this pr.

PavelMakarchuk:
> @PavelMakarchuk @MaxGhenis I realized that AND-CS (CO state supplement) has the same issue as OAP, so I added that change to this pr.

@CalebPena could you adjust the `co_state_supplement` unit test as well

PR #5280, opened 2024-10-29
Alter VA reduced itemized deductions based on changes to TCJA

### LINKED ISSUE #5279: `va_reduced_itemized_deductions` doesn't properly handle changes to federal income tax policy
Unclear yet as to what, but if running locally and following the below link, it appears that `wrappedJsonParse` fails. This is a bit strange, since all it's doing is running `JSON.parse()`, with an added reviver to convert `String`-typed `"Infinity"` into `Number`-typed `Infinity`.

https://policyengine.org/us/policy?focus=policyOutput.policyBreakdown&reform=68885&region=enhanced_us&timePeriod=2027&baseline=2

anth-volk:
Fixes #5279 

Virginia follows pre-TCJA tax policy, but does not follow its removal of two tax parameters. To keep Virginia calculations in line with federal policy (minus TCJA), we previously hardcoded a method of determining the value of these parameters based on date. However, TCJA extension policies that set these params to `inf` in order to simulate continued nullification of said params were not properly handled by the hardcoded date, causing `NaN` issues for certain calculations that then crashed the front end.

This PR attempts to better represent Virginia's response to the TCJA by instead taking the standard federal params and filtering `inf` values out. In this way, we create two custom internal parameters that track the federal ones, but aren't explicitly tied to them.

This PR also adds a test for one of the offending years, 2027, to ensure that the changes work properly.


PavelMakarchuk (approved):
LGTM

MaxGhenis (approved):
Merging as it's causing issues, though filed #5283 to add statutory references and ensure this logic aligns

PR #5285, opened 2024-10-30
Remove uprating from some SNAP parameters

### LINKED ISSUE #5284: Don't uprate SNAP standard medical deduction and SNAP standard utility allowance
@MaxGhenis What do you think about removing the uprating on the following (there may be others as well, but these are the ones that I found are incorrect for 2025 in CO) because they don't appear to be updated each year:

- [SNAP standard medical deduction](https://github.com/PolicyEngine/policyengine-us/blob/master/policyengine_us/parameters/gov/usda/snap/income/deductions/excess_medical_expense/standard.yaml)
- [SNAP standard utility allowance](https://github.com/PolicyEngine/policyengine-us/blob/master/policyengine_us/parameters/gov/usda/snap/income/deductions/utility/standard/main.yaml)

CalebPena:
Fixes #5284

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    fixed: 
+      - remove uprating for SNAP variables that don't change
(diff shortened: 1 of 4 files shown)


MaxGhenis (changes requested):
Needs some test updates

MaxGhenis (approved):

PR #5318, opened 2024-11-08
Correct meets_school_meal_categorical_eligibility for vectorized inputs

### LINKED ISSUE #5317: Bug in `meets_school_meal_categorical_eligibility`
Visible from this household: https://policyengine.org/us/household?focus=householdOutput.earnings&household=49385

image

A few things to note:
1. This is immediately reproducible and confirmed across two systems running two different browsers
2. This is very likely a duplicate of https://github.com/PolicyEngine/policyengine-api/issues/1792, but not 100% sure - here and there, there's a difference of $127 between the line and dot
3. I believe this issue is separate from the one solved through https://github.com/PolicyEngine/policyengine-core/pull/288, https://github.com/PolicyEngine/policyengine-core/pull/294, and related fixes in the API

anth-volk:
Fixes #5317 

See the write-up in #5317 that explains how this bug functions. This PR changes how we evaluate whether or not someone is categorically eligible for free school meals, checking whether the total value they receive from eligible programs is greater than 0. This is supported for Booleans (since True is cast to 1) and supported by an added test.

### DIFF
--- a/policyengine_us/variables/gov/usda/school_meals/meets_school_meal_categorical_eligibility.py
+++ b/policyengine_us/variables/gov/usda/school_meals/meets_school_meal_categorical_eligibility.py
@@ -12,4 +12,4 @@ class meets_school_meal_categorical_eligibility(Variable):
     def formula(spm_unit, period, parameters):
         p = parameters(period).gov.usda.school_meals
         programs = add(spm_unit, period, p.categorical_eligibility)
-        return np.any(programs)
+        return programs > 0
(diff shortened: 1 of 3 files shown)


policyengine_us/variables/gov/usda/school_meals/meets_school_meal_categorical_eligibility.py
```diff
@@ -12,4 +12,4 @@ class meets_school_meal_categorical_eligibility(Variable):
     def formula(spm_unit, period, parameters):
```
MaxGhenis, line 12: Can simplify further
```suggestion
    adds = "gov.usda.school_meals"
```

anth-volk: But this variable returns a Boolean. Would that still work with this structure?

MaxGhenis: Yes, sums get currently cast to bools. We use that paradigm elsewhere. Could be clearer about it though.

anth-volk: Will modify and re-test Sunday

anth-volk:
Pushed the changes and confirmed they ran locally

MaxGhenis (approved):

PR #5394, opened 2024-12-09
Model take-up of itemized medical deduction

### LINKED ISSUE #5378: Model take-up of itemized medical deduction
This paper estimates a 50.5% take-up rate: https://www.nber.org/papers/w33213

vrathi101:
Fixes #5378


MaxGhenis (changes requested):
Please see EITC take up, we no longer assign it in this repo but instead in us-data. This will require a pair of prs

PR #5401, opened 2024-12-12
Patch structural reforms

### LINKED ISSUE #5400: Structural reforms fail when not starting in 2024
E.g., when running a structural reform with parameters starting in 2026 over 2026, no impact is produced: https://policyengine.org/us/policy?focus=policyOutput.policyBreakdown&reform=72044&region=us&timePeriod=2026&baseline=2&dataset=enhanced_cps

anth-volk:
Fixes #5400.

This is a temporary patch for structural reforms failing when their parameters are initialized on a date other than 2024-01-01. This is meant to be supplanted by a refactor of structural reforms, which we aim to do in the coming couple of months.

The root of the issue is that, when instantiating a new `Simulation` or `Microsimulation`, structural reforms are conditionally applied based on the value of their parameters at a given date. This was previously hard-coded to 2024-01-01, and so in #5400, a reform instantiated in 2026 would not be treated as "active" by the `Microsimulation` initializer.

This code makes the start instant a passable argument, allowing a user to override the default behavior and, in the example in #5400, set "2026-01-01" as the start instant. The effect of this change is that structural reforms will require this parameter to be passed to both the baseline and reform `Microsimulation` instances that we typically initialize. If a given reform contains multiple structural reforms with multiple different start dates, multiple `Microsimulation` instances will be required.

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    added:
+    - Capability to select custom start time for simulations; this is a patch for structural reforms that occur at non-default time periods.
--- a/docs/usage/structural_reform_dating.ipynb
+++ b/docs/usage/structural_reform_dating.ipynb
@@ -1,0 +1,146 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Inputting Dates for Structural Reforms\n",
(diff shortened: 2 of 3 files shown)


PavelMakarchuk:
Could you add an example usage notebook?

PavelMakarchuk (approved):

MaxGhenis (approved):

PR #5412, opened 2024-12-19
Update SSI, AND-CS and OAP values for 2025

### LINKED ISSUE #5413: Update SSI, AND-CS and OAP values for 2025
cc @CalebPena

CalebPena:
Fix #5413


MaxGhenis (approved):
Verified these values match the real-time webpages.

I also filed https://github.com/PolicyEngine/policyengine-us/issues/5414 around uprating the Colorado programs.

PR #5418, opened 2024-12-23
Adjust the EPP max pregnancy month value

### LINKED ISSUE #5417: Adjust the EPP max pregnancy month value
Currently at 6 but it should be 9

PavelMakarchuk:
Fixes #5417

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    fixed:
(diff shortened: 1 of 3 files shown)


MaxGhenis (approved):
Make sense given it's the three months before due date and the min year is 6 https://github.com/PolicyEngine/policyengine-us/blob/master/policyengine_us/parameters/gov/local/ca/la/dss/expectant_parent_payment/pregnancy_month/min.yaml

PR #5425, opened 2024-12-30
Adjust the ia_alternate_tax_unit formula to avoid using np.inf

### LINKED ISSUE #5424: Adjust the `ia_alternate_tax_unit` formula to avoid using np.inf
When running a remove head of household reform - some household return a net income value of N/A which is caused but [this](https://github.com/PolicyEngine/policyengine-us/blob/master/policyengine_us/variables/gov/states/ia/tax/income/ia_alternate_tax_unit.py) formula - not entirely sure why this is but when setting the value manually to e.g. $1bn the issue disappears 

The easiest solution seems to hard code the value for now but will think about alternative solutions

PavelMakarchuk:
Fixes #5424


policyengine_us/variables/gov/states/ia/tax/income/alternate_tax/ia_alternate_tax_eligible.py
```diff
@@ -0,0 +1,20 @@
...
+        "https://tax.iowa.gov/sites/default/files/2022-01/IA1040%2841-001%29.pdf"
+        "https://tax.iowa.gov/sites/default/files/2023-01/2021%20Expanded%20Instructions_010323.pdf#page=53"
+        "https://tax.iowa.gov/sites/default/files/2023-01/2022IA1040%2841001%29.pdf"
+        "https://tax.iowa.gov/sites/default/files/2023-03/2022%20Expanded%20Instructions_022023.pdf#page=53"
+    )
+    defined_for = StateCode.IA
+
+    def formula(tax_unit, period, parameters):
+        filing_status = tax_unit("filing_status", period)
+        is_single = filing_status == filing_status.possible_values.SINGLE
```
MaxGhenis, line 19:
```suggestion
        return filing_status ~= filing_status.possible_values.SINGLE
```

policyengine_us/variables/gov/states/ia/tax/income/alternate_tax/ia_alternate_tax_eligible.py
```diff
@@ -0,0 +1,20 @@
...
+        return where(is_single, False, True)
```
MaxGhenis, line 20: don't need this anymore but fyi you could do `return ~is_single`

MaxGhenis (changes requested):

policyengine_us/variables/gov/states/ia/tax/income/alternate_tax/ia_alternate_tax_eligible.py
MaxGhenis: did this not work?

PavelMakarchuk: Worked with a small adjustment

MaxGhenis (approved):

PR #5494, opened 2025-01-19
Reform - limit the SALT deduction to property taxes

### LINKED ISSUE #5493: Reform - limit the SALT deduction to property taxes

PavelMakarchuk:
Fixes #5493

### DIFF
--- a/policyengine_us/reforms/salt_phase_out/salt_phase_out_reform.py
+++ b/policyengine_us/reforms/salt_phase_out/salt_phase_out_reform.py
@@ -13,14 +13,14 @@ def create_salt_phase_out() -> Reform:
         reference = "https://www.law.cornell.edu/uscode/text/26/164"
 
         def formula(tax_unit, period, parameters):
-            salt_amount = add(
-                tax_unit,
-                period,
-                ["state_and_local_sales_or_income_tax", "real_estate_taxes"],
-            )
             p = parameters(
                 period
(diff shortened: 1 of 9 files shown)


MaxGhenis (changes requested):
Please change the changelog to minor, this isn't a bug fix it adds a feature

Can you also check we can't do this with a parametric reform by making it a list parameter?

PavelMakarchuk:
> Please change the changelog to minor, this isn't a bug fix it adds a feature
> 
> Can you also check we can't do this with a parametric reform by making it a list parameter?

I believe we could but I don't think the web app currently supports parametric reforms which adjust list parameters. [issue](https://github.com/PolicyEngine/policyengine-app/issues/1887). 
I can make the change but this reform will not be available through the web app, only in python notebooks

MaxGhenis:
I mean keeping it as a switch but having the switch activate the parametric reform

policyengine_us/reforms/salt_phase_out/salt_phase_out_reform.py
MaxGhenis, line 1: Can't we remove this since it's no longer a structural reform?

MaxGhenis (changes requested):

MaxGhenis (approved):
I misunderstood - we need it as a separate reform for the phase-out

PR #5521, opened 2025-01-27
added 2025 values

daphnehanse11:
Updated Texas TANF income limits for 2025.


MaxGhenis (commented):
Why move away from the scale parameter? This enables us to capture the `5+` category without additional logic.

daphnehanse11:
@MaxGhenis not sure, I was just coping nj format because I wasn't which was more correct. the other format is in my first commit if you like that one better

MaxGhenis:
NJ caps instead then looks up:

https://github.com/PolicyEngine/policyengine-us/blob/c5737567248a6633c84f111bff101c3fde5306b1/policyengine_us/variables/gov/states/nj/njdhs/tanf/nj_tanf_maximum_benefit.py#L16

We'd have to adjust the TX approach to adopt that:

https://github.com/PolicyEngine/policyengine-us/blob/c5737567248a6633c84f111bff101c3fde5306b1/policyengine_us/variables/gov/states/tx/tanf/tx_tanf_income_limit.py#L18

I filed https://github.com/PolicyEngine/policyengine-us/issues/5524 to adopt a consistent standard here.

MaxGhenis (changes requested):
Can you try to find the effective date here? The source only says these are current values.

policyengine_us/variables/gov/states/tx/tanf/one_time_payment/one_time_tx_tanf_income_limit.py
```diff
@@ -1,10 +1,10 @@
 from policyengine_us.model_api import *
 
 
-class tx_tanf_income_limit(Variable):
+class one_time_tx_tanf_income_limit(Variable):
```
MaxGhenis, line 4: We assign state prefixes to all state variables
```suggestion
class tx_one_time_tanf_income_limit(Variable):
```

policyengine_us/parameters/gov/states/tx/tanf/one_time_tanf/one_time_monthly_income_limit.yaml
```diff
@@ -13,19 +13,24 @@ brackets:
       2023-01-01: 1
```
PavelMakarchuk, line 13: Is there a reference for the previous values? Want to make sure that the value did not update in 2024 as well

PavelMakarchuk (changes requested):

PR #5639, opened 2025-02-25
Hide program takeup parameters from web UI by setting economy: false

### LINKED ISSUE #5638: Hide takeup parameters from web UI by setting economy: false
Takeup parameters (like SNAP takeup, EITC takeup, etc.) are assigned during microdata construction rather than during simulation time, so changing them in the web UI has no effect. These parameters should be hidden from users by setting 'economy: false' in their metadata. This affects all takeup parameters including SNAP, EITC, and DC PTC.

MaxGhenis:
These parameters don't have any effect in the web UI since takeup is assigned during microdata construction, not simulation time. Setting economy: false prevents confusion by hiding them from users.

Fixes #5638


PavelMakarchuk (approved):
LGTM if the `household: false` metadata tag is irrelevant here

MaxGhenis:
Good call, I added household: false too.

PR #5646, opened 2025-02-25
Arizona 2024 state tax update 

### LINKED ISSUE #5616: Update 2024 Arizona Income Tax Parameters

juaristi22:
Fix #5616


juaristi22:
Closed so I wait on mn pr to avoid merge conflicts

policyengine_us/parameters/gov/states/az/tax/income/deductions/standard/amount.yaml
```diff
@@ -18,15 +20,19 @@ JOINT:
   2021-01-01: 25_100
   2022-01-01: 25_900
   2023-01-01: 27_700
+  2024-01-01: 29_200
```
DTrim99, line 23: Looks like this mirrors the federal standard deduction, so we should add uprating (chained CPI)

DTrim99 (changes requested):
Remove non-AZ files

policyengine_us/parameters/gov/states/az/tax/income/deductions/standard/amount.yaml
MaxGhenis: check the statutes. if it matches the federal standard deduction, remove this file and point to the federal in the formula.

MaxGhenis (commented):
looks like you might have mn stuff in here which could have arisen from checking out this branch from the other branch, instead of from master

PR #5649, opened 2025-02-28
Cap the FISC Act AGI floor at 0

### LINKED ISSUE #5648: Cap the FISC Act AGI floor at 0
The FISC phases-in with AGI - meaning it is capped at AGI (min_(AGI, base)) - however, given this implementation, we have some households with negative AGI which are receiving a negative credit amount

PavelMakarchuk:
Fixes #5648

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    fixed:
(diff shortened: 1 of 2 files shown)


MaxGhenis (approved):
could use positive_agi instead but merging

PR #5652, opened 2025-02-28
Fix LA federal tax deduction

### LINKED ISSUE #5651: Fix LA federal tax deduction
Currently we reduce the deduction by the PTC but it should be the PTC repayment amount

Image

Image

PavelMakarchuk:
Fixes #5651

### DIFF
--- a/policyengine_us/tests/policy/baseline/gov/states/la/tax/income/deductions/la_federal_tax_deduction.yaml
+++ b/policyengine_us/tests/policy/baseline/gov/states/la/tax/income/deductions/la_federal_tax_deduction.yaml
@@ -6,14 +6,14 @@
     form_4972_lumpsum_distributions: 600
     aca_ptc: 500
   output:
-    la_federal_tax_deduction: 100
+    la_federal_tax_deduction: 600
 
 - name: Capped at 0
   period: 2021
   input:
     state_code: LA
     income_tax_before_refundable_credits: 1_000
-    form_4972_lumpsum_distributions: 600
+    form_4972_lumpsum_distributions: 1_400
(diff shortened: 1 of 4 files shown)


policyengine_us/variables/gov/states/la/tax/income/deductions/la_federal_tax_deduction.py
```diff
@@ -18,10 +18,11 @@ def formula(tax_unit, period, parameters):
         tax_before_refundable_credits = tax_unit(
             "income_tax_before_refundable_credits", period
         )
+        # The deduction is also reduced by the ACA PTC repayment amount
```
MaxGhenis, line 21:
```suggestion
        # The deduction is also reduced by the ACA PTC repayment amount.
        # We do not model this as it depends on the advance PTC received throughout the year.
```

MaxGhenis (approved):
if my suggestion is right

policyengine_us/tests/policy/baseline/gov/states/la/tax/income/deductions/la_federal_tax_deduction.yaml
```diff
@@ -6,14 +6,14 @@
...
+    la_federal_tax_deduction: 600
 
 - name: Capped at 0
   period: 2021
   input:
     state_code: LA
     income_tax_before_refundable_credits: 1_000
-    form_4972_lumpsum_distributions: 600
+    form_4972_lumpsum_distributions: 1_400
     aca_ptc: 800
```
MaxGhenis, line 17: remove from all unit tests as no longer an input

PR #5699, opened 2025-03-12
Correct counties

### LINKED ISSUE #5698: Correct list of counties
There are various mistakes and typos in the county list

anth-volk:
Fixes #5698 

This makes corrections to the list of counties, including:
* Fixing typos
* Removing counties that were duplicated into other states
* Adding counties that were not present for some states
* Commenting out Puerto Rico municipalities
* Updating the list of counties for Connecticut to reflect 2024 legal and statistical changes


PavelMakarchuk:
We also need to make adjustments in the `medicaid_rating_area.yaml` file as well as in the `medicaid_rating_areas` and `aca_rating_areas` csv files 

However, this makes me think about how to approach changing county structures E.g. if in 2015 Fairfield county Connecticut officially existed and was then depreciated in 2024, we should be able to represent both for accuracy purposes - I don't think our current ENUM files allow for such an adjustment - a hackish way of doing this would be to create a new county ENUM file which is used post year x, not sure when it changed -- this does not seem sustainable and will require significant changes in the data repo -- would not recommend

Not sure if we will need to make changes to the us-data repository as well as we map county_fips to the CPS

anth-volk:
Okay, let's address this when we chat today. Another thought - why do we manually define these enums at all? Surely there's a package we could use to do county lookup and could just centralize around FIPS code (since FIPS never deprecates old counties for backward compatibility purposes) or around name, but find a package with backwards compatibility. We should cut manual maintenance wherever possible, in my opinion.

MaxGhenis:
Agree we should migrate to FIPS codes, though we could potentially separate that from this and the current ACA work.

anth-volk:
Closing in light of #5715.

PR #5735, opened 2025-03-18
Map county FIPS to `county` and prefer FIPS over ZIP code

### LINKED ISSUE #5734: Map county FIPS codes to county enum and use FIPS inputs to set `county`

anth-volk:
Fixes #5734 

This PR modifies the `county` variable to determine a situation's `county` value from FIPS code if all items in a given vector input possess one, otherwise `county` continues to apply its default ZIP code-based formula. This will allow us to create households with input county FIPS values without modifying how we do society-wide simulations.

Requires https://github.com/PolicyEngine/policyengine-us-data/pull/195 to be closed and merged before tests will pass.


policyengine_us/variables/household/demographic/geographic/county/county.py
```diff
@@ -14,21 +17,18 @@ class county(Variable):
     definition_period = YEAR
 
     def formula(household, period, parameters):
+        # First look if county FIPS is provided; if so, map to county name
+        county_fips: int | None = household("county_fips", period)
```
PavelMakarchuk, line 21: Is the typehint here accurate? Might have missed something, just checking

anth-volk: You're right, thanks for catching

anth-volk: In fact, this is extra incorrect, since it's vectorized, right?

PavelMakarchuk (approved):
LGTM

PavelMakarchuk:
Non-blocking - only thing I could suggest is to think about using `pd.merge` instead of `.loc` for the ability to debug through Nan values and future flexibility

PR #5761, opened 2025-03-20
Set default to true for branching to determine itemization choice

### LINKED ISSUE #5748: Set default to true for branching to determine itemization choice
Recently modeling shows the current default (a simplification for performance reasons) can significantly impact results, especially for reforming policies regarding AMT.

PavelMakarchuk:
Fixes #5748


PavelMakarchuk:
per conversation with @nikhilwoodruff 
Due to memory issues we need to move the test run to `larger-runner` instead of `ubuntu-latest`

MaxGhenis:
Depends on https://github.com/PolicyEngine/policyengine-us-data/issues/203

PavelMakarchuk:
Tests are taking 4 hours and still not done @MaxGhenis @nikhilwoodruff

PR #5784, opened 2025-03-25
Indiana 2024 state tax update

### LINKED ISSUE #5760: Update 2024 Indiana Income Tax Parameters

juaristi22:
Fix #5760 

Some changes that I am not too sure how to address (see 2024 Changes section in page 3):

- There is a new healthcare sharing deduction, I created a new file for it but I'm not sure if it requires any formula or additional parameters as the deduction applies to the full amount paid on the specific healthcare ministry.
- There seems to be a new NOL add-back and some other NOL modifications but I don't really understand how they are different from the already existing implementation or if there is anything to update really.
- I don't have access to IN-OCC forms, but I don't think we currently model credits being carried forward across years or if it is necessary to add new files for it?
- Out of all the new credits described, have I understood correctly that they are all specific to businesses and employers (and thus not relevant to our modeling) or should I implement any of them?


policyengine_us/parameters/gov/states/in/tax/income/agi_rate.yaml
```diff
@@ -18,3 +18,5 @@ metadata:
       href: http://iga.in.gov/legislative/laws/2023/ic/titles/006#6-3-2-1
     - title: 2023 IT-40 Income Tax Form
       href: https://forms.in.gov/Download.aspx?id=15790
+    - title: 2024 IT-40 Income Tax Form
+      href: https://www.in.gov/dor/tax-forms/individual/current/
```
PavelMakarchuk, line 22: This link will probably change next year to 2025 form - we should use:

https://forms.in.gov/Download.aspx?id=16379

Even if it is download link

PavelMakarchuk:
We can leave the healthcare sharing deduction without a formula, we don't have good rules for healthcare ministries in Indiana. 

The add-back is designed to capture some of the carry-forward operating losses - we can leave this out for now as we dont include debt discharges in our model

We exclude employer side and business credits as well as logic for credits being carried forward 

Summarized - the PR lgtm in it's current state - we currently don't represent most of the new programs introduced in Indiana in 2024

DTrim99 (approved):

policyengine_us/parameters/gov/states/in/tax/income/deductions/military_service/max.yaml
```diff
@@ -10,3 +11,5 @@ metadata:
       href: http://iga.in.gov/legislative/laws/2021/ic/titles/006#6-3-2-4
     - title: 2023 IT-40 Income Tax Form Instruction Booklet, pg. 17
       href: https://forms.in.gov/Download.aspx?id=15792
+    - title: 2024 IT-40 Income Tax Form Instruction Booklet, pg. 18
```
MaxGhenis, line 14: I see from this:
>For 2024, the deduction for active duty military compensation has been increased to 100% of compensation. See page 18 for more information.

But the statute still has the cap. Could you leave a comment above the statute reference noting that it hasn't yet been reflected in statute? If you could also find a bill that does this that'd be helpful.

juaristi22:
I got a test error related to hugging face, apparently, it is unable to download the cps dataset to run microsim tests. Any hints on how to solve this?

MaxGhenis:
Looks like the error resolved so merging!

PR #5807, opened 2025-04-01
Include Massachusetts EAEDC and TAFDC in the net income tree

### LINKED ISSUE #5806: Include Massachusetts EAEDC and TAFDC in the net income tree

PavelMakarchuk:
Fixes #5806


MaxGhenis (approved):

anth-volk:
My guess on the speed test failing here is that part of the test is to calculate `household_net_income`. I think this code is running out of memory, failing to calculate properly, or timing out due to inactivity. When running the speed test script locally, it's taking a very long time. Is it possible there's some cross-dependency issues for the two enabled programs?

PR #5883, opened 2025-04-18
Limit itemized deductions to taxable income

### LINKED ISSUE #5881: Limit itemized deductions to taxable income
This is actually the solution to the AMT-rising-with-SALT issue described in https://github.com/PolicyEngine/policyengine-us/issues/5880

Note that the IRC does not require reporting the full value of itemized deductions, so this is legal

We should make it a parameter that defaults to true to control this, in the simulation folder to reflect that it's effectively a behavioral response

PavelMakarchuk:
Fixes #5881


MaxGhenis (approved):
Fyi @nikhilwoodruff this is the kind of logic we should move out of the rules engine at some point

PR #5927, opened 2025-04-30
Update MA utility allowances

CalebPena:
https://www.snapscreener.com/guides/massachusetts#:~:text=Deduction-,Utility%20Allowances,-Households%20can%20only

I think the other states need to be updated as well.


policyengine_us/parameters/gov/usda/snap/income/deductions/utility/limited/main.yaml
```diff
@@ -92,6 +92,7 @@ MA:
   2010-01-01: 0
   2021-10-01: 421
   2023-10-01: 520
+  2024-10-01: 542
```
PavelMakarchuk, line 95: Could you add a reference showing where the values came from?

PavelMakarchuk (approved):

MaxGhenis (approved):

PR #6047, opened 2025-05-26
Refactor input income variables

baogorek:
## Summary
- move new income-related input variables into their own modules
- trim `variables/input/income.py`

## Testing
- `make test` *(fails: ModuleNotFoundError: No module named 'policyengine_core')*


PavelMakarchuk:
Was this supposed to be merged - I dont think I see a changelog here

PR #6067, opened 2025-05-29
QBID formula Logic and Tests

### LINKED ISSUE #6011: Ensure Reconciliated QBID logic is correct
This is regarding [https://github.com/PolicyEngine/policyengine-us/blob/master/policyengine_us/reforms/reconciliation/reconciled_qbid.py](reconciled_qbid.py)

A CPA wrote [this essay](https://www.currentfederaltaxdevelopments.com/blog/2025/5/14/key-modifications-to-the-section-199a-qualified-business-income-deduction-in-the-proposed-one-big-beautiful-bill) which talks about a "two-step" process involving the new phase-in. He claims that SSTBs have a zero value in step 1 (so an immediate cliff), and it also uses the word "greater" when talking about how to aggregate Steps 1 and 2 (though it seems he's mixing income and deductions in that sentence.) He also mistypes the wage and property percentages in calculating "the sum" (as the bill puts it), so maybe it was a rushed job. I left comments seeking clarification, because I do believe this CPA is competent and I would like to hear his perspective.

I went back to the bill itself, and I really struggled to understand this language:

Image

I believe this "sum" is the original `full_cap` quantity in the code, and I think the words EXCESS and OVER make it clear that this is taking the difference between the two, as the code does. What is less clear is whether SSTBs have the immediate cliff when they hit their threshold, as the CPA's essay suggests. The bill text only mentions SSTBs to acknowlege that they can be considered a qualified trade when their income is under the threshold.

What this also means is that this deduction eventually zeros out for every qualified business that makes sufficient income. Perhaps that will reduce the revenue cost of reconcilliation somewhat, as the TCJA QBID never goes to zero as long as the business has W2 wages, to my understanding.

The other thing that I think is missing, which could be a rounding error, is that 23% of REIT dividends, Publicly Traded Partnership (PTP) income, and, under reconcilliation, Business Development Company (BDC) dividends should be included in the deduction. We don't have direct data for any of those, to my knowledge, so the question is whether we want to create the variables and simulate.

So,

1) Do we have the correct phase-in logic?
2) Do we need to add in the "is_sstb" variable?
3) Should we add in these dividends and income from the special sources?

baogorek:
fixes #6011


policyengine_us/variables/gov/irs/income/taxable_income/deductions/qualified_business_income_deduction/qbid_amount.py
PavelMakarchuk, line 1: could we break this formula out for clarity - the final output here should be the person level qbid amount but maybe we can have components such as the phase-in separately for clarity if not too much code duplication

policyengine_us/variables/gov/irs/income/taxable_income/deductions/qualified_business_income_deduction/qbid_amount.py
```diff
@@ -15,42 +15,81 @@ class qbid_amount(Variable):
...
-        # compute phase-out limited QBID amount
-        reduction_rate = min_(  # Worksheet 12-A, line 24; Schedule A, line 9
-            1, (max_(0, taxinc_less_qbid - po_start)) / po_length
+
+        threshold = p.phase_out.start[filing_status]
+        phase_in_range = p.phase_out.length[filing_status]
+
+        # 2. 20 % of QBI ---------------------------------------------------------
+        qbid_max = p.max.rate * qbi
+        # 20% of qualified business income (Worksheet 12‑A)
```
PavelMakarchuk, line 44: Can we remove specific parameter values from the comments? If e.g. the 20% rate changes these would be outdated and confusing if reviewing the logic in a couple of weeks / months / years

policyengine_us/variables/gov/irs/income/taxable_income/deductions/qualified_business_income_deduction/qualified_business_income.py
```diff
@@ -13,6 +13,12 @@ class qualified_business_income(Variable):
 
     def formula(person, period, parameters):
         p = parameters(period).gov.irs.deductions.qbi
-        gross_qbi = add(person, period, p.income_definition)
+        gross_qbi = sum(
```
MaxGhenis: is `sum()` vectorized? if not, you could do this instead:
```suggestion
        gross_qbi = 0
        for var in p.income_definition:
            gross_qbi += person(var, period) * person(var + "_would_be_qualified", period)
```

PavelMakarchuk: done

MaxGhenis (approved):

PR #6222, opened 2025-07-08
Update income tax parameters following passage of OBBB

### LINKED ISSUE #6219: Update income tax parameters following passage of OBBB

DTrim99:
Fixes #6219


policyengine_us/parameters/gov/irs/credits/ctc/amount/base.yaml
```diff
@@ -8,8 +8,13 @@ brackets:
       values:
         2013-01-01: 1_000
         2018-01-01: 2_000
+        2025-01-01: 2_200
+        2027-01-01: 2_300
```
MaxGhenis, line 12: can we rely on uprating instead?

DTrim99: A lot of the provisions use a base year before 2025 (in this case, 2024). If we uprate as usual, we may be off. Is there a way to set the base year in the uprating?

MaxGhenis: I believe @anth-volk implemented this so correct me if I'm wrong, but I think you'd do it this way:
```yaml
example_benefit:
  description: Example benefit amount
  values:
    "2020-04-01": 1000
  metadata:
    uprating:
      parameter: cpi
      at_defined_interval:
        enactment: "2021-04-01"  # First uprated value applies on this date
        start: "2020-04-01"      # Base year for uprating (for the ratio)
        end: "2021-04-01"        # End of uprating interval (for the ratio)
        interval: year           # Uprate every year
```
See this Copilot chat in -core:
image

PavelMakarchuk:
Are we trying to merge OBBBA related updates asap?

policyengine_us/parameters/gov/irs/credits/ctc/amount/base.yaml
anth-volk: Does this mean that 2026 uses 2024's values, 2027 uses 2025's, etc.? If not, could you go more in depth @DTrim99?

DTrim99: So the CTC's new value starts in 2025 and then is uprated using 2024 as the base year. So the inflation is always tied to 2024, but does not begin uprating until 2026.

anth-volk: Chatting with David now on this

PavelMakarchuk: trying this with David now, but I will file a core issue to rename some of the metadata:
`start` should be `base` and `enactment` should be `start` IMO

anth-volk: That's a good idea. The original purpose was for some UK programs, so I'd just want to confirm that we properly apply those changes there, too.

DTrim99: We are going to stick with inputting each individual parameter value based on our web app reform, and when CBO releases the projections of each parameter, I will update the model.

DTrim99:
Ready for review, but leaving as a draft since we don't want to merge until the structural PR is ready.

PavelMakarchuk:
LGTM, lets not merged before #6294 is approved

MaxGhenis (approved):

PR #6236, opened 2025-07-11
Make itemisation optimise federal tax only

nikhilwoodruff:
Fix https://github.com/PolicyEngine/policyengine-us/issues/6237


PavelMakarchuk (approved):
Not sure if there is a way to tests if there are any weird state-level interactions but we can monitor sim runs. Lgtm

MaxGhenis (approved):

PR #6238, opened 2025-07-11
Separate DACA and TPS as Immigration statuses

### LINKED ISSUE #6230: Separate DACA and TPS as Immigration statuses
Currently we bundle the two, but they have different eligibility rules.

daphnehanse11:
Fixes #6230


PavelMakarchuk:
Are we consolidating the three in the future?

PR #6243, opened 2025-07-11
Downgrade from larger-runner to ubuntu-latest

### LINKED ISSUE #6242: Downgrade from larger-runner to ubuntu-latest
Let's see if the branching change allows for this

### LINKED ISSUE #6244: Skip reform tests in CI
They're causing it to not complete on ubuntu-latest, so let's only run them locally for now

MaxGhenis:
Fixes #6242
Fixes #6244


MaxGhenis:
Need to remove the structural one from the workflow for now (contrib != reform)

PR #6264, opened 2025-07-18
2018 CHIP pregnant income limit

### LINKED ISSUE #6254: 2018 CHIP pregnant income limit
source: https://data.medicaid.gov/dataset/d7e4cccb-1c56-5b5d-acce-5e7744c6d3b4

states where values differ: 
Kentucky - N/A
West Virginia - N/A

benccann:
Fixes #6254


policyengine_us/parameters/gov/hhs/chip/pregnant/income_limit.yaml
```diff
@@ -7,6 +7,8 @@ metadata:
   reference:
     - title: Medicaid, Children's Health Insurance Program, & Basic Health Program Eligibility Levels as of December 2023 | CMS
       href: https://www.medicaid.gov/medicaid/national-medicaid-chip-program-information/medicaid-childrens-health-insurance-program-basic-health-program-eligibility-levels
+    - title: Medicaid and CHIP Eligibility Levels
+      href: https://data.medicaid.gov/dataset/d7e4cccb-1c56-5b5d-acce-5e7744c6d3b4
 AK:
   2021-01-01: -.inf
```
PavelMakarchuk, line 13: we will need a value for 2018 for each entry. So if it doesn't changes as in this case in AK we need to change 
`2021-01-01: -.inf` to `2018-01-01: -.inf` so we can run a Microsim. 

We will need to do this for each entry.

PR #6267, opened 2025-07-21
Update FY2025 HHS SMI parameter

### LINKED ISSUE #6262: FY2025 State Median Income parameter update

hua7450:
Fixes #6262


policyengine_us/parameters/gov/hhs/smi/amount.yaml
```diff
@@ -1,222 +1,277 @@
...
-    - title: State Median Income (SMI) by household size FFY 2022
-      href: https://www.acf.hhs.gov/sites/default/files/documents/ocs/COMM_LIHEAP_IM03%20Attachment1%20SMITable_FY2022.pdf
-    - title: State Median Income (SMI) by Household Size for Optional Use in FFY 2022 and Mandatory Use in LIHEAP for FFY 2023
-      href: https://www.acf.hhs.gov/sites/default/files/documents/ocs/COMM_LIHEAP_Att1SMITable_FY2023.pdf
-    - title: State Median Income (SMI) by Household Size for Mandatory Use in LIHWAP for FY 2024
-      href: https://www.acf.hhs.gov/sites/default/files/documents/ocs/COMM_LIHWAP_Att1SMITable_FY2023.pdf
   period: year
-  uprating: calibration.gov.cbo.income_by_source.adjusted_gross_income
+  label: HHS SMI 4-person household
+  #uprating: calibration.gov.cbo.income_by_source.adjusted_gross_income
```
hua7450: When testing for 2025, the uprating line is causing the output becomes higher than expected. 

Image


Two ways to solve this. 
1. Remove the uprating. 
2. Or, add the values for FY2026 with the same values as FY2025. 
Image

MaxGhenis: isn't the test wrong?

hua7450: Current test is passing because I commented the uprating line.
If I don't remove the uprating line, the test will be wrong. It would return the result as 128,908.34, where it is supposed to be 121,926.

MaxGhenis: If the test assumes no uprating it is wrong and we should fix it. I'd just set it to a known month or year to avoid an uprating dependency though

MaxGhenis (approved):

PavelMakarchuk (approved):

PR #6306, opened 2025-07-28
Add variables needed for calibration

### LINKED ISSUE #6292: Add variables that support calibration
These include:
- [x] 'ucgid'
- [ ] 'person_count'
- [x] 'household_count'
- [x] 'tax_unit_count'

juaristi22:
Fix #6292

I added a `ucgid` variable that supports both specifying a specific ucgid code to a household, and retrieving the hierarchical relationship between codes: a houshold with a district code also belongs to the corresponding state and national geographies and will have those codes, while a household with the national us code will not have state or district information (default behavior).

I also added `household_count`, `spm_unit_count`, and `tax_unit_count`. 

A `person` variable that counts the number of people represented already exists, should we rename to `person_count`?


juaristi22:
My understanding is that tax_unit_count would represent the number of total tax units in the simulation while converting with map_result the tax_unit_count variable from the `tax_unit`entity to the `household` entity would be the number of tax units in a household. At least thats what I got from playing around with the `people` variable and different entities but my conclusions may be wrong. Let me know if it should be, or is, otherwise.

juaristi22:
I think the purpose may be a bit different in this case. The rationale behind adding _count variables comes from the way we are defining calibration targets. Every variable we calibrate for has a "count" value that represents the number of households  within a given strata that contribute to the target total (eg. `employment_income` targets broken down by agi bracket also have a "count" of how many people had an `employment_income`  within an agi bracket different from 0). 
We realized we had assumed that every target defined would have a -us variable that it could map to, but until now , "counting" variables didn't really exist as there wasn't a need to count the number of households/tax units/etc that meet a certain condition, which is why we are adding variables that could store and be used to operate this kind of information. 



> * Do we need `household_count`?

It is likely that we will need a _count for every entity type as the variables that define strata constraints are often defined at different entity levels. We would want to make sure we are counting at the entity level that defines the stratum constraint and then convert to the count of a different entity. 


> * Should we create household level variables and derive the number of units via formula based on unique identifiers such as IDs or # of heads in the unit?

Maybe. It's probably worth ensuring that the map_result function is flexible enough for all the conversions we need and adding functionality or specific formulas if we realize some conversions aren't meeting our needs.




> * Should we add a count for `marital_unit`?

Also maybe. There is a chance we will eventually target the counts for marital units or this functionality could become useful for some other anlysis, happy to look into it if its relevant.

PR #6322, opened 2025-07-29
Fix ucgid_str 

juaristi22:
Fix #

### DIFF
--- a/policyengine_us/tests/utilities/test_ucgid_hierarchical.py
+++ b/policyengine_us/tests/utilities/test_ucgid_hierarchical.py
@@ -56,3 +56,17 @@ class TestUCGIDHierarchical:
         # Test hierarchy matching in simulation context
         is_in_us = ucgid_enum.matches_hierarchy("0100000US")
         assert is_in_us == True
+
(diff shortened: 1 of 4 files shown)


policyengine_us/tests/utilities/test_ucgid_hierarchical.py
```diff
@@ -50,9 +48,74 @@ def test_ucgid_simulation_usage(self):
...
+        assert len(hierarchical_codes) == 2
+        assert (
+            hierarchical_codes[1] == "0100000US"
+        )  # Second code should always be US
+        assert hierarchical_codes[0].startswith(
+            "0400000US"
+        )  # First should be state-level
 
         # Test hierarchy matching in simulation context
         is_in_us = ucgid_enum.matches_hierarchy("0100000US")
```
MaxGhenis, line 62:
```suggestion
        assert ucgid_enum.matches_hierarchy("0100000US")
```

MaxGhenis (approved):
going to merge without this but just for the future. in general checking `condition` is more conventional than `condition is True`

PR #6331, opened 2025-07-29
Fix typo in Colorado state supplement directory name

MaxGhenis:
Fixed typo: 'state_suplement' → 'state_supplement' (was missing second 'p')

This is a simple directory rename with no functional changes.


docs/_data/program_menu.yaml
PavelMakarchuk, line 1: The manu changes should not be here right?

MaxGhenis:
Superseded by #6886 which contains only the typo fix without the mixed changes from the program menu PR.

Issue #6362, opened 2025-08-05
Remove confusing comment in `.github/update_api.py`


anth-volk:
Very minor issue, but as part of #6342, I mistakenly left a testing comment in the pushed code:
```
    with open("pyproject.toml", "rb") as f:
        data = tomllib.load(f)
        version = data["project"]["version"]  # Would give us "1.358.0"
```

We should remove "Would give us..."

PR #6371, opened 2025-08-06
Removing confusing comment from update_api.py as requested in PR #6362

Vimal-Shady:


PavelMakarchuk:
Could you add changelog to the PR

Issue #4062, opened 2024-03-03
Add blind SGA


MaxGhenis:
Low priority as we don't currently use this, since it's only for SSDI not SSI

https://www.ssa.gov/oact/cola/sga.html

PR #6378, opened 2025-08-07
Updated SSI SGA param values

nicholasburka:
Updates the SSI Substantial Gainful Activity non-blind parameter with the 2025 value, and adds a blind parameter with a 2025 value.


nicholasburka:
Don't think I'm successfully reverting the changelog to its current status correctly?

policyengine_us/parameters/gov/ssa/sga/blind.yaml
PavelMakarchuk, line 1: Do we want to add the SGA logic in this PR as well? Currently this parameter won't affect the computation treee

MaxGhenis: No see https://github.com/PolicyEngine/policyengine-us/issues/4062

PR #6430, opened 2025-08-22
Negative state AGI subtractions act as additions in the adds / subtracts structure with negative income

### LINKED ISSUE #6428: Negative state AGI subtractions act as additions in the adds / subtracts structure with negative income
In the following test case:
```
- name: taxsim_record_848_2024.yaml
  absolute_error_margin: 2
  period: 2024
  input:
    people:
      person1:
        age: 72
        employment_income: 0.0
        short_term_capital_gains: -3000000
        is_tax_unit_head: true
    tax_units:
      tax_unit:
        members: [person1]
        premium_tax_credit: 0
        local_income_tax: 0
        state_sales_tax: 0
    spm_units:
      spm_unit:
        members: [person1]
        snap: 0
        tanf: 0
    households:
      household:
        members: [person1]
        state_fips: 13
  output:
    state_income_tax: 
```

Federal AGI is -$3,000 while state AGI returns $2,997,000. 

We should make sure subtractions do not increase AGI

PavelMakarchuk:
Fixes #6428

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.377.0] - 2025-08-25 18:03:27
+
+### Fixed
+
+- Fixed IL AABD integration tests.
+
+## [1.376.4] - 2025-08-25 13:45:00
+
+### Fixed
+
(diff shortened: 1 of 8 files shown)


MaxGhenis (approved):
LGTM! This correctly fixes the issue where negative state AGI subtractions were acting as additions for taxpayers with negative income.

The solution of adding `max_(0, total_subtractions)` guards to all state subtraction variables is the right approach. This ensures subtractions can never go negative and inadvertently increase AGI.

The fix has been consistently applied across all 28 affected states, and the test case clearly demonstrates the problem and solution.

PR #6441, opened 2025-08-26
Memory Fix Option A

### LINKED ISSUE #6445: Tests fail with OOM (Killed: 9) and timeout (Killed: 14) errors in GitHub Actions

hua7450:
Fixes #6445 

## Summary
This PR fixes memory-related test failures in GitHub Actions by implementing an optimized batched test runner with proper memory management.

## Problems Solved
1. **Memory Issues**: Tests consuming excessive memory causing "Killed: 9" errors
2. **Memory Inefficiency**: Tests running out of memory in GitHub Actions
3. **Code Duplication**: Multiple batch runners with redundant code

### Key Features
- **Smart batching by folder**: Groups tests by state/folder to share data structures efficiently
- **Memory cleanup**: `gc.collect()` between batches prevents OOM
- **Configurable batching**: `--batches` parameter controls splitting
- **Clean code**: Consolidated into single utility

### Batching Strategy
- **Baseline**: 2 batches
  - Batch 1: All state tests (~5,638 tests) - share data structures efficiently
  - Batch 2: Federal/other tests (~2,034 tests)
  - Memory cleared between batches
- **Contrib**: Tests grouped by folder for optimal memory usage

## Changes
- ✅ Updated `test_batched.py` with folder-based batching for contrib tests
- ✅ Optimized memory usage by processing related tests together
- ✅ Updated GitHub Actions to run comprehensive tests with better visibility
- ✅ Split test runs into separate GitHub Actions steps for clearer output

## Testing
```bash
# Run all tests locally (same as GitHub Actions)
make test-yaml-no-structural  # Baseline (2 batches) + reform
make test-yaml-structural     # Contrib (folder-based batching)
make test-other              # Unit tests

# Run specific batch only
python policyengine_us/tests/test_batched.py policyengine_us/tests/policy/baseline --batches 2
```

## Impact
- ✅ Fixes memory exhaustion errors
- ✅ Improved test organization and visibility in GitHub Actions
- ✅ More efficient memory usage through smart batching
- ✅ Reliable GitHub Actions runs
- ✅ Better test output clarity with separated test steps


.github/workflows/pr.yaml
hua7450, line 1: If it passed all checks in Github Action, we can undo the changes in this `pr.yaml` before merge this PR into master branch.

policyengine_us/tests/test_contrib_batched.py
```diff
@@ -0,0 +1,176 @@
...
+
+import os
+import sys
+import subprocess
+import gc
+from pathlib import Path
+from datetime import datetime
+from typing import List, Dict, Any
+
+
```
anth-volk: __question:__ Couldn't this also be run by making `test_baseline_batched` into a utility that takes a folder name and a batch size as an arg, then for the contrib folder, make the batch size as large as possible?

anth-volk (commented):
Left a single comment here. I think there are a few ways to review this. My view of this code depends on your needs and intentions with it:

1. __We need this code this week and this is meant to be temporary:__ I think this will probably fit your needs as it stands. In fact, you could even do the same thing by having a ton of jobs in GitHub Actions, but that's more brittle.
2. __We need this code in the next two weeks and it's temporary:__ You might consider making this script a bit more versatile. `test_baseline_batched` and `test_contrib_batched` are nearly the same thing, just `contrib` tells the script to have an infinite batch size.
3. __We need this code in general and/or it's more permanent:__ You might consider modifications to `-core` instead. The durable fix is likely better memory management there in the test builder script.

.gitignore
```diff
@@ -48,6 +48,11 @@ docs/book/_build
 **/*.exe
 
 # Personal utility scripts
+
+# Test output files
+*_test_results*.json
+baseline_test_results_batched.json
+contrib_test_results_batched.json
```
anth-volk: __nit:__ I know you said no nits, but I don't think you're generating these files anymore

policyengine_us/tests/test_baseline_batched.py
```diff
@@ -0,0 +1,414 @@
...
+        cmd = (
+            [
+                python_exe,
+                "-m",
+                "policyengine_core.scripts.policyengine_command",
+                "test",
+            ]
+            + test_dirs_list
+            + ["-c", "policyengine_us"]
+        )
```
anth-volk: __question:__ I don't think this code will measure coverage via `coverage`; is that alright?

policyengine_us/tests/test_baseline_batched.py
```diff
@@ -0,0 +1,414 @@
...
+        summary_match = re.search(r"(\d+) passed", output)
+        if summary_match:
+            passed_count = int(summary_match.group(1))
+        summary_match = re.search(r"(\d+) failed", output)
+        if summary_match:
+            failed_count = int(summary_match.group(1))
+
+        print(
+            f"    Completed in {elapsed:.1f}s: {passed_count} passed, {failed_count} failed"
+        )
```
anth-volk: __question:__ Doesn't this block of code just override the output parsing you do above it?

policyengine_us/tests/test_baseline_batched.py
```diff
@@ -0,0 +1,414 @@
...
+        batches.extend(
+            [
+                f"{baseline}/household",  # ~81 tests
+                f"{baseline}/contrib",  # ~7 tests
+                f"{baseline}/calcfunctions",  # ~1 test
+                f"{baseline}/income",  # ~1 test
+            ]
+        )
+
+        return batches
```
anth-volk: __suggestion:__ Kind of seems like it could just be a defined array, since you hard-code all the dirs

policyengine_us/tests/test_baseline_batched.py
```diff
@@ -0,0 +1,414 @@
...
+
+        dir_p = Path(dir_path)
+
+        if not dir_p.exists():
+            continue
+
+        # Special handling for parent directories that should only test direct files
+        if dir_path.endswith("/gov") or dir_path.endswith("/gov/states"):
+            # Only get files directly in this directory, not subdirectories
+            test_files = set(str(f) for f in dir_p.glob("*.yaml"))
```
anth-volk: __question:__ Can you walk me through why? I feel like I'm missing some context on `-us`.

hua7450: The old code would sometimes run /gov as a batch, but /gov contains /gov/states as a subdirectory. This would cause double testing - states tested once in their own batch, then again when /gov runs. So it tried to say "if running /gov, only run yaml files directly in /gov, not in subdirectories".
But we don't need to worry about this anymore, I delete them in the latest version.

policyengine_us/tests/test_batched.py
```diff
@@ -0,0 +1,365 @@
...
+        + test_paths
+        + ["-c", "policyengine_us"]
+    )
+
+    print(f"    Running {batch_name}...")
+    print(f"    Paths: {len(test_paths)} items")
+    print()
+
+    # Use Popen for more control over process lifecycle
+    process = subprocess.Popen(
```
nikhilwoodruff, line 165: I think there is a way to run pytest in Python itself but this is fine

nikhilwoodruff (approved):
Yeah OK. We do need to fix this so let's get this in

PR #6477, opened 2025-08-29
Remove Coverage Check

### LINKED ISSUE #6478: Remove Coverage Check in PR  and Push Action
In order to fix the memory issue in Actions, we optimized batched test runner in PR #6441. However, with the batched scripts, the coverage check is not working properly. All the attempts have been failed or made the testing process extremely slow. Hence, remove the coverage check might be the best option here.

hua7450:
Fixes #6478


hua7450:
Change of plan, we do partial coverage check now.

PR #6489, opened 2025-09-02
add maximum payment variable for NC SCCA

### LINKED ISSUE #6456: Add intermediate variable for CCAP maximum benefit amount
States need to adjust:
CO
NC
MA

hua7450:
Fixes #6456

### DIFF
--- a/policyengine_us/variables/gov/states/nc/ncdhhs/scca/nc_scca_maximum_payment.py
+++ b/policyengine_us/variables/gov/states/nc/ncdhhs/scca/nc_scca_maximum_payment.py
@@ -1,0 +1,22 @@
+from policyengine_us.model_api import *
+
+
+class nc_scca_maximum_payment(Variable):
+    value_type = float
+    entity = SPMUnit
+    label = "North Carolina Subsidized Child Care Assistance Program maximum payment"
+    unit = USD
+    definition_period = MONTH
+    defined_for = "nc_scca_entry_eligible"
+
+    def formula(spm_unit, period, parameters):
+        person = spm_unit.members
+        age_eligible = person("nc_scca_child_age_eligible", period)
+
+        # Collect the market rates only for eligible children
+        total_market_rate = spm_unit.sum(
+            person("nc_scca_market_rate", period) * age_eligible
+        )
+
+        parent_fee = spm_unit("nc_scca_parent_fee", period)
+        return max_(total_market_rate - parent_fee, 0)
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: minor
+  changes:
+    added:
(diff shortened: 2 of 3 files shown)


policyengine_us/variables/gov/states/nc/ncdhhs/scca/nc_scca_maximum_payment.py
MaxGhenis, line 1: use this in the final calc

hua7450: I can't. In the final calc, we use `min(max(expenses - copay, 0), max_reimbursement)`. Here it's `max_reimbursement-copay` It will be double minus copay if I add this in the final calc.

MaxGhenis: i dont follow. im looking at `nc_scca`, couldn't that just be `min_(spm_unit_pre_subsidy_childcare_expenses, nc_scca_maximum_payment)`?

hua7450: If the expense is $100, reimbursement is $50, copay is $10, In reality, the benefit is `min_(100 - 10, 50) = 50`,
but with this logic `min_(spm_unit_pre_subsidy_childcare_expenses, nc_scca_maximum_payment)`, it would be `min_(100, 50-10) = 40`

hua7450:
[Rules
](https://www.publichealthlawcenter.org/sites/default/files/NC%20Chapter%2010%20Subsidized%20Child%20Care%20Rules.pdf#page=2)Screenshot 2025-09-03 at 4 37 50 PM
Last commit based on this rule.

MaxGhenis:
Validated the NC SCCA approach and it looks correct.

- Order of operations: `nc_scca` now computes `min(max(expenses − parent_fee, 0), maximum_payment)`, with `maximum_payment = max(total_market_rate − parent_fee, 0)`. Algebraically this is `max(min(expenses, total_market_rate) − parent_fee, 0)`, i.e., apply the market-rate cap first, then subtract the parental fee. This matches NC Chapter 8 guidance on parental fees being applied to the allowable cost.
- Scope: `nc_scca_market_rate` is now defined for `nc_scca_child_age_eligible`, so only eligible children’s market rates are summed.
- Tests: Updated integration expectations reflect the corrected ordering (e.g., disabled school-age case → 108; two-young-children case → 1,392; ages 0 & 3 → 1,280).

Minor nit
- `nc_scca_market_rate.reference` concatenates two URLs in a single string; consider splitting into two entries for clarity.

Overall LGTM — the split into `nc_scca_maximum_payment` clarifies intent and prevents double counting of the parental fee.

PR #6517, opened 2025-09-09
Fix Social Security taxation formula parameter issue

### LINKED ISSUE #6516: Social Security taxation formula incorrectly uses base rate parameter for combined income calculation
## Issue Description
The Social Security taxation formula has a bug where the `rate.base` parameter is incorrectly used for two different purposes:
1. To determine what fraction of Social Security benefits to add to MAGI for the combined income test
2. As the actual taxation rate for the first tier of benefits

This creates incorrect behavior when modifying parameters for policy analysis.

## Current Behavior
In `tax_unit_combined_income_for_social_security_taxability.py`, the formula uses:
```python
ss_fraction = p.rate.base * gross_ss
return tax_unit("taxable_ss_magi", period) + ss_fraction
```

This means when `rate.base` is changed (e.g., to 0.85 for policy modeling), it incorrectly changes the combined income calculation, which should always add 50% of Social Security benefits per the tax law.

## Impact
When setting parameters to test full taxation of Social Security benefits:
- Base thresholds: 0
- Base rate: 0.85
- Additional rate: 0.85

Expected result: 85% of all benefits should be taxable
Actual result: Only 81.8% of benefits are taxable

This happens because low-MAGI households get an effective rate of 0.85² = 72.25% instead of 85% due to the parameter being used twice in the calculation chain.

## Proposed Solution
Create a new parameter `combined_income_ss_fraction` (or similar) that:
- Defaults to 0.5 (50%) to match current law
- Is used specifically for the combined income calculation
- Separates this from the `rate.base` parameter which should only control the taxation rate

This would involve:
1. Adding a new parameter file under `parameters/gov/irs/social_security/taxability/`
2. Updating `tax_unit_combined_income_for_social_security_taxability.py` to use the new parameter
3. Ensuring backward compatibility with existing calculations

## References
- IRC Section 86: https://www.law.cornell.edu/uscode/text/26/86
- The law specifies that combined income includes "one-half of the social security benefits" regardless of the taxation rates

DTrim99:
## Summary
- Fixes bug where the `rate.base` parameter was incorrectly used for both the combined income calculation and the taxation rate
- Introduces new parameter `combined_income_ss_fraction` that is fixed at 0.5 per IRC Section 86(b)(1)
- Ensures that policy analysis modifications to taxation rates don't affect the combined income calculation

## Problem
The Social Security taxation formula had a bug where the `rate.base` parameter was used for two different purposes:
1. To determine what fraction of Social Security benefits to add to MAGI for the combined income test
2. As the actual taxation rate for the first tier of benefits

This caused incorrect results when modifying parameters for policy analysis. For example, when setting base rate to 0.85 and thresholds to 0, only 81.8% of benefits were taxed instead of the expected 85%.

## Solution
Created a new parameter `combined_income_ss_fraction` that:
- Defaults to 0.5 (50%) to match current law
- Is used specifically for the combined income calculation
- Separates this from the `rate.base` parameter which now only controls the taxation rate

## Changes
- Added new parameter file `combined_income_ss_fraction.yaml`
- Updated `tax_unit_combined_income_for_social_security_taxability.py` to use the new parameter
- Updated comments in related files to clarify the separation
- Added test to verify the fix works correctly

## Testing
- Added test case to verify combined income calculation uses the fixed 0.5 fraction
- All existing tests continue to pass

Fixes #6516

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/policyengine_us/parameters/gov/irs/social_security/taxability/combined_income_ss_fraction.yaml
+++ b/policyengine_us/parameters/gov/irs/social_security/taxability/combined_income_ss_fraction.yaml
@@ -1,4 +1,4 @@
-description: Fraction of Social Security benefits included in combined income for determining taxability
+description: The United States includes this fraction of Social Security benefits for determining taxability.
 values:
   1984-01-01: 0.5
 
--- a/policyengine_us/parameters/gov/irs/social_security/taxability/combined_income_ss_fraction.yaml
+++ b/policyengine_us/parameters/gov/irs/social_security/taxability/combined_income_ss_fraction.yaml
@@ -1,0 +1,14 @@
+description: Fraction of Social Security benefits included in combined income for determining taxability
+values:
+  1984-01-01: 0.5
+
+metadata:
+  unit: /1
+  period: year
+  label: Social Security fraction for combined income calculation
+  reference:
+  - title: 26 U.S. Code § 86(b)(1)
+    href: https://www.law.cornell.edu/uscode/text/26/86#b_1
+  - title: IRS Publication 915 - Social Security and Equivalent Railroad Retirement Benefits
+    href: https://www.irs.gov/publications/p915
+ 
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    fixed:
+    - Social Security taxation formula now uses a separate parameter for combined income calculation instead of incorrectly reusing the base taxation rate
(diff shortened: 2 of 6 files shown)


policyengine_us/parameters/gov/irs/social_security/taxability/combined_income_ss_fraction.yaml
```diff
@@ -0,0 +1,14 @@
+description: Fraction of Social Security benefits included in combined income for determining taxability
```
MaxGhenis:
```suggestion
description: The United States includes this fraction of Social Security benefits for determining taxability.
```

MaxGhenis (approved):

PR #6538, opened 2025-09-15
Avoid counting the snap self-employment income deduction twice

### LINKED ISSUE #6537: Avoid counting the snap self-employment income deduction twice
We apply it once when calculating earned income for SNAP and then once again when calculating the net income since it is part of the deductions

PavelMakarchuk:
Fixes #6537


PavelMakarchuk:
Taking a different approach based on [this](https://www.snapscreener.com/guides/ohio) documentation

PR #6548, opened 2025-09-16
Branching for itemization results in higher state income tax liability

### LINKED ISSUE #6542: Branching for itemization results in higher state income tax liability
When the tax liability is equal, we default to the standard deduction. However in certain cases such as the one below, the tax liability is 0 in both cases, resulting in a higher state tax liability in Nebraska at it is tied to federal itemization.

Example test:

```
- name: taxsim_record_3022_2024.yaml
  absolute_error_margin: 2
  period: 2024
  input:
    people:
      person1:
        age: 72
        deductible_mortgage_interest: 50000
        taxable_interest_income: 34967.594
        is_tax_unit_head: true
      person2:
        age: 71
        taxable_interest_income: 34967.594
        is_tax_unit_spouse: true
      person3:
        age: 7
        is_tax_unit_dependent: true
        is_tax_unit_head: false
        is_tax_unit_spouse: false
      person4:
        age: 12
        is_tax_unit_dependent: true
        is_tax_unit_head: false
        is_tax_unit_spouse: false
      person5:
        age: 20
        is_tax_unit_dependent: true
        is_tax_unit_head: false
        is_tax_unit_spouse: false
    tax_units:
      tax_unit:
        members: [person1, person2, person3, person4, person5]
    spm_units:
      spm_unit:
        members: [person1, person2, person3, person4, person5]
    households:
      household:
        members: [person1, person2, person3, person4, person5]
        state_fips: 31
  output:
    tax_liability_if_itemizing: 0
    tax_liability_if_not_itemizing: 0
    income_tax: 0
    state_income_tax: 898.58
```

PavelMakarchuk:
Fixes #6542


policyengine_us/variables/gov/irs/income/taxable_income/deductions/tax_unit_itemizes.py
```diff
@@ -18,11 +22,11 @@ def formula(tax_unit, period, parameters):
             tax_liability_if_not_itemizing = tax_unit(
                 "tax_liability_if_not_itemizing", period
             )
-            return tax_liability_if_itemizing < tax_liability_if_not_itemizing
+            return where(
+                tax_liability_if_itemizing == tax_liability_if_not_itemizing,
+                itemized_deductions > standard_deduction,
```
MaxGhenis: Don't want to compare state versions?

PavelMakarchuk: Thought we could avoid state branching at first but found edge cases where it is not optimal to itemize

MaxGhenis: I meant state itemized deductions vs state standard deduction. Do we still need to double-branch if we do that?

PavelMakarchuk: I see, 

I think the issue with using deductions is that some states apply the federal deduction amount (either SD or Item or both) so we can't sure `state_itemized_deductions` and `state_standard_deduction` consistently

MaxGhenis: we don't have a state standard deduction that inherits from federal as needed for taxsim validation?

PavelMakarchuk: No, it is currently a sum of the state standard deduction list. We can't add the standard deduction variable since it will be added to every state. We currently implement the logic in the emulator directly but we can add a formula to the [state_standard_deduction](https://github.com/PolicyEngine/policyengine-us/blob/master/policyengine_us/variables/gov/states/tax/income/state_standard_deduction.py) variable and migrate to the `-us` repo

MaxGhenis: Ok yes let's do that in this pr or in advance of it to avoid branching

PavelMakarchuk: After the new implementation approach we introduced a circular reference error - fixing now

PavelMakarchuk:
For documentation purposes. Circular reference errors when using state itemized deductions:

`mt_federal_income_tax_deduction_unit` -> depends on `income_tax_before_refundable_credits`

`ia_fedtax_deduction` -> `income_tax_before_refundable_credits`

`la_itemized_deductions` -> `deductions_if_itemizing` -> `qbid` -> `taxable_income_less_qbid` -> itemization choice 

`me_itemized_deductions_pre_phaseout` depends on itemization choice -> simple fix to migrate that logic to `me_deductions`

PavelMakarchuk:
> For documentation purposes. Circular reference errors when using state itemized deductions:
> 
> * [ ]  `mt_federal_income_tax_deduction_unit` -> depends on `income_tax_before_refundable_credits`
> * [ ]  `ia_fedtax_deduction` -> `income_tax_before_refundable_credits`
> * [ ]  `la_itemized_deductions` -> `deductions_if_itemizing` -> `qbid` -> `taxable_income_less_qbid` -> itemization choice
> * [x]  `me_itemized_deductions_pre_phaseout` depends on itemization choice -> simple fix to migrate that logic to `me_deductions`

In Louisiana our current logic should be adjusted to use `itemized_taxable_income_deductions` instead of `taxable_income_deductions_if_itemizing` . This is an error in the model itself since the [tax form](https://dam.ldr.la.gov/taxforms/IT540i%20WEB%20(2024)D18%20INSTRUCTIONS.pdf#page=3) states to use the deductions from [schedule A](https://www.irs.gov/pub/irs-pdf/f1040sa.pdf) which does not include QBID

PavelMakarchuk:
> For documentation purposes. Circular reference errors when using state itemized deductions:
> 
> * [ ]  `mt_federal_income_tax_deduction_unit` -> depends on `income_tax_before_refundable_credits`
> * [ ]  `ia_fedtax_deduction` -> `income_tax_before_refundable_credits`
> * [x]  `la_itemized_deductions` -> `deductions_if_itemizing` -> `qbid` -> `taxable_income_less_qbid` -> itemization choice
> * [x]  `me_itemized_deductions_pre_phaseout` depends on itemization choice -> simple fix to migrate that logic to `me_deductions`

For Iowa and Montana, the issue are the federal tax deductions which are counted as itemized deductions. There seem to be two options to account for the circular reference:

1. Use an intermediate fed. tax deduction value, applying the tax withholding amount (used in the SALT deduction) to determine itemization and keep the final state tax calculation 

2. omit the federal tax deductions from the computation for the itemization choice

policyengine_us/variables/gov/irs/income/taxable_income/deductions/tax_unit_itemizes.py
```diff
@@ -18,11 +19,29 @@ def formula(tax_unit, period, parameters):
...
+            state_standard_deduction = tax_unit(
+                "state_standard_deduction", period
+            )
+            state_itemized_deductions = tax_unit(
+                "state_itemized_deductions", period
+            )
+            # Use a small tolerance for floating-point comparison due to floating point imprecision
+            tolerance = 0.0001
+            federal_tax_equal = (
+                abs(
```
MaxGhenis: This is base `abs` which isn't vectorized. We don't do `from numpy import abs as abs` in https://github.com/PolicyEngine/policyengine-core/blob/master/policyengine_core/model_api.py
```suggestion
                np.abs(
```

policyengine_us/variables/gov/irs/income/taxable_income/deductions/tax_unit_itemizes.py
```diff
@@ -18,11 +19,29 @@ def formula(tax_unit, period, parameters):
...
             )
-            return tax_liability_if_itemizing < tax_liability_if_not_itemizing
+            state_standard_deduction = tax_unit(
+                "state_standard_deduction", period
+            )
+            state_itemized_deductions = tax_unit(
+                "state_itemized_deductions", period
+            )
+            # Use a small tolerance for floating-point comparison due to floating point imprecision
+            tolerance = 0.0001
```
MaxGhenis:
```suggestion
            TOLERANCE = 0.0001
```

policyengine_us/variables/gov/states/tax/income/state_itemized_deductions.py
```diff
@@ -8,3 +8,70 @@ class state_itemized_deductions(Variable):
...
+                "joint": "de_itemized_deductions_joint",
+            },
+            "AR": {
+                "indiv": "ar_itemized_deductions_indiv",
+                "joint": "ar_itemized_deductions_joint",
+            },
+        }
+
+        # Calculate maximum itemized deductions for applicable states
+        state_specific_with_max = state_specific_base
```
MaxGhenis: clean up this and line 64, shouldn't need reassignment. same with state standard

policyengine_us/variables/gov/states/tax/income/state_standard_deduction.py
```diff
@@ -7,4 +7,76 @@ class state_standard_deduction(Variable):
...
+            joint_deductions = add(tax_unit, period, [variables["joint"]])
+            max_deductions = np.maximum(indiv_deductions, joint_deductions)
+            state_specific_with_max = where(
+                is_state, max_deductions, state_specific_with_max
+            )
+
+        state_specific = state_specific_with_max
+
+        # Check if the state adopts federal standard deduction
+        uses_federal = np.isin(state_code, FEDERAL_STANDARD_DEDUCTION_STATES)
```
MaxGhenis, line 71: would a defined_for improve performance here?

changelog_entry.yaml
```diff
@@ -0,0 +1,4 @@
+- bump: patch
+  changes:
+    fixed:
+    - Default itemization to be based on state taxes when federal tax liability is equal.
```
MaxGhenis: Complete

MaxGhenis (changes requested):

policyengine_us/variables/gov/states/tax/income/state_standard_deduction.py
PavelMakarchuk: We can separate the state itemized and standard deductions into multiple calculations (states that adopt federal logic, states with married filing separately, states with state-specific logic ). I would like to do this in separate PR just to avoid adding complexity here, but it's not a heavy lift

PR #6592, opened 2025-09-24
Adjust the `state_agi` variable to reflect states that adopt federal AGI

### LINKED ISSUE #6591: Adjust the `state_agi` variable to reflect states that adopt federal AGI
We should also reflect married filing separately situations similar to state standard and itemized deductions

PavelMakarchuk:
Fixes #6591


policyengine_us/variables/gov/states/tax/income/state_agi.py
```diff
@@ -8,3 +8,57 @@ class state_agi(Variable):
...
+        # Get the sum of state-specific AGIs
+        state_specific_base = add(
+            tax_unit,
+            period,
+            parameters(period).gov.states.household.state_agis,
+        )
+
+        # Special handling for states that have separate individual and joint AGI calculations
+        # Arkansas and Delaware use different AGI calculations based on filing status
+        STATES_WITH_INDIVIDUAL_JOINT_AGI = {
```
MaxGhenis: can we just make it AR and DE since the variable names derive consistently from the state?

PavelMakarchuk: Instead of `STATES_WITH_INDIVIDUAL_JOINT_AGI` ?

MaxGhenis: `STATES_WITH_INDIVIDUAL_JOINT_AGI = ["AR", "DE"]` then string operations to get the variables from there

PavelMakarchuk: done

PR #6601, opened 2025-09-24
Include the missing state benefit programs in the computation tree

### LINKED ISSUE #6598: Include the DART program in the computation tree
Currently the `tx_dart_benefit_person` does not affect the computation tree

PavelMakarchuk:
Fixes #6598


PavelMakarchuk:
> As part of this PR, I am testing whether we can produce microsim results for each program which is breaking in some instances, fixing and commenting as I go along

There seems to issues to apply yearly values in monthly variables when the entity types differ. As seen in `il_aabd_shelter_allowance` @hua7450 FYI

PR #6891, opened 2025-12-02
Fix StateGroup enum warnings for contiguous US states

### LINKED ISSUE #6890: StateGroup enum warnings for contiguous US states
## Description

When running simulations or calculations, warnings are emitted for contiguous US states:

```
Invalid values for enum StateGroup: ['AL']. These will be encoded as index 0.
Invalid values for enum StateGroup: ['AZ']. These will be encoded as index 0.
Invalid values for enum StateGroup: ['AR']. These will be encoded as index 0.
Invalid values for enum StateGroup: ['CA']. These will be encoded as index 0.
```

## Root Cause

In `policyengine_us/variables/household/demographic/geographic/state_group.py`, the formula uses `where()` which evaluates both branches before applying the condition:

```python
return where(
    np.isin(state_code, NON_CONTIGUOUS_STATES),
    StateGroup.encode(state_code).decode(),  # Called on ALL states
    StateGroup.CONTIGUOUS_US,
)
```

The `StateGroup` enum only has members for non-contiguous states (AK, HI, GU, PR, VI), so when `encode()` is called on contiguous states like AL, AZ, etc., it triggers the warning.

## Fix

Refactor the formula to only pass non-contiguous states to `encode()`:

```python
is_non_contiguous = np.isin(state_code, NON_CONTIGUOUS_STATES)
encoded = np.empty(state_code.shape, dtype=object)
encoded[is_non_contiguous] = StateGroup.encode(
    state_code[is_non_contiguous]
).decode()
encoded[~is_non_contiguous] = StateGroup.CONTIGUOUS_US
return encoded
```

## Impact

- Warnings only; no functional impact on calculations
- All existing tests pass with the fix

baogorek:
## Summary

- Fixes warnings like `Invalid values for enum StateGroup: ['AL']. These will be encoded as index 0.`
- The `where()` function evaluates both branches before applying the condition, causing `StateGroup.encode()` to be called on all state codes including contiguous US states that don't have enum members
- Refactored to only pass non-contiguous states (AK, HI, GU, PR, VI) to `encode()`

## Test plan

- [x] All 21 geographic variable tests pass
- [x] state_group.yaml tests pass
- [x] Manual testing confirms no warnings with mixed state codes
- [x] `make format` passes

Fixes #6890


baogorek:
@MaxGhenis is a yaml okay for a test? The latest addition has a few blocks like:
```
       15 +  - name: Contiguous state does not cause empty array indexing error
       16 +    period: 2024
       17 +    input:
       18 +      state_code: TX
       19 +    output:
       20 +      state_group: CONTIGUOUS_US
```

Before, this PR unconditionally called StateGroup.encode(state_code[is_non_contiguous]). When all states are contiguous (common case), that slice is empty, and encode() crashes trying to access array[0].

Now we only encode if there's something to encode. Fingers crossed on these tests.

PR #7021, opened 2025-12-19
Fix county variable period carry-over for dataset simulations

### LINKED ISSUE #7020: County variable loses assigned value when calculating for different periods on dataset simulations
## Problem

When running a simulation over a dataset (e.g., the NYC H5 file), the `county` variable returns incorrect values when calculated for a period other than the one stored in the dataset.

### Reproduction

```python
from policyengine_us import Microsimulation

sim = Microsimulation(dataset="hf://policyengine/policyengine-us-data/cities/NYC.h5")

# Base period (2025) - correct NYC counties
sim.calculate("county")
# Returns: QUEENS_COUNTY_NY, BRONX_COUNTY_NY, etc.

# Different period (2026) - WRONG: falls back to first county alphabetically
sim.calculate("county", period=2026)
# Returns: ALBANY_COUNTY_NY for all households
```

## Root Cause

The `county` variable has a formula, unlike `state_fips` and `congressional_district_geoid` which are pure input variables. Variables without formulas automatically carry over their values via `auto_carry_over_input_variables`. However, since `county` has a formula:

1. When requesting `county` for period 2026 (not stored), the formula runs
2. The formula calls `household("county_fips", 2026)` which returns nothing (only 2025 data exists)
3. The formula falls through to `first_county_in_state` as a fallback
4. This returns ALBANY_COUNTY_NY (alphabetically first in NY)

## Impact

This affects any dataset-based simulation where:
- Counties were explicitly assigned (e.g., stacked CD datasets with county assignments)
- Users calculate county-dependent variables for periods other than the base dataset period

## Proposed Fix

Add an early-exit check in the `county` formula: when running over a dataset, check if `county` has stored data for any period and return that value directly, treating county as a time-invariant geographic variable.

baogorek:
## Summary
- Fixes county variable losing its stored value when calculating for periods other than the base dataset period
- When running over a dataset, the formula now checks for stored county data and returns it directly

## Test plan
- [x] Tested with NYC.h5 dataset - counties correctly persist from 2025 to 2026
- [x] Verified no ALBANY_COUNTY_NY fallback occurs

Fixes #7020

### DIFF
--- a/policyengine_us/tests/run_selective_tests.py
+++ b/policyengine_us/tests/run_selective_tests.py
@@ -60,6 +60,11 @@ class SelectiveTestRunner:
                 "file_pattern": r"policyengine_us/variables/household/",
                 "test_pattern": r"policyengine_us/tests/policy/baseline/household",
             },
+            # Match household variables to microsimulation tests (for dataset-level tests)
+            {
+                "file_pattern": r"policyengine_us/variables/household/",
+                "test_pattern": r"policyengine_us/tests/microsimulation",
+            },
             # Match reforms in specific states
             {
                 "file_pattern": r"policyengine_us/reforms/states/([^/]+)",


policyengine_us/tests/run_selective_tests.py
```diff
@@ -60,6 +60,11 @@ def __init__(self, base_branch: str = "master"):
                 "file_pattern": r"policyengine_us/variables/household/",
                 "test_pattern": r"policyengine_us/tests/policy/baseline/household",
             },
+            # Match household variables to microsimulation tests (for dataset-level tests)
```
baogorek, line 63: This is me flailing here trying to get the code coverage

hua7450: If's fine, we can ignore this coverage failing check.

PR #7271, opened 2026-01-28
Exclude Medicaid recipients from QI eligibility

### LINKED ISSUE #7270: Exclude Medicaid recipients from QI elibility

daphnehanse11:
## Summary

- Added Medicaid exclusion check to `is_qi_eligible.py`
- Per medicare.gov: "QI is only available for people who don't qualify for any other Medicaid coverage or benefits."

## Changes

- `is_qi_eligible.py`: Added check `& ~medicaid_eligible` to exclude Medicaid recipients
- `msp_category.yaml`: Updated QI test to explicitly set `is_medicaid_eligible: false`
- `is_qi_eligible.yaml`: New test file verifying Medicaid exclusion works

## Test plan

- [x] New `is_qi_eligible.yaml` tests pass
- [x] Existing `msp_category.yaml` tests pass

Fixes #7270

### DIFF
--- a/policyengine_us/tests/policy/baseline/gov/hhs/medicare/savings_programs/msp_benefit_value.yaml
+++ b/policyengine_us/tests/policy/baseline/gov/hhs/medicare/savings_programs/msp_benefit_value.yaml
@@ -43,6 +43,8 @@
     spm_unit_is_married: false
     base_part_a_premium: 0
     base_part_b_premium: 2_096.40
+    # QI requires not being eligible for other Medicaid
+    is_medicaid_eligible: false
   output:
     msp_benefit_value: 174.70
 


changelog_entry.yaml
```diff
@@ -0,0 +1,4 @@
+- bump: patch
+  changes:
+    fixed:
+    - Exclude Medicaid recipients from QI (Qualifying Individual) eligibility.
```
MaxGhenis, line 4: for future reference should read as:
```
Disqualify Medicaid-eligible individuals from QI (Qualifying Individual) coverage.
```

MaxGhenis: also what source are we citing for this? is it an existing one?

PR #7377, opened 2026-02-12
Fix Michigan surtax reform not activating when in_effect is toggled

### LINKED ISSUE #7376: Fix Michigan surtax reform not activating when in_effect is toggled
## Bug

The Michigan surtax ballot initiative (PR #6127) does not work in the app when users toggle `in_effect` to true. The simulation shows no impact even for high-income filers in 2027.

## Root Cause

The reform was checking `in_effect` at **system initialization time** (in `create_mi_surtax_reform`), when the parameter always has its default value of `false`. When users later change the parameter to `true` in the app, the structural reform check has already happened and the reform is never applied.

## Fix

1. Always apply the structural reform (create the `mi_surtax` variable)
2. Check `in_effect` in the formula at calculation time, returning 0 when false

This follows the correct pattern for user-toggled reform parameters.

DTrim99:
## Summary

Fixes #7376

The Michigan surtax ballot initiative reform was not working in the app when users toggled `in_effect` to true. Users reported "no results" even for high-income filers in 2027.

## Root Cause

The `create_mi_surtax_reform` function was checking `in_effect` at **system initialization time**, when the parameter always has its default value of `false`. When users later change the parameter to `true` in the app, the structural reform check has already happened and the reform was never applied.

## Fix

1. **Always apply the structural reform** - removed the `in_effect` check from `create_mi_surtax_reform`
2. **Check `in_effect` in the formula** - the `mi_surtax` formula now checks `in_effect` at calculation time and returns 0 when false

This follows the correct pattern for user-toggled reform parameters.

## Test plan

- [x] All 9 existing + new tests pass
- [x] Added test case verifying surtax is 0 when `in_effect` is false
- [ ] Verify in app that toggling `in_effect` to true now shows impact for MI high-income filers in 2027

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,4 @@
+- bump: patch
+  changes:
+    fixed:
+      - Michigan surtax reform now activates correctly when in_effect parameter is toggled in the app.
--- a/policyengine_us/reforms/states/mi/surtax.py
+++ b/policyengine_us/reforms/states/mi/surtax.py
@@ -1,5 +1,4 @@
 from policyengine_us.model_api import *
-from policyengine_core.periods import period as period_
 
 
 def create_mi_surtax() -> Reform:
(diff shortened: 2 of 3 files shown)


MaxGhenis (approved):
LGTM - Clean fix for the initialization-time vs calculation-time bug. Moving the in_effect check from create_mi_surtax_reform (build time) into the formula (calculation time) is the correct pattern for user-toggled reform parameters. The parameter path adjustment (p.rate.joint/single instead of p.joint/single) correctly follows from reading p at the surtax level rather than surtax.rate. Good test coverage for the false case.

MaxGhenis (approved):
LGTM - Good bug fix. Moving the in_effect check from initialization time into the formula ensures the parameter is evaluated at calculation time when users toggle it. The parameter path adjustment is correct, tests cover the fix, and all CI checks pass.

PR #7401, opened 2026-02-14
Wire up all 51 state TANF programs, fix childcare cycle artifact, add takeup

MaxGhenis:
## Summary

Wires up all 51 state TANF programs into the `tanf` variable, adds takeup support, removes the `tanf_reported` short-circuit, and fixes two CO CCAP issues that created circular dependencies.

Closes #7400

### TANF wiring and takeup

- **`tanf.py`**: Removed `tanf_reported` short-circuit; added all 51 state programs; applied takeup multiplier
- **`takes_up_tanf_if_eligible.py`**: New bool variable (default True; set by microdata for microsim)
- Added KS, KY, MT, NH, TN, AR, ID, IA, LA, MA, NE, OH, UT, WV, WY to state list

### Childcare cycle fix (false cycle removed)

An audit revealed that the pre-subsidy childcare workaround across 24 state TANF programs was unnecessary. The suspected cycle ran: TANF → state programs → childcare deduction → `childcare_expenses` → childcare subsidies (CO CCAP) → SNAP → TANF.

**Root cause**: CO CCAP's `co_ccap_countable_income` used SNAP income as a placeholder instead of its actual statutory income sources. No childcare subsidy program (CA, CO, NE, MA) includes SNAP in its income definition, so no cycle existed through this path.

**Fix**: Replaced the placeholder with the actual income sources from [8 CCR 1403-1 §7.105](https://www.sos.state.co.us/CCR/GenerateRulePdf.do?ruleVersionId=11042&fileName=8%20CCR%201403-1#page=22), extracted to a YAML parameter at `gov.states.co.ccap.income.countable_income.sources`.

**Result**: All 24 state TANF programs and SNAP's dependent care deduction now use `childcare_expenses` (post-subsidy) instead of the pre-subsidy workaround, which is more accurate.

### CO CCAP FPG cycle fix (real cycle fixed)

A separate real cycle existed through CO CCAP's FPG eligibility check: TANF → `childcare_expenses` → `co_ccap_subsidy` → `co_ccap_fpg_eligible` → `snap_fpg` → `snap_unit_size` → `is_snap_ineligible_student` → `tanf_person` → TANF.

CO CCAP was borrowing SNAP's FPG variable, which depends on SNAP unit size (which excludes ineligible students based on TANF receipt).

[...]

PR #7412, opened 2026-02-17
Fix false childcare cycle: use correct CO CCAP income, switch 24 TANF states and SNAP to childcare_expenses

MaxGhenis:
## Summary

- **Root cause fix**: CO CCAP's `co_ccap_countable_income` used SNAP income variables as a placeholder (with a TODO), creating a false dependency chain through the static graph. Replaced with actual income sources per [8 CCR 1403-1 Section 7.105](https://www.sos.state.co.us/CCR/GenerateRulePdf.do?ruleVersionId=11042&fileName=8%20CCR%201403-1#page=22).
- **24 state TANF programs**: Switched from `spm_unit_pre_subsidy_childcare_expenses` workaround to `childcare_expenses` (the accurate post-subsidy amount). States: AK, AL, AZ, DC, DE, GA, HI, IL, KS, KY, MD, ME, MN, MO, MT, NH, NM, OK, RI, TN, TX, VA, VT, WV.
- **SNAP dependent care deduction**: Switched from `pre_subsidy_childcare_expenses` to `childcare_expenses` — the workaround is no longer needed since no childcare subsidy program (CA, CO, NE, MA) depends on SNAP income.
- **65+ test files** updated to match.

### Why the cycle was false

The suspected cycle ran: `tanf` → state TANF → childcare deduction → `childcare_expenses` → childcare subsidies → SNAP → `tanf`. An audit found:

1. The **only** path back to SNAP was through CO CCAP, which used `snap_earned_income` + `snap_unearned_income` as a placeholder
2. California's childcare subsidies use their own income definition (no SNAP)
3. Nebraska uses AGI
4. Massachusetts explicitly exempts SNAP from countable income

With CO CCAP fixed to use the correct statutory income sources, the cycle no longer exists.

## Test plan

- [x] Model loads without circular dependency errors (`Microsimulation()` succeeds)
- [x] All 205 SNAP tests pass
- [x] All 58 CO CCAP tests pass
- [x] All 11 childcare expense tests pass
- [x] `make format` applied
- [ ] CI passes

### DIFF
--- a/policyengine_us/parameters/gov/states/co/ccap/income/countable_income/sources.yaml
+++ b/policyengine_us/parameters/gov/states/co/ccap/income/countable_income/sources.yaml
@@ -1,0 +1,31 @@
+description: Colorado counts these income sources as countable income under the Child Care Assistance Program.
+values:
+  2020-01-01:
+    # Earned income
+    - employment_income
+    - self_employment_income
+    - farm_income
+    # Unearned income
+    - social_security
+    - pension_income
+    - retirement_distributions
+    - military_retirement_pay
+    - unemployment_compensation
+    - workers_compensation
+    - child_support_received
+    - alimony_income
+    - interest_income
+    - dividend_income
+    - rental_income
+    - veterans_benefits
+    - disability_benefits
+    - capital_gains
+    - gi_cash_assistance
+
+metadata:
+  unit: list
+  period: year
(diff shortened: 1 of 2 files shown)


MaxGhenis:
Merged into #7401 — single PR now covers both TANF wiring and the childcare cycle fix.

PR #7700, opened 2026-03-04
Add extend_single_year_dataset for fast dataset year projection

### LINKED ISSUE #7699: Add extend_single_year_dataset for fast dataset year projection
## Motivation

The API v2 alpha's `create_datasets()` is extremely slow (~1hr+ per state) because it routes every variable through `sim.calculate()`, invoking the full simulation engine for each variable × each year. The UK avoids this entirely: `policyengine-uk` has `extend_single_year_dataset()` which uprates DataFrames via simple multiplication — no simulation engine needed.

Now that `policyengine-us-data` is being updated to publish entity-level HDFStore files alongside the existing h5py files (PolicyEngine/policyengine-us-data#567), `policyengine-us` needs the machinery to consume them.

### 1. Dataset schema classes (`data/dataset_schema.py`)

- `USSingleYearDataset` — holds one DataFrame per entity (person, household, tax_unit, spm_unit, family, marital_unit) + time_period. Supports load/save/copy.
- `USMultiYearDataset` — holds a dict of `USSingleYearDataset` indexed by year. `load()` returns `{variable: {year: array}}` compatible with policyengine-core's `TIME_PERIOD_ARRAYS` format.

### 2. `extend_single_year_dataset()` (`data/economic_assumptions.py`)

Copies base-year DataFrames for each year through `end_year` (default 2035), then applies multiplicative uprating:

- Reads uprating parameter path from `system.variables[var].uprating` at runtime
- Computes growth factor as `param(current_year) / param(prev_year)` (absolute index ratio)
- Variables without uprating are carried forward unchanged
- No separate uprating list to maintain — picks up changes from variable definitions and `default_uprating.py` automatically

### 3. Dual-path loading in `Microsimulation` (`system.py`)

- `_is_hdfstore_format()` — detects entity-level HDFStore vs variable-centric h5py by checking top-level HDF5 keys
- `_resolve_dataset_path()` — resolves HuggingFace URLs and local paths
- `Microsimulation.__init__` — when an HDFStore file is detected, loads it as `USSingleYearDataset`, extends via `extend_single_year_dataset`, and passes the resulting `USMultiYearDataset` to the parent class
- Legacy h5py files continue to work via the existing code path

## Depends on

- PolicyEngine/policyengine-us-data#567 (HDFStore output format)

anth-volk:
Fixes #7699

## Why this is needed

The API v2 alpha and the `policyengine` Python package require **entity-level Pandas HDFStore datasets** (one table per entity: person, household, tax_unit, etc.) to run microsimulations. The current US data pipeline (`policyengine-us-data`) publishes **variable-centric h5py files** (`variable/year → array`), so converting between the two formats currently requires routing every variable through `sim.calculate()` via `create_datasets()` — a process that takes over an hour per state and doesn't scale to the 500+ geographic datasets we need to serve.

The UK avoids this entirely: `policyengine-uk-data` publishes entity-level HDFStore files directly, and `policyengine-uk` has `extend_single_year_dataset()` which projects a single base-year dataset to multiple years via simple multiplicative uprating on DataFrames — no simulation engine involved. This PR brings the same capability to the US.

### Dataset schema classes (`dataset_schema.py`)

`USSingleYearDataset` holds six entity DataFrames (person, household, tax_unit, spm_unit, family, marital_unit) plus a `time_period`. It can load from / save to Pandas HDFStore files, and provides `.copy()` for deep-copying all DataFrames.

`USMultiYearDataset` wraps a `dict[int, USSingleYearDataset]` keyed by year. Its `.load()` returns data in `{variable: {year: array}}` format (`time_period_arrays`), which is what policyengine-core's `Microsimulation` expects for multi-year datasets.

### Uprating logic (`economic_assumptions.py`)

`extend_single_year_dataset(dataset, end_year=2035)` takes a single base-year dataset and produces a multi-year dataset by:

1. **Copying** the base-year DataFrames for each year from `base_year` through `end_year`
2. **Applying multiplicative uprating** year-over-year: for each variable column, it looks up `system.variables[var].uprating` to get a dotted parameter path (e.g. `"calibration.gov.irs.soi.employment_income"`), resolves it against `system.parameters`, and computes `factor = param(current_year) / param(previous_year)`. The column values are then multiplied by that factor.
3. **Carrying forward** variables without an uprating parameter unchanged (e.g. `age`, entity IDs).

This is the same approach used by `policyengine-uk`. The uprating mapping is derived entirely from `system.variables` at runtime — the 62 variables with explicit `uprating = "..."` and the 108 variables assigned via `default_uprating.py` are all picked up automatically. No separate list to maintain.

### Dual-path loading (`system.py`)

`Microsimulation.__init__` now auto-detects dataset format before calling `super().__init__()`:

- **HDFStore format** (entity names like `person`, `household` as top-level HDF5 keys): loads as `USSingleYearDataset`, extends via `extend_single_year_dataset()`, and passes the resulting `USMultiYearDataset` to policyengine-core.
- **Legacy h5py format** (variable names as top-level keys): falls through to the existing `CoreMicrosimulation` code path, unchanged.

Format detection (`_is_hdfstore_format`) inspects the top-level HDF5 keys — entity names indicate HDFStore, variable names indicate h5py.

### Unit tests (22 tests, ~0.3s)

The test suite in `tests/microsimulation/data/` uses mock system objects (mock parameters, mock variables) to avoid loading the full tax-benefit system, keeping tests fast and deterministic. Coverage includes:

- **`_resolve_parameter`** (3 tests): valid dotted path, invalid path, partially valid path
- **`_apply_single_year_uprating`** (7 tests): correct multiplicative scaling, non-uprated variables unchanged, household entity uprating, unknown columns passed through, unresolvable uprating path, division-by-zero guard (previous param value = 0), zero base values preserved
- **`extend_single_year_dataset`** (12 tests): correct year count, single-year edge case, default end year (2035), base year values unchanged, year 1 uprating, year 2 chaining (verifies uprating compounds from year N to N+1 to N+2, not from base), non-uprated variable identical across all years, row counts preserved, time_period correctness per year, return type, input dataset immutability, multi-entity uprating (person + household)

### Roundtrip validation (policyengine-us-data PR #568)

A separate one-off validation script in `-us-data` reads an existing h5py state dataset (e.g. NV.h5), converts it to HDFStore using the same splitting logic, and compares all ~183 variables between the two formats. This passed 183/183 on the Nevada dataset.

## Depends on

- [PolicyEngine/policyengine-us-data#568](https://github.com/PolicyEngine/policyengine-us-data/pull/568) — adds HDFStore output format alongside h5py in the data pipeline

## Test plan
- [ ] `make test-other` passes (runs the 22 unit tests via pytest)
- [ ] Load an HDFStore file via `Microsimulation(dataset="path/to/STATE.hdfstore.h5")` — verify it loads and extends correctly
- [ ] Load a legacy h5py file via `Microsimulation(dataset="path/to/STATE.h5")` — verify existing path still works
- [ ] Verify uprated variables (e.g. `employment_income`) grow year-over-year
- [ ] Verify non-uprated variables (e.g. `age`) are carried forward unchanged


PavelMakarchuk:
> Which of the following are you saying:
> 
> * We need to generally support uprating out to 2100
> * We need USSingleYear datasets out to 2100
> * We need USSingleYear datasets to automatically extend out to whatever year we update the CPI-U to (for now, 2035)
> * Something else?

We need USSingleYear datasets to automatically extend out to whatever year we update the CPI-U to (for now, 2035)

PR #7706, opened 2026-03-05
Add Working Americans' Tax Cut Act reform

PavelMakarchuk:
## Summary
- Implements the Working Americans' Tax Cut Act (WATCA) as a contributed structural reform
- **Cost of living exemption**: Filing-status-indexed deduction from taxable income ($46K single, $64.4K HOH, $92K joint) with linear phase-out up to 175% of the exemption amount
- **Millionaire surtax**: Tiered marginal surtax on AGI (single: 5%/$1M, 10%/$2M, 12%/$5M; joint: 5%/$1.5M, 10%/$3M, 12%/$7.5M) added to income tax before credits

## Test plan
- [x] 15 YAML integration tests covering exemption phase-out, surtax brackets, taxable income reduction, and income_tax_before_credits integration
- [ ] CI passes

### DIFF
--- a/policyengine_us/tools/variables.py
+++ b/policyengine_us/tools/variables.py
@@ -13,9 +13,7 @@ def add_n(word):
 
 def print_variable_summary(variable_name: str):
     variable = variables.get(variable_name)
-    return Markdown(
-        f"""
+    return Markdown(f"""
         ## {variable.name}
         This variable models a{add_n(variable.entity.label)}**{variable.entity.label}**'s **{variable.label}**.
-        """
-    )
+        """)


policyengine_us/reforms/congress/watca/working_americans_tax_cut_act.py
```diff
@@ -0,0 +1,109 @@
...
+        unit = USD
+
+        def formula(tax_unit, period, parameters):
+            agi = tax_unit("adjusted_gross_income", period)
+            p = parameters(period).gov.contrib.congress.watca.surtax
+            filing_status = tax_unit("filing_status", period)
+            joint = filing_status == filing_status.possible_values.JOINT
+            return where(
+                joint,
+                p.rate.joint.calc(agi),
```
daphnehanse11, line 59: SURVIVING_SPOUSE should use joint surtax thresholds here — currently it falls through to the single rate. Consider:                                   
  joint = (filing_status == filing_status.possible_values.JOINT) | (                                                                                    
      filing_status == filing_status.possible_values.SURVIVING_SPOUSE                                                                                   
right?

policyengine_us/reforms/congress/watca/working_americans_tax_cut_act.py
```diff
@@ -0,0 +1,111 @@
...
+
+        def formula(tax_unit, period, parameters):
+            agi = tax_unit("adjusted_gross_income", period)
+            p = parameters(period).gov.contrib.congress.watca.surtax
+            filing_status = tax_unit("filing_status", period)
+            joint = (filing_status == filing_status.possible_values.JOINT) | (
+                filing_status == filing_status.possible_values.SURVIVING_SPOUSE
+            )
+            return where(
+                joint,
```
daphnehanse11, line 58: **Filing status enum comparison may not work in microsimulation context**

In `watca_millionaire_surtax`, the joint/single surtax selection uses:
```python
joint = filing_status == filing_status.possible_values.JOINT
```

When running against the Enhanced CPS dataset, `filing_status` values are strings like `'JOINT'`, but `filing_status.possible_values.JOINT` encodes as `'Joint'` — the comparison always returns `False`. This means:

1. **Joint filers with AGI $1M–$1.5M incorrectly pay surtax** (they hit the single $1M threshold instead of the joint $1.5M threshold)
2. **Joint filers above $1.5M use the wrong (single) bracket schedule** — lower thresholds than intended

This doesn't affect the YAML unit tests (which use the enum correctly), but it would produce wrong results in any microsimulation or population-level analysis.

Separately, the same pattern is used in `watca_cost_of_living_exemption` for the `filing_status` breakdown lookup — worth checking that the parameter bracket lookup handles this correctly too.

policyengine_us/reforms/congress/watca/working_americans_tax_cut_act.py
MaxGhenis: Already addressed — SURVIVING_SPOUSE is included in the joint check on line 38-39:
```python
joint = (filing_status == filing_status.possible_values.JOINT) | (
    filing_status == filing_status.possible_values.SURVIVING_SPOUSE
)
```

policyengine_us/reforms/congress/watca/working_americans_tax_cut_act.py
MaxGhenis: This pattern (`filing_status == filing_status.possible_values.JOINT`) is the standard PolicyEngine idiom and works correctly in both unit test and microsimulation contexts — the framework's enum encoding handles the comparison. The `cost_of_living_exemption` uses parameter bracket lookup (`p.amount[filing_status]`) which also handles this correctly.

daphnehanse11 (approved):

Issue #7751, opened 2026-03-11
Proposal: revamping HR1 rules for Medicaid Community Engagement


harrisj:
Hello,

While I appreciate there has been some initial work related to HR1 compliance (see the is_medicaid_eligible variable for instance), I think it has a few issues in its implementation that I would like to correct:

- It seems like it's applying Medicaid work requirements to all when they really just apply to people in the Medicaid expansion population. Is there a definition to tell if a user is in_medicaid_expansion population?
- Grouping compliance into separate major buckets for exemptions, hardships (temp. exemptions for things like disasters, etc.) as well as compliance might be useful. Eligibility is then rolling up those specific categories
- CMS' current interpretation of the rule is that lookback period for an application is the 6 months immediately before the month of the current application date, but that the state can use any single month in that lookback to show compliance (meaning we should test each individual month)
- States actually have the option of picking 1-3 months to validate for CE. We expect most states will pick one, but it's something to have to support.

Here is a non-complete list of the different type of variables at play (more to fill in later), with a check of implementation progress for an associated PR
- [ ] Exemptions
   - [ ] Pregnant
   - [ ] Under 19 or over 65
   - [ ] Taking care of a child 13 or younger
   - [ ] Taking care of a person with a disability
   - [ ] Is pregnant
   - [ ] Blind
   - [ ] Deaf
   - [ ] Disabling mental health condition (to be defined)
   - [ ] Medically frail (to be defined)
   - [ ] Has a substance use disorder and/or is in treatment
   - [ ] Disabled under SSA definition
   - [ ] American Indian or California Indian or receiving services from the IHS
   - [ ] Has been incarcerated in the past 3 months
- [ ] Compliance
   - [ ] Income
     - [ ] Earned more than $580 (80xfederal minimum wage) in a 
     - [ ] Earned more than $580x3 in a single quarter (if it fits in prior 6 months)
   - [ ] Worked more than 80 hours in a month (combined from work, volunteering or job training) 
   - [ ] Enrolled half or full-time in an educational institution
- [ ] Hardship
  - [ ] County of residence unemployment the min of (1.5x national avg or 8%).  
  - [ ] County of residence was declared a federal disaster site
  - [ ] Having to travel outside of county long distances to receive medical care

Possible challenges:
- Is identifying if someone is in the Medicaid expansion group straightforward > 1 PFL but <= 1.38x PFL? or does it vary depending on the state and the person's Medicaid category?
[...]

PR #7752, opened 2026-03-11
Reworking HR1 Medicaid Rules

harrisj:
See #7751 for details, but this is a plan to revise some of the code for HR1 Medicaid work requirements to better match what CMS is providing as its provisional rules


daphnehanse11 (commented):
Blocking: several work-requirement formulas use Python or with PolicyEngine arrays instead of vectorized bitwise operators. For example, medicaid_work_requirement_eligible returns has_exemption or has_ce, has_medicaid_work_requirement_ce returns enrolled_half_time or passed_hours or passed_income, and has_medicaid_work_requirement_exemption chains the exemption flags with or. This raises ValueError: The truth value of an array with more than one element is ambiguous in multi-person simulations. Repro on this branch with two people in 2027-03: has_medicaid_work_requirement_ce, medicaid_work_requirement_eligible, and yearly is_medicaid_eligible all raise that error. Please replace these with parenthesized vectorized expressions using |.

PR #7753, opened 2026-03-11
Gate aca_ptc on tax_unit_is_filer for non-filers

### LINKED ISSUE #7748: aca_ptc should return 0 for non-filers (produces 5x overestimate in calibrated datasets)
## The problem

`aca_ptc.formula_2024` returns nonzero values for non-filing tax units. In calibrated microsimulation datasets, this inflates total ACA PTC from the IRS target (~$112M for SC) to ~$578M — a roughly 5× overestimate.

## How calibrated datasets work

For readers unfamiliar with [policyengine-us-data](https://github.com/PolicyEngine/policyengine-us-data):

- policyengine-us-data calibrates CPS survey weights so that `sum(variable × weight)` matches administrative targets
- ACA PTC targets come from IRS SOI (Statistics of Income) data, which by definition covers only tax return filers
- The calibration matrix applies a `tax_unit_is_filer == 1` mask, so only filer tax units' `aca_ptc` values are constrained to the IRS target
- The calibration works correctly — filer-only `aca_ptc` matches the target

## Why it breaks

`sim.calculate("aca_ptc")` computes PTC for **all** tax units. ~28% of tax units are non-filers. These are disproportionately lower-income households with high ACA PTC eligibility, contributing ~$466M that was never constrained. The calibration can't fix this because it only sees filer `aca_ptc` in the constraint matrix.

## The legal argument

The ACA PTC can only be claimed by filing Form 8962 with a federal tax return ([26 USC § 36B](https://www.law.cornell.edu/uscode/text/26/36B)). Non-filers cannot receive it. This isn't a modeling choice — it's how the credit works.

## Proposed fix

Gate on `tax_unit_is_filer` in `formula_2024` (file: `policyengine_us/variables/gov/aca/ptc/aca_ptc.py`):

```diff
     def formula_2024(tax_unit, period, parameters):
         plan_cost = tax_unit("slcsp", period)
         income = tax_unit("aca_magi", period)
         applicable_figure = tax_unit("aca_required_contribution_percentage", period)
         takes_up_aca_if_eligible = tax_unit("takes_up_aca_if_eligible", period)
+        is_filer = tax_unit("tax_unit_is_filer", period)
         return (
-            max_(0, plan_cost - income * applicable_figure) * takes_up_aca_if_eligible
+            max_(0, plan_cost - income * applicable_figure) * takes_up_aca_if_eligible * is_filer
         )
```

The `tax_unit_is_filer` variable already exists and models filing behavior (required filers + credit-motivated filers + voluntary filers).

## Note on precedent

No other credit in policyengine-us currently gates on `tax_unit_is_filer`. This may be worth revisiting for other credits that require filing (EITC, refundable CTC), but `aca_ptc` is the most acute case due to the inverse correlation between PTC eligibility and filing propensity.

juaristi22:
## Summary
- Gates `aca_ptc.formula_2024` on `tax_unit_is_filer` so non-filers correctly receive $0 PTC
- Fixes ~5x overestimate in calibrated microsimulation datasets (non-filer tax units were computing PTC that was never constrained by IRS SOI calibration targets)
- ACA PTC legally requires filing Form 8962 with a federal return ([26 USC § 36B](https://www.law.cornell.edu/uscode/text/26/36B))

Closes #7748

## Test plan
- [x] Added test case verifying non-filers get $0 PTC
- [x] Updated existing tests to explicitly set `tax_unit_is_filer: true` so each test validates its intended logic path
- [x] All 5 YAML tests pass

### DIFF
--- a/changelog.d/fix-aca-ptc-non-filers.fixed.md
+++ b/changelog.d/fix-aca-ptc-non-filers.fixed.md
@@ -1,0 +1,1 @@
+Gate aca_ptc on tax_unit_is_filer to prevent non-filers from receiving PTC.
--- a/policyengine_us/tests/policy/baseline/gov/aca/ptc/aca_ptc.yaml
+++ b/policyengine_us/tests/policy/baseline/gov/aca/ptc/aca_ptc.yaml
@@ -17,6 +17,7 @@
     slcsp: 1_200
     aca_magi: 25_000
     aca_required_contribution_percentage: 0.04
+    tax_unit_is_filer: true
   output:
     aca_ptc: 200  # 1200 - 0.04 * 25000 = 200
 
@@ -28,6 +29,7 @@
     slcsp: 1_200
     aca_magi: 25_000
     aca_required_contribution_percentage: 0.05
+    tax_unit_is_filer: true
   output:
     aca_ptc: 0  # 1200 - 0.05 * 25000 = -50, floored to 0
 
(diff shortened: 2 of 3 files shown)


baogorek:
Oh I'm so stuck here. Concerned about this:
```
  This is a circularity bug: is_filer depends on aca_ptc (via eligible_for_refundable_credits), and aca_ptc depends on county. The constraint evaluates is_filer at state level where aca_ptc = 0, so is_filer = False. But the actual county-level aca_ptc > 0 would make is_filer = True.

```

PR #7821, opened 2026-03-20
Add at-interview health coverage rule inputs

### LINKED ISSUE #7816: Add effective health coverage variables separate from reported CPS coverage
Context

We need a clean separation between reported CPS health coverage and modeled policy truth in `policyengine-us`.

Related data-side work in `policyengine-us-data` will import current CPS `NOW_*` health coverage flags as independent person-level `reported_..._at_interview` booleans. Those inputs should remain raw survey facts.

Rules-side need

Add calculated variables that represent policy-relevant / effective current coverage without mutating the reported inputs. In particular, the rules engine should be able to represent cases where reported coverage conflicts with modeled eligibility or subsidy rules.

Suggested calculated layer

- `has_qualifying_non_marketplace_health_coverage_at_interview`
- `eligible_for_medicaid` / existing eligibility vars where relevant
- `eligible_for_marketplace_coverage`
- `eligible_for_aca_ptc`
- `effective_current_health_coverage_for_aca_rules`
- `coverage_report_model_conflict`

Example motivating case

If someone reports both Marketplace and Medicaid coverage at interview time, we should preserve both reported inputs. But the rules engine should still reflect policy truth, e.g. Marketplace premium tax credits should generally be blocked when the person has qualifying Medicaid coverage (subject to any limited-benefit exceptions we model).

Relationship to take-up

The desired behavior is to match the SSI/SNAP-style pattern used in `policyengine-us-data`, not to replace take-up logic with raw reported coverage.

That means:

- reported CPS current coverage variables are inputs/signals
- enrollment/take-up variables still need to move people around to match ACA / Medicaid enrollment targets
- effective/canonical rules-side coverage should reflect the post-assignment modeled state, not just the raw reported booleans

So the rules-side work here should coexist with the take-up machinery: preserve reported survey facts, use them as informative anchors, but continue to let calibrated take-up determine final modeled enrollment / program consequences.

Why this matters

- preserves raw survey information
- lets rules-side formulas reflect legal/policy constraints accurately
- supports explicit conflict detection instead of silent data mutation
- keeps ACA/PTC logic consistent when reported coverage combinations are unusual or impossible under modeled rules

A follow-up implementation may also want a small derived status bucket for UI/reporting, but the source of truth should remain multi-flag inputs plus calculated effective coverage, not a single enum.

daphnehanse11:
Closes #7816

## Summary
- add person-level `has_*_health_coverage_at_interview` input variables for CPS current-coverage data imported in `policyengine-us-data`
- add `has_qualifying_non_marketplace_health_coverage_at_interview`, `qualifying_non_marketplace_health_coverage_type_count_at_interview`, and `coverage_report_model_conflict` for ACA-facing reported-coverage reconciliation
- add `has_marketplace_health_coverage_at_interview` so Marketplace/non-Marketplace conflict checks use an explicit reported current-coverage input
- add Medicaid regression coverage for current-coverage inputs and preserve continuous-coverage behavior in the CA Medicaid immigration eligibility rule

## Testing
- `uv run policyengine-core test policyengine_us/tests/policy/baseline/gov/aca/eligibility/has_qualifying_non_marketplace_health_coverage_at_interview.yaml policyengine_us/tests/policy/baseline/gov/aca/eligibility/coverage_report_model_conflict.yaml policyengine_us/tests/policy/baseline/gov/hhs/medicaid/medicaid_enrolled.yaml -c policyengine_us`
- `make format`


policyengine_us/variables/gov/aca/eligibility/has_qualifying_non_marketplace_health_coverage_at_interview.py
```diff
@@ -0,0 +1,28 @@
+from policyengine_us.model_api import *
+
+
+class has_qualifying_non_marketplace_health_coverage_at_interview(Variable):
+    value_type = bool
+    entity = Person
+    label = "Person has qualifying non-Marketplace health coverage at interview"
+    definition_period = YEAR
+
+    def formula(person, period, parameters):
```
MaxGhenis, line 10: use adds

policyengine_us/variables/input/health_coverage.py
MaxGhenis, line 1: one variable per file

baogorek:
Nice work — the mixin pattern keeps the input definitions clean, and the test coverage looks solid.

Two things worth double-checking before merge:

1. **`has_marketplace_health_coverage` default flips from `True` to `False`.** Previously every person defaulted to having marketplace coverage, which fed into `is_aca_ptc_eligible`. Now it defaults to `False` (via the reported flag). That's more correct for microsim, but in the web app where users build individual households and don't set the reported flag, this could silently zero out ACA PTC for everyone. Worth confirming that pathway still works as expected.

2. **`receives_medicaid` was an explicit input variable** (docstring says it "anchors takeup calibration"). Adding a formula that delegates to the reported flag changes it from a user-settable input to a derived value. Same question: does direct assignment still take priority over the formula in contexts that depend on it?

policyengine_us/variables/input/health_coverage.py
```diff
@@ -0,0 +1,124 @@
+from policyengine_us.model_api import *
+
+
+class ReportedHealthCoverageAtInterview:
```
MaxGhenis: frame around reality not surveys. what insurance do they have at the moment?

policyengine_us/variables/input/health_coverage.py
```diff
@@ -0,0 +1,124 @@
...
+    label = "Reported private health coverage at interview"
+
+
+class reported_has_public_health_coverage_at_interview(
+    ReportedHealthCoverageAtInterview, Variable
+):
+    label = "Reported public health coverage at interview"
+
+
+class reported_is_insured_at_interview(ReportedHealthCoverageAtInterview, Variable):
```
MaxGhenis: this would be a formula right

policyengine_us/variables/input/health_coverage.py
```diff
@@ -0,0 +1,124 @@
...
+    label = "Reported direct-purchase health coverage at interview"
+
+
+class reported_has_marketplace_health_coverage_at_interview(
+    ReportedHealthCoverageAtInterview, Variable
+):
+    label = "Reported Marketplace health coverage at interview"
+
+
+class reported_has_subsidized_marketplace_health_coverage_at_interview(
```
MaxGhenis: isnt this a formula? marketplace + aca_ptc>0? but we should just remove it

MaxGhenis (approved):
Rechecked after the cleanup pass. The structural review comments are addressed and the targeted tests pass.

PR #7899, opened 2026-03-29
[Alt] Add TAXSIM compatibility variables in PE-US

MaxGhenis:
## Summary
- add namespaced `taxsim_*` compatibility variables in PE-US for ambiguous TAXSIM outputs
- keep the migration surface in PE-US rather than encoding those semantics primarily in policyengine-taxsim
- add focused tests for the compat layer and the OK/MN split-credit adapters

## Notes
- This is a competing alternative to PolicyEngine/policyengine-us#7898 and PolicyEngine/policyengine-taxsim#793.
- The matching TAXSIM PR prefers these PE-US compatibility vars and keeps an explicit fallback for older `policyengine-us` releases.

### DIFF
--- a/changelog.d/7897-taxsim-compat-layer.added.md
+++ b/changelog.d/7897-taxsim-compat-layer.added.md
@@ -1,0 +1,1 @@
+Add explicit `taxsim_*` compatibility variables in PolicyEngine US for TAXSIM migration and adapter outputs.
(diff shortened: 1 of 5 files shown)


MaxGhenis:
Closing this alternative in favor of #7898, which matches the repo split we chose and is already green. Keeping the  compat-layer approach here as a documented alternative if we revisit a migration-oriented PE-US compatibility surface later.

PR #7940, opened 2026-04-06
Fix DE Filing Status 4 per-column credit allocation

### LINKED ISSUE #7931: DE Filing Status 4: non-refundable credits must be allocated per-spouse, not pooled at unit level
## Summary

When Delaware married couples file combined separate (Filing Status 4), non-refundable credits must be allocated per-column (per-spouse) and applied against each spouse's individual tax. PE currently pools all credits at the tax-unit level and subtracts them from total tax, which overstates the credit benefit when one spouse has little or no tax liability.

For a joint elderly couple (ages 65/80) with $314 pwages, $20K swages, $1K interest, $41K pensions, 2 dependents:

| Approach | Credits Used | DE Tax |
|----------|-------------|--------|
| PE (pools unit-level) | $710 of $710 | **$199** |
| Optimal per-column | $550 of $660 | **$359** |
| TaxAct (suboptimal allocation) | $220 of $660 | **$689** |
| Joint filing | $710 | **$388** |

PE's $199 is too low because it pools credits that DE law requires to be per-column.

## Legal Basis

**PIT-RES Instructions, Page 5 (Filing Status 3 and 4):**
> "For Filing Status 3 or 4, you must each report your own income, **personal credits**, deductions, and one-half of the income derived from securities, bank accounts, real estate, etc."

Credits are per-column for Filing Status 4.

**Line 27a (Personal Credits):**
The form allows taxpayers to **choose** how many exemptions to allocate to each column: "On Line 27a, enter the number of exemptions for: Column A ___, Column B ___". Optimal allocation puts more exemptions in the column with higher tax.

**Line 27b (Aged Credits, Page 9):**
> "If you are filing a combined separate return (Filing Status 4), enter $110 in the column(s) that correspond to the checked box(es)."

Aged credits are locked to the person's column.

**EITC (Page 10):**
> "the credit may only be applied against the tax imposed on the spouse with the **higher taxable income** reported on Line 23."

## Root Cause

`de_income_tax_before_refundable_credits.py` computes:
```python
max(de_income_tax_before_non_refundable_credits_unit - de_non_refundable_credits, 0)
```

This subtracts total credits from total unit tax. For Filing Status 4, credits should be applied per-column with each column's credits capped at that column's tax.

## Suggested Fix

For FS4, implement per-column credit allocation:
1. Optimize personal exemption allocation between columns (assign to the column with higher tax)
2. Lock aged credits to each person's column
3. Apply EITC to the higher-income spouse's column
4. Cap each column's credits at that column's tax
5. Sum the per-column net tax

## Integration Test

```yaml
- name: taxsim_812_de_joint_elderly_2025_per_column_credits
  absolute_error_margin: 10
  period: 2025
  input:
    people:
      person1:
        age: 65
        employment_income: 314
        taxable_interest_income: 628
        taxable_private_pension_income: 20_840
        is_tax_unit_head: true
      person2:
        age: 80
        employment_income: 20_000
        taxable_interest_income: 628
        taxable_private_pension_income: 20_840
        is_tax_unit_spouse: true
      child1:
        age: 8
      child2:
        age: 15
    tax_units:
      tax_unit:
        members: [person1, person2, child1, child2]
        premium_tax_credit: 0
    families:
      family:
        members: [person1, person2, child1, child2]
    spm_units:
      spm_unit:
        members: [person1, person2, child1, child2]
        snap: 0
        tanf: 0
    households:
      household:
        members: [person1, person2, child1, child2]
        state_fips: 10
  output:
    de_income_tax: 359
```

Note: Expected value $359 assumes optimal per-column exemption allocation (all 4 personal exemptions to Column A/spouse who has $909 tax; aged credits locked per-person). TaxAct gives $689 because it suboptimally allocates 3 of 4 exemptions to Column B (primary, who has $0 tax).

## References

- Originating TAXSIM issue: PolicyEngine/policyengine-taxsim#812
- DE PIT-RES Instructions (2025): https://revenuefiles.delaware.gov/2025/PITForms_Instructions/Instructions/PIT-RES_Instructions_2025-01.pdf
  - Page 5: Filing Status 3/4 per-spouse credit requirement
  - Page 9: Line 27a/27b credit instructions
  - Page 10: EITC allocation to higher-income spouse
- DE PIT-RES Form (2025): https://revenuefiles.delaware.gov/2025/PITForms_Instructions/PIT-RES_2025-01_PaperInteractiveIPM.pdf
- `de_income_tax_before_refundable_credits.py` — pools credits at unit level instead of per-column

PavelMakarchuk:
## Summary
- For combined separate filing (FS4), non-refundable credits are now allocated per-column instead of pooled at the tax-unit level
- Personal credits (line 27a) are optimally allocated between columns
- Aged credits (line 27b) are locked to each person's column
- CDCC (line 28) and non-refundable EITC (line 30) go to the higher-income spouse's column
- Each column's credits are capped at that column's tax

Previously PE pooled all credits, overstating the credit benefit when one spouse has little or no tax liability (e.g., elderly couple with $314 primary wages — PE computed $199 tax vs correct $359).

## Test plan
- [ ] Existing DE integration tests pass (updated taxsimid 2682 expected value for per-column allocation)
- [ ] New FS4 elderly couple test (from issue #7931) passes with expected $359 tax
- [ ] Existing DE unit tests pass unchanged (joint/single path unchanged)

Closes #7931

### DIFF
--- a/changelog.d/de-per-column-credits.fixed.md
+++ b/changelog.d/de-per-column-credits.fixed.md
@@ -1,0 +1,1 @@
+Fix Delaware Filing Status 4 to allocate non-refundable credits per-column instead of pooling at the tax-unit level, and elect the filing status on post-credit liability so the joint path is chosen when combined separate filing would waste credits.
--- a/policyengine_us/reforms/states/de/dependent_credit/de_dependent_credit_reform.py
+++ b/policyengine_us/reforms/states/de/dependent_credit/de_dependent_credit_reform.py
@@ -53,7 +53,23 @@ def create_de_dependent_credit_reform() -> Reform:
             p = parameters(period).gov.contrib.states.de.dependent_credit
 
             filing_status = tax_unit("filing_status", period)
-            files_separately = tax_unit("de_files_separately", period)
+            # Select the AGI basis from the pre-credit filing comparison rather
+            # than de_files_separately. The baseline election now compares
+            # post-credit liabilities, which depend on the personal credit this
+            # reform overrides; reading it here would create a circular
+            # dependency. The pre-credit comparison reproduces the election this
+            # phase-out was designed against.
+            itax_indiv = add(
+                tax_unit,
+                period,
+                ["de_income_tax_before_non_refundable_credits_indv"],
+            )
+            itax_joint = add(
+                tax_unit,
+                period,
+                ["de_income_tax_before_non_refundable_credits_joint"],
+            )
+            files_separately = itax_indiv < itax_joint
             agi_indiv = add(tax_unit, period, ["de_agi_indiv"])
             agi_joint = add(tax_unit, period, ["de_agi_joint"])
             agi = where(files_separately, agi_indiv, agi_joint)
--- a/policyengine_us/reforms/states/de/dependent_credit/de_dependent_credit_reform.py
+++ b/policyengine_us/reforms/states/de/dependent_credit/de_dependent_credit_reform.py
@@ -145,6 +145,18 @@ def create_de_dependent_credit_reform() -> Reform:
             self.update_variable(de_dependent_credit_phaseout)
             self.update_variable(de_dependent_credit)
             self.update_variable(de_older_dependents_count)
+            # Known limitation (follow-up): the dependent credit is
+            # delivered by overriding the tax-unit de_personal_credit,
+            # which only the joint (single-column) path consumes. The
+            # combined-separate (Filing Status 4) path instead reads the
+            # per-column de_personal_credit_indv, which this reform does
+            # not override, so a couple electing FS4 does not yet receive
+            # the dependent credit on the separate path. The
+            # de_files_separately election therefore compares a joint
+            # liability that includes the credit against a separate one
+            # that does not. Delivering it on both paths would require
+            # also overriding de_personal_credit_indv to allocate the
+            # dependent credit per column.
             self.update_variable(de_personal_credit)
 
     return reform
--- a/policyengine_us/reforms/states/de/dependent_credit/de_dependent_credit_reform.py
+++ b/policyengine_us/reforms/states/de/dependent_credit/de_dependent_credit_reform.py
@@ -156,7 +156,11 @@ def create_de_dependent_credit_reform() -> Reform:
             # liability that includes the credit against a separate one
             # that does not. Delivering it on both paths would require
             # also overriding de_personal_credit_indv to allocate the
-            # dependent credit per column.
+            # dependent credit per column, for which the (hypothetical)
+            # credit has no Delaware per-column allocation rule. This
+            # limitation is pinned by the guard test "FS4 election forgoes
+            # the dependent credit" in
+            # tests/policy/contrib/states/de/dependent_credit_reform_test.yaml.
             self.update_variable(de_personal_credit)
 
     return reform
--- a/policyengine_us/tests/code_health/test_non_refundable_credit_downstream_consumers.py
+++ b/policyengine_us/tests/code_health/test_non_refundable_credit_downstream_consumers.py
@@ -8,9 +8,18 @@ REFORMS_ROOT = REPO / "reforms"
 
 
 REVIEWED_APPLIED_CREDIT_EXTERNAL_REFERENCES = {
+    "de_cdcc": {
+        # Delaware combined separate (FS4) routes the applied CDCC to the
+        # lower-income spouse's column (PIT-RES Line 31); it is re-capped at
+        # that column's tax in de_income_tax_after_non_refundable_credits_indv.
+        "variables/gov/states/de/tax/income/credits/cdcc/de_cdcc_indv.py",
+    },
     "de_non_refundable_eitc": {
         "variables/gov/states/de/tax/income/credits/eitc/de_eitc.py",
         "variables/gov/states/de/tax/income/credits/eitc/refundability_calculation/de_income_tax_if_claiming_non_refundable_eitc.py",
+        # Combined separate (FS4) path applies the EITC to the higher-income
+        # spouse's column (PIT-RES Line 34), capped at that column's tax.
+        "variables/gov/states/de/tax/income/de_income_tax_before_refundable_credits_separate.py",
     },
     "ky_personal_tax_credits": {
         "variables/gov/states/ky/tax/income/credits/family_size_credit/ky_family_size_tax_credit_potential.py",
--- a/policyengine_us/tests/policy/baseline/gov/states/de/tax/income/credits/cdcc/de_cdcc_indv.yaml
+++ b/policyengine_us/tests/policy/baseline/gov/states/de/tax/income/credits/cdcc/de_cdcc_indv.yaml
@@ -1,0 +1,108 @@
+- name: CDCC locked to lower-income spouse (head has higher income)
+  period: 2025
+  input:
+    people:
+      person1:
+        is_tax_unit_head: true
+        de_taxable_income_indv: 50_000
+      person2:
+        is_tax_unit_spouse: true
+        de_taxable_income_indv: 10_000
+    tax_units:
+      tax_unit:
+        members: [person1, person2]
+        de_cdcc: 200
+    households:
+      household:
+        members: [person1, person2]
+        state_code: DE
+  output:
+    de_cdcc_indv: [0, 200]
+
+- name: CDCC locked to lower-income spouse (spouse has higher income)
+  period: 2025
+  input:
+    people:
+      person1:
+        is_tax_unit_head: true
+        de_taxable_income_indv: 5_000
+      person2:
+        is_tax_unit_spouse: true
+        de_taxable_income_indv: 60_000
+    tax_units:
+      tax_unit:
+        members: [person1, person2]
+        de_cdcc: 200
+    households:
+      household:
+        members: [person1, person2]
+        state_code: DE
+  output:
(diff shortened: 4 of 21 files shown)


hua7450 (changes requested):
I think there are two correctness issues here:

P1: `de_personal_credit_indv` still implements line 27a as an all-or-nothing allocation. The comment acknowledges that PIT-RES allows splitting the personal credit between Columns A and B in $110 increments, but the formula routes the full amount to whichever column has more remaining capacity. That can overstate liability whenever both columns can absorb part of the credit.

P2: `de_income_tax_before_refundable_credits` still relies on `de_files_separately`, but `de_files_separately` is computed from pre-credit tax only. After this refactor, FS4 can waste non-refundable credits on a per-column basis, so the optimal filing choice can change after credits are applied. I do not see a later recomputation of that choice, which means the model can now select FS4 even when the joint path produces the lower final liability.

MaxGhenis:
Please rebase from upstream/main (there are conflicts) and address the requested changes. CI and review will proceed once those are cleared.

hua7450:
**@PavelMakarchuk** — heads-up on a contrib-reform interaction this PR introduces. Not blocking (contrib-only, baseline + CI unaffected), but it needs your call.

**The gap:** `de_dependent_credit_reform` delivers the dependent credit by overriding the tax-unit `de_personal_credit`, which only the **joint** path consumes (`de_income_tax_before_refundable_credits_joint` → `de_non_refundable_credits` → `de_personal_credit`). The new **combined-separate (FS4)** path instead reads the per-column `de_personal_credit_indv`, which the reform does **not** override. So under this reform:
- an FS4 couple electing *separate* receives **$0 dependent credit** on that path, and
- the `de_files_separately` election compares a joint liability *with* the credit against a separate liability *without* it (apples-to-oranges).

**Test gap:** no test exercises the `gov.contrib.states.de.dependent_credit` reform on the FS4-separate path, so neither the dropped credit nor the biased election is currently guarded.

**Your call:**
1. Also override `de_personal_credit_indv` to allocate the dependent credit per column (Columns A/B), and add a contrib test covering FS4-separate; or
2. Accept the limitation as-is — I've already documented it inline in `de_dependent_credit_reform.py` (in the reform's `apply()`), and a guard test could pin the current behavior.

hua7450 (dismissed):
Requesting changes so this doesn't merge before the contrib-reform interaction is resolved (decision, not a baseline bug).

**Open item:** the `de_dependent_credit` reform overrides the tax-unit `de_personal_credit` (joint path) but not the new per-column `de_personal_credit_indv` (FS4 combined-separate path), so under the reform an FS4 couple electing *separate* receives $0 dependent credit and the `de_files_separately` election compares a joint liability *with* the credit against a separate one *without* it. No contrib test covers the FS4-separate path. Details and options in https://github.com/PolicyEngine/policyengine-us/pull/7940#issuecomment-4836206558.

Either (1) override `de_personal_credit_indv` per column + add a contrib FS4-separate test, or (2) accept the documented limitation with a guard test.

Note: the earlier critical finding — a non-functional test (`absolute_error_margin: 2` on the boolean `de_files_separately` output) — has already been fixed and pushed in this branch, along with a reference dedup and a stray `unit = USD` removal.

PavelMakarchuk:
Resolved the contrib-reform interaction via **option 2 (accept the documented limitation with a guard test)**, and rebased on main.

**Why option 2 over overriding `de_personal_credit_indv`:** the dependent credit is a *hypothetical contrib* credit, so there is no Delaware rule for how it would be allocated across the FS4 columns (the per-column personal credit splits in $110 exemption increments; a phased-out dependent credit doesn't map onto that). Inventing a per-column split would be *less* faithful than transparently not delivering it. The practical effect is a small, conservative understatement confined to a narrow edge case: because the reform lowers the joint-path liability by the credit, any couple for whom the credit is decisive is pushed onto the joint path that *does* deliver it; only couples whose FS4 bracket savings already exceed the credit elect FS4 and forgo it.

**Guard test added** (`dependent_credit_reform_test.yaml` → "FS4 election forgoes the dependent credit"): a two-earner couple ($80k/$70k, one child, $110 credit, no phase-out) elects FS4 and its realized liability equals the separate path:
- `de_files_separately`: true
- joint path (with credit): 8,124.50 · separate path (without credit): 7,108 · **elected: 7,108**
- `de_dependent_credit`: 110, `de_personal_credit`: 330 (reaches only the joint path)

If the credit were ever delivered on the separate path, the elected liability would drop to 6,998 and this test fails — so the limitation can't silently change. The reform's code comment now points to this test. All 7 reform-test cases pass; the earlier fixes (non-functional `de_files_separately` margin, reference dedup, stray `unit = USD`) remain in.

Re-requesting review.

DTrim99:
Closing #8726 (my DE Filing-Status-4 per-spouse-cap attempt) in favor of this PR — the per-column structure here is the correct approach, and @PavelMakarchuk's review of #8726 is what convinced me. Since the two edit overlapping FS4 logic and can't both merge, this is the one to land.

For anyone cross-referencing: I checked this PR at `db9dffd` against the regulatory routing Pavel laid out in the #8726 review, and it already covers all of it:
- **CDCC → lower-income column** — column-tied `de_cdcc_indv` ✅
- **Non-refundable EITC → higher-income column** — pinned by the `de_income_tax_before_refundable_credits` FS4 EITC cases ("EITC to higher-income spouse", "…capped at higher-income column tax", "…waste", "tied → route to head") ✅
- **Line-27a personal credit split in $110 increments, with spillover** — the optimal split in `de_personal_credit_indv` (resolves hua7450's **P1**) ✅
- **Post-credit FS4-vs-joint election** — `de_income_tax_before_refundable_credits` selects on the post-credit amounts (resolves **P2**) ✅
- **Head/spouse gating** — `is_tax_unit_head_or_spouse`, so per-column capacity doesn't range over dependents ✅

So P1/P2 both look resolved as of `db9dffd`, and the `CHANGES_REQUESTED` appears to predate those commits — a re-review would likely clear it.

One additive suggestion, aligned with the #8726 analysis: if it isn't already covered by the existing FS4 cases, consider a regression test for the exact **taxsimid 2682** scenario Pavel worked through — head (lower) pre-credit tax $6.05, spouse (higher) $3,225.65, CDCC $252.50. Because the CDCC is tied to the head and his $6.05 is already exhausted by his own personal/aged credits, the $252.50 is wasted → separate ≈ $2,895.65 > joint $2,705.30, so the couple elects **joint** and `de_income_tax = 2,705.30`. That's the precise case the highest-capacity approach in #8726 got wrong, so pinning it would guard the column routing and the post-credit election together.

MaxGhenis:
@hua7450 — following up on your requested-changes review (the contrib-reform interaction, [comment 4836206558](https://github.com/PolicyEngine/policyengine-us/pull/7940#issuecomment-4836206558)). Verified at HEAD `db9dffd`: this is resolved via **option 2** (the documented limitation + guard test) that you offered, and I confirmed it holds by re-running the tests and independently reproducing the guard scenario.

**What's in place**
- `de_dependent_credit_reform.apply()` overrides the tax-unit `de_personal_credit` only (joint path), and carries an inline comment explaining why it does *not* override the per-column `de_personal_credit_indv`, pointing to the guard test.
- Guard test "FS4 election forgoes the dependent credit" (`tests/policy/contrib/states/de/dependent_credit_reform_test.yaml`) pins the behavior. I reproduced its scenario ($80k/$70k two-earner couple, one child, $110 credit, no phase-out) end-to-end:
  - `de_files_separately`: true · `de_dependent_credit`: 110 · `de_personal_credit`: 330
  - joint (with credit): 8,124.50 · separate (without credit): 7,108 · **elected: 7,108**
  - If the credit ever reached the separate path, the elected liability would drop to 6,998 and the test fails — so the limitation can't silently change.

**Why option 2 rather than overriding `de_personal_credit_indv`**
The dependent credit is a hypothetical contrib credit with no Delaware rule for allocating it across the FS4 columns. The per-column personal credit splits in $110 exemption increments (PIT-RES Line 27a); a phased-out dependent credit doesn't map onto that, so a per-column split would have to be invented and would be *less* faithful than transparently not delivering it. The effect is conservative and narrow: because the reform lowers the joint-path liability by the credit, any couple for whom the credit is decisive is pushed onto the joint path that delivers it; only couples whose FS4 bracket savings already exceed the credit (here 1,016.50 ≫ 110) elect FS4 and forgo it.

**Test status** (local, this branch): DE baseline directory `tests/policy/baseline/gov/states/de` — **317 passed**; contrib reform `dependent_credit_reform_test.yaml` — **7 passed** (incl. the FS4 guard). All PR CI checks are green; `mergeable: MERGEABLE`, blocked only on this stale `CHANGES_REQUESTED`.

The per-column baseline restructure also addresses your original P1 (optimal $110-increment split in `de_personal_credit_indv`) and P2 (post-credit FS4-vs-joint election in `de_income_tax_before_refundable_credits`), consistent with @DTrim99's read at `db9dffd`. Related issues #7931/#8710 (baseline per-spouse credit capping) are the motivation for that restructure and out of scope for this reform-interaction item. If option 2 looks good to you, a re-review would clear the block.

MaxGhenis (approved):
Verified the reform-interaction resolution matches the reviewing teammate's offered option 2: the contrib reform documents why it does not override the per-column FS4 path (no DE statutory basis for a per-column split of a phased-out credit), with a guard test locking the FS4-forgoes-credit behavior. DE baseline 317 tests + 7 contrib tests pass; guard scenario reproduced independently. CI green.

PR #7943, opened 2026-04-07
[codex] Recompute future SPM thresholds

MaxGhenis:
## What changed
- stop treating `spm_unit_spm_threshold` as a generic uprated input in future years
- add a shared SPM reference-threshold helper and rebuild future thresholds from prior-year implied geography, current tenure, and current SPM-unit composition
- add regression tests covering composition changes, tenure changes, and explicit future-year overrides

## Why
SPM thresholds should not just be carried forward by CPI on the total stored amount. That loses the structure we built this for: the reference threshold, the equivalence scale, and the unit-specific geographic adjustment.

## Impact
Future-year SPM thresholds in `policyengine-us` now respond to household composition and tenure changes while preserving the implied geographic adjustment from the stored base-year threshold.

## Root cause
`spm_unit_spm_threshold` was defined as an input variable with `gov.bls.cpi.cpi_u` uprating, so future periods inherited a simple scalar uprate of the total threshold instead of recomputing the threshold components.

## Validation
- `PYTHONPATH=. pytest -q policyengine_us/tests/policy/baseline/household/income/spm_unit`

### DIFF
--- a/policyengine_us/tests/policy/baseline/household/income/spm_unit/test_spm_unit_spm_threshold.py
+++ b/policyengine_us/tests/policy/baseline/household/income/spm_unit/test_spm_unit_spm_threshold.py
@@ -1,0 +1,136 @@
+from __future__ import annotations
+
+import pytest
+
+from policyengine_us import CountryTaxBenefitSystem, Simulation
+
+SYSTEM = CountryTaxBenefitSystem()
+CPI_U = SYSTEM.parameters.gov.bls.cpi.cpi_u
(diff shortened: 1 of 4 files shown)


MaxGhenis:
Closing as superseded by the merged SPM threshold/geography series: #8020 migrated thresholds to spm-calculator, #8249 computes SPM geographic adjustments from geography, and #8250 fixed the projected-year childcare cap follow-up. The old draft is now conflicting and its remaining future-threshold behavior is covered by the current formula-owned implementation.

PR #7971, opened 2026-04-10
[codex] Model Medicare Part B premiums net of MSP coverage

MaxGhenis:
## What changed
- model baseline `medicare_part_b_premiums` from enrollment and IRMAA/base premium schedules instead of treating Part B as a copied input
- keep the reported CPS value as `medicare_part_b_premiums_reported` for audit use
- net modeled Part B out-of-pocket premiums by a cycle-free MSP standard-premium offset
- avoid the prior filing-status cycle in IRMAA calculations by using joint/separate status directly

## Why
Part B premiums feed baseline SPM medical out-of-pocket expenses and therefore poverty projections. The prior baseline path copied reported/imputed Part B values, which is especially weak on the PUF/clone side. This PR moves the baseline to a model-consistent Part B path and nets out standard premiums covered on the beneficiary's behalf through MSP-like support.

## Impact
- makes baseline SPM MOOP more coherent for projections
- preserves only IRMAA above the MSP-covered standard premium as out-of-pocket
- keeps a separate reported variable for Census-style audits

## Validation
- `uv run pytest -q policyengine_us/tests/policy/baseline/gov/hhs/medicare/test_part_b_msp_offset.py`
- direct local rebuild/measurement showed the MSP offset materially lowers modeled Part B totals but only modestly moves enhanced SPM poverty

Closes #7967.

### DIFF
--- a/changelog_entry.yaml
+++ b/changelog_entry.yaml
@@ -1,0 +1,5 @@
+- bump: patch
+  changes:
+    added:
+    - Add Census SPM weekly work-expense parameters and formulas for work and childcare expense caps.
+  date: 2026-04-09 00:00:00
--- a/policyengine_us/parameters/gov/census/index.yaml
+++ b/policyengine_us/parameters/gov/census/index.yaml
@@ -1,0 +1,4 @@
+metadata:
+  propagate_metadata_to_children: true
+  economy: false
+  household: false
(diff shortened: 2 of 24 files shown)


MaxGhenis:
Superseded by clean replacement PR #7972 from codex/medicare-partb-msp-clean.

PR #8035, opened 2026-04-17
Support Python 3.9 and 3.10

MaxGhenis:
## Summary

Lowers `requires-python` to `>=3.9,<3.15` so users on older Python versions (e.g., Census) can install `policyengine-us`.

- **Source fixes**: five function signatures used PEP 604 `X | Y` union annotations that fail at `def`-execution time on 3.9; rewritten as `Optional[X]` / `Optional[List[X]]` to match the package's existing style.
- **pyproject.toml**: add 3.9/3.10 classifiers, lower `requires-python` floor. The `policyengine-core>=3.23.5` pin is unchanged — pip will naturally resolve to a 3.9-compatible core version (3.24.0+) on older Python once that release ships.
- **CI**: add a new `Python-Compat` matrix job that installs the package and runs a smoke-import across 3.9–3.14.

## Dependency chain

Full 3.9 support across the PolicyEngine stack requires:

1. **PolicyEngine/policyengine-core#454** — relaxes `requires-python` + numpy env markers. Must merge first.
2. **This PR** — source + classifier fixes.
3. PolicyEngine/policyengine-uk — parallel PR (coming).
4. PolicyEngine/policyengine.py — parallel PR (coming).

Until (1) publishes core 3.24.x, the new `Python-Compat` CI matrix will fail on 3.9/3.10 because no compatible core exists on PyPI. The existing 3.14-pinned jobs (which use `uv.lock`) are unaffected.

## Test plan

- [x] All five edited files compile under Python 3.9 (`py_compile`)
- [x] Rewritten `Optional[...]` signatures evaluate at def-time on 3.9 (verified with isolated repro)
- [x] `ruff format --check` clean
- [ ] CI: new `Python-Compat` job passes on 3.11, 3.12, 3.13, 3.14 (will pass 3.9/3.10 once core 3.24.x ships)
- [ ] Existing full-suite jobs still green


MaxGhenis:
The `Install + smoke-import (py3.9)` failure is expected and documented in the PR description — it requires a policyengine-core release that ships the relaxed Python floor. The core-side change (`PolicyEngine/policyengine-core#454`) merged a few minutes ago, so once a 3.24.x release is published to PyPI and pip can resolve a 3.9-compatible core, this job will pass without further changes here.

All other CI (3.10–3.14 smoke imports + full test suite on 3.14) is green.

MaxGhenis:
core 3.24.0 is now on PyPI — re-running CI to pick up 3.9/3.10 smoke-imports.

PR #8069, opened 2026-04-18
Bump policyengine-core floor to 3.24.4 (closes #8066)

### LINKED ISSUE #8066: Core 3.24.x Enum cast regression blocks core floor bump
## Summary

`policyengine-us` can't cleanly bump its `policyengine-core` floor past 3.24.0 because core regressions break the `tax_unit_itemizes.yaml` integration test (and many others via the same cycle) with a `TypeError` during formula result casting.

Tracking the core fix here so the bump can land once core releases it.

## Symptom under core 3.24.1

```
File ".../policyengine_core/simulations/simulation.py", line 1067, in _cast_formula_result
    return value.astype(variable.dtype)
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
```

Surfaces in:

- `policyengine_us/tests/policy/baseline/gov/irs/income/taxable_income/deductions/tax_unit_itemizes.yaml` — "Integration test which fails without tolerance in the itemization choice" and "Oklahoma itemization test requiring higher tolerance"
- any household simulation whose `tax_unit_itemizes` path crosses the `itemizing` reform branch (same class of test as #8059)

## Root cause (in core, not us)

Two regressions in core 3.24.0 combine to break us:

1. **H3 `_invalidate_all_caches` over-reaches.** `Simulation.apply_reform` wipes every variable's `_memory_storage._arrays`, including user-provided `set_input` values. `policyengine-us`'s `Simulation.__init__` applies a structural reform AFTER the situation is built, so every household input (`state_fips`, `age`, `premium_tax_credit`, …) gets wiped between init and the first `calculate`. The `TypeError` above is the downstream consequence: `state_fips` returns `None`, `state_name.formula` maps it through `pd.Series.map({...})` → `NaN`, and the `.astype(int64)` cast blows up.

2. **C1 `Holder.get_array` fallback is too narrow.** After fixing #1, the `tax_unit_itemizes` cycle surfaces (same stack as #8059). `tax_liability_if_itemizing` sets `tax_unit_itemizes=True` on an `itemizing` branch; `ctc_limiting_tax_liability` forks a `no_salt` sub-branch from it. Under 3.24.x, `Holder.get_array` on the `no_salt` branch only falls back to `default`, not to the parent `itemizing` branch, so `tax_unit_itemizes` re-runs its formula on `no_salt` → calls `tax_liability_if_itemizing` again → infinite recursion.

Plus a latent `GroupPopulation.clone` bug that only surfaces once #2 is fixed: the cloned holder's `.simulation` reference was pointing at the *source* sim, not the clone, so `parent_branch` walks started from the wrong node.

## Core fix

PolicyEngine/policyengine-core#475 covers all three:

1. Track user `set_input` provenance in `Simulation._user_input_keys` (both via `Simulation.set_input` and `Holder.set_input` — the latter covers the `Simulation(situation=...)` path through `SimulationBuilder.finalize_variables_init`) and replay on `_invalidate_all_caches`.
2. Walk `simulation.parent_branch` in `Holder.get_array` before falling back to `default`.
3. `GroupPopulation.clone` now passes the cloned population to `holder.clone`.

With that branch applied to core 3.24.1, the failing us tests pass; without it, they crash.

## Plan

- [ ] PolicyEngine/policyengine-core#475 merges
- [ ] New core release (3.24.3 or 3.25.0)
- [ ] Bump `policyengine-core>=<new-version>` in this repo
- [ ] Confirm `tax_unit_itemizes.yaml` and the full IRS test suite pass on the bump

MaxGhenis:
## Summary

Bump `policyengine-core` floor to **3.24.4** and drop the temporary Python-version restriction on `uv.lock`. Closes #8066.

## Why

`policyengine-core 3.24.0` added a strict validator that errors on `breakdown` / parameter-children mismatches — the exact class of bug that hit Martin Holmer on us 1.644.0 and 1.645.0 (issue #8055) and that three partial-fix PRs had to chase: #8045, #8049, #8051.

But bumping the core floor used to be blocked by three coupled regressions inside core 3.24.0–3.24.3:

1. `Simulation.apply_reform` wiped user-provided dataset inputs (the H3 cache invalidation hit `_memory_storage._arrays` for user inputs as well as formula-output caches).
2. `Holder.get_array` fell back only to `default`, missing the parent-branch walk needed by country-package nested branches (e.g. `no_salt` under `itemizing`).
3. `GroupPopulation.clone` passed the source population (not the clone) to `holder.clone`, so group-entity holders on cloned branches pointed at the wrong simulation.

Together these caused the `tax_unit_itemizes.yaml` integration test to crash with `TypeError: int() argument ... not 'NoneType'` (state_fips wiped) plus an infinite recursion in `tax_liability_if_itemizing`.

PolicyEngine/policyengine-core#475 fixed all three, shipped in core 3.24.4 (latest at time of writing: 3.25.0).

## Effect

- **`test_system_import.py`** (added in #8058) now actually does work: constructing `CountryTaxBenefitSystem()` runs the strict validator, so any future breakdown/children mismatch fails CI rather than reaching end users at import time.
- **PyPI installs** no longer silently pull a core version with the partial-fix regressions in 3.24.0–3.24.3.
- The `[tool.uv]` `environments = ["python_version >= '3.11'"]` restriction can be removed — core 3.24+ supports 3.9 and 3.10 on PyPI.

## Verified locally

- 794/794 tests pass under `policyengine_us/tests/policy/baseline/gov/irs/` with core 3.25.0.
- `tax_unit_itemizes.yaml` 7/7 pass (was 2 failing on core 3.24.1).
- `test_system_import.py` passes as expected.

## Test plan

- [x] IRS test suite: 794 passed
- [ ] Full CI passes

### DIFF
--- a/changelog.d/bump-core-floor-3244.changed.md
+++ b/changelog.d/bump-core-floor-3244.changed.md
@@ -1,0 +1,1 @@
+Bump policyengine-core floor to 3.24.4 so the breakdown/children validator runs in CI and PyPI installs, closing the remaining gap from issue #8055. Remove the [tool.uv] Python-version restriction since core 3.24+ supports 3.9/3.10.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ classifiers = [
 dependencies = [
     "microdf-python>=1.0.0",
     "pandas>=2.0",
-    "policyengine-core>=3.23.5",
+    "policyengine-core>=3.24.4",
     "spm-calculator>=0.2.0",
     "tables>=3.9",
     "tqdm>=4.67.1",
@@ -50,15 +50,6 @@ dev = [
 [tool.hatch.build.targets.wheel]
 packages = ["policyengine_us"]
 
-[tool.uv]
-# Until policyengine-core ships 3.9/3.10 support (3.24.0+), restrict
-# uv.lock resolution to >=3.11. `requires-python` above still declares
-# 3.9 is supported — pip users on 3.9/3.10 will resolve against PyPI
-# once core 3.24+ is published. Remove this block once the lockfile
-# can cleanly resolve across 3.9-3.14.
-environments = [
-    "python_version >= '3.11'",
-]
 
 [tool.towncrier]
 package = "policyengine_us"
--- a/uv.lock
+++ b/uv.lock
@@ -11,9 +11,8 @@ resolution-markers = [
     "python_full_version == '3.11.*' and sys_platform == 'emscripten'",
     "python_full_version >= '3.12' and python_full_version < '3.14' and sys_platform != 'emscripten' and sys_platform != 'win32'",
     "python_full_version == '3.11.*' and sys_platform != 'emscripten' and sys_platform != 'win32'",
-]
-supported-markers = [
-    "python_full_version >= '3.11'",
+    "python_full_version == '3.10.*'",
+    "python_full_version < '3.10'",
 ]
 
 [[package]]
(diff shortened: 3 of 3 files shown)


MaxGhenis:
Superseded by a rebased PR (the original branch accumulated 3 rounds of GitHub runner preemptions / test cancellations; rebasing onto current main gives a clean CI run).

PR #8070, opened 2026-04-18
Bump policyengine-core to >=3.25.0

MaxGhenis:
## Summary
Bumps \`policyengine-core\` minimum to \`>=3.25.0\` to pick up the \`set_input\` preservation fix from PolicyEngine/policyengine-core#475.

## Why
PE-core 3.24.0–3.24.3 introduced a cache-invalidation cascade that wiped \`set_input\` values across \`apply_reform\`. In policyengine-us this manifested as \`state_fips\` being cleared whenever a structural reform was applied during simulation construction, and the downstream \`state_name\`/\`state_code\` chain returning \`None\`. Closes #8058.

3.25.0 preserves user-provided inputs (including dataset loads via \`holder.set_input\`) while still invalidating formula-output caches.

## Test plan
- [x] \`test_ctc_itemizing_branch_cycle\` passes
- [ ] CI verifies full microsim tests

Generated with [Claude Code](https://claude.com/claude-code)

PR #8077, opened 2026-04-18
Bump policyengine-core to >=3.25.1

MaxGhenis:
## Summary
Bumps `policyengine-core` minimum to `>=3.25.1` to pick up:

1. The `set_input` preservation fix (PolicyEngine/policyengine-core#475) — closes #8058 where `apply_reform` wiped `state_fips` and broke the `state_name`/`state_code` chain.
2. The `_invalidate_all_caches` performance fix (PolicyEngine/policyengine-core#478) — the initial 3.25.0 version walked every variable in the tax-benefit system and lazy-created a Holder for each, which blew up the YAML full-suite from ~17 min to ~51 min per job and started hitting the 1-hour GitHub Actions timeout. 3.25.1 iterates only existing holders.

Supersedes the closed #8070 which bumped to 3.25.0 and hit the perf regression.

## Test plan
- [x] `tests/test_ctc_itemizing_branch_cycle` passes locally
- [ ] CI passes including Full Suite back at normal duration (<20 min per job)

Generated with [Claude Code](https://claude.com/claude-code)


MaxGhenis:
Multiple reruns have all hit GitHub Actions runner preemptions (`The runner has received a shutdown signal`) on the Full Suite jobs — same infrastructure flakiness flagged in #8069 / #8078. The one Full Suite job that did complete (Structural (States)) passes in 36.5 min, consistent with main's recent runs.

Key points for reviewer:
- This is a superset of #8078's 3.24.4 bump. 3.25.1 adds the `set_input` preservation fix (#8058) and the `_invalidate_all_caches` perf fix from PolicyEngine/policyengine-core#478 on top.
- Core + downstream unit tests pass locally (`tests/test_ctc_itemizing_branch_cycle` etc.).
- The Full Suite pattern of passing Structural (States) + infra-killed others matches every other PR right now, not this specific bump.

Leaving this open in case someone wants to retry when the runner fleet stabilises, or close if the 3.24.4 baseline from #8078 is good enough for now.

PR #8177, opened 2026-04-28
Cap CHIP premiums at five percent of income

### LINKED ISSUE #8101: Enforce 42 CFR 457.560 5% family income cap on chip_premium
Follow-up from #8086 (found in session review).

## Problem

Federal law (42 CFR 457.560) caps **total CHIP cost sharing** at 5% of the family's income for the 12-month enrollment period. "Cost sharing" in this regulation combines:

- Premiums and enrollment fees
- Copays, deductibles, coinsurance
- Any other charges imposed on CHIP enrollees

PolicyEngine's `chip_premium` (from #8086) is computed from state schedules without this cap. For high-income CHIP-eligible families in states with steep schedules — especially **WI** (no per-child cap, up to $97.53/child/month at the top bracket) and **MO** (up to $508/month for a large family of 8 at 225-300% FPL) — the computed premium alone can approach or exceed the 5% cap.

Example (MO family of 8 at 280% FPL, ~$80k income): encoded premium tier 3 = $508/mo = $6,096/yr. 5% cap = $4,000. State would reduce premium to $4,000. We currently return $6,096.

## Fix (first-pass)

Apply the 5% cap on `chip_premium` alone:

```python
capped_premium = min(computed_state_premium, 0.05 * tax_unit_income)
```

in each `<state>_chip_premium` variable, or as a post-computation cap in the federal `chip_premium` aggregator.

## Known imperfection

Because the real cap binds on premium + copays + other cost-sharing, capping premium alone:

- Is **too lenient** on the cap: if copays are $600 and premium computed at $3,000 for a family with $50k income, real combined = $3,600 > $2,500 cap → state would reduce. We return the uncapped $3,000 because we don't see the copays. Real collect = $1,900 (cap-adjusted). Difference: $1,100.
- Is **never too harsh**: premium-only capping never binds harder than the true combined cap would.

Net effect: under-applies the cap. Better than not applying it at all, since the few cases where premium alone exceeds 5% are real.

## Full fix requires copay modeling

Proper enforcement of 42 CFR 457.560 needs `chip_copay` or similar — rules-based or imputed — added to the cost-sharing sum. Each state has its own copay schedule by service type (per office visit, per ER, per prescription, etc.). Modeling utilization is hard; a simplified state-average-per-enrollee imputation would be a first pass.

Separate follow-up — more research effort than the cap alone.

## Acceptance

- 42 CFR 457.560 cap enforced on `chip_premium` (not combined with copays).
- Test: MO family of 8 at 280% FPL, $80k income: computed premium capped at $4,000 (5% of $80k), not $6,096.
- Test: WI with many children at top FPL tier shows capping behavior.
- Variable docstring calls out the copay-absent under-enforcement explicitly.

## Related

- #8086 — CHIP premium framework (where the uncapped computation lives)
- #8089 — MOOP architecture

daphnehanse11:
## Summary

- Enforces the federal 42 CFR 457.560 CHIP cost-sharing cap on modeled `chip_premium`.
- Adds a federal CHIP cumulative cost-sharing cap-rate parameter set to 5 percent of family income.
- Adds MO and WI regression tests covering premium-only cap behavior, while preserving existing aggregation tests.

## Notes

This first pass caps modeled premiums and enrollment fees only. The variable documentation explicitly notes that CHIP copays, deductibles, coinsurance, and similar charges are not yet modeled, so the combined cost-sharing cap can still be under-enforced.

Fixes #8101.

## Tests

- `uv run pytest policyengine_us/tests/test_parameter_files.py -q`
- `uv run python -m policyengine_core.scripts.policyengine_command test policyengine_us/tests/policy/baseline/gov/hhs/chip/chip_premium.yaml policyengine_us/tests/policy/baseline/household/income/spm_unit/spm_unit_medical_out_of_pocket_expenses.yaml policyengine_us/tests/policy/baseline/household/income/spm_unit/spm_unit_spm_expenses_chip_premium.yaml -c policyengine_us`


daphnehanse11:
Codex review pass before creating the fix worktree:

- I checked the PR diff against #8101's acceptance criteria. The implementation applies the cap at the federal `chip_premium` aggregator, adds the 5% parameter, covers the requested MO/WI cap cases, and documents the known under-enforcement from unmodeled copays.
- I did not find a diff-level blocker in this pass.
- One status issue: the latest GitHub Actions run has several cancelled jobs, so the PR does not yet have a complete green CI signal. I am creating a fresh worktree now to rerun the focused parameter and CHIP premium tests and will patch anything that reproduces locally.

daphnehanse11:
Code-quality pass pushed in 7ca1044c6f. I renamed the CHIP cap base from family_income to annual_family_income so the formula is more specific about the period/unit used for the 5 percent cap. Focused CHIP cap/SPM tests pass locally: 11 passed.

MaxGhenis (approved):
Focused review and subagent review found no blocking issues; CI is green.

PR #8185, opened 2026-04-29
Improve WIC package valuation

MaxGhenis:
## Summary
- assign federal WIC food packages before valuing benefits, including infant feeding variants, child age splits, breastfeeding multiples, and package-specific pregnancy/breastfeeding rules
- add current CVB parameters and subtract the embedded 2018 CVB before adding FY 2024/FY 2026 amounts
- use WIC countable income while preserving the existing `is_pregnant` WIC FPG adjustment used by API customers

## Tests
- `uv run policyengine-core test policyengine_us/tests/policy/baseline/gov/usda/wic`
- `uv run ruff check policyengine_us/variables/gov/usda/wic`


daphnehanse11 (changes requested):
Thanks for the WIC package cleanup. I found one correctness issue that should be fixed before merge:

`policyengine_us/variables/gov/usda/wic/wic_food_package.py:69`

`monthly_age` does not return age in months in this codebase; it returns the true age in years for monthly periods. Because these new branches compare that value to `4` and `6`, every real infant under age one satisfies `age_months < 4`, so the 4-5 month and 6-11 month package branches are unreachable unless a test or caller manually overrides `monthly_age`. This misassigns normal API inputs: a six-month formula-fed infant with `age: 0.5` currently gets `INFANT_0_3_FULLY_FORMULA_FED` instead of `INFANT_6_11_FULLY_FORMULA_FED`, even though 7 CFR 246.10 separates Food Package I for birth through 5 months from Food Package II for 6 through 11 months.

Please convert the age to months before these comparisons, or introduce/use a true age-in-months variable, and add tests that derive infant package selection from `age` rather than setting `monthly_age` directly.

Verification:

- `uv run policyengine-core test policyengine_us/tests/policy/baseline/gov/usda/wic` passed (`60 passed`)
- `uv run --extra dev ruff check policyengine_us/variables/gov/usda/wic` passed
- A repro test with `age: 0.5` and `wic_infant_feeding_category: FULLY_FORMULA_FED` failed with `INFANT_0_3_FULLY_FORMULA_FED` instead of `INFANT_6_11_FULLY_FORMULA_FED`

daphnehanse11 (approved):
Fixed the WIC infant package age-threshold issue in f2ae7e92b7 by deriving infant package thresholds from annual age converted to months, and added age-based regression coverage. Verified with the full WIC policy suite and the WIC ruff check.

PR #8219, opened 2026-05-01
Add refundable credit conversion contrib reform

DTrim99:
## Summary

Adds a new contrib reform `gov.contrib.refundable_credit_conversion` that introduces a configurable flat refundable credit composed of independently toggleable components. The credit is appended to the federal refundable credit list when `in_effect` is true.

Repealing baseline credits and deductions (CTC, EITC, CDCC, standard deduction, itemized deductions, above-the-line deductions, Head of Household filing) is intentionally **out of scope** for this PR and is handled separately via direct overrides to baseline parameters.

## Implementation

**Parameters under `gov.contrib.refundable_credit_conversion/`:**
- `in_effect.yaml` — bool gate for the reform.
- `credit/{per_taxpayer, per_ctc_dependent, per_other_dependent, per_household}.yaml` — currency-USD component amounts, default 0.
- `credit/{use_taxpayer_credit, use_ctc_dependent_credit, use_other_dependent_credit, use_household_credit}.yaml` — bool toggles, default false.
- `earnings_credit/{rate, cap, use_earnings_credit}.yaml` — `/1` rate, currency-USD per-worker cap, and bool toggle for the earnings subsidy. "Earnings" follows the EITC convention of wages plus self-employment income.

**Reform Python at `policyengine_us/reforms/refundable_credit_conversion/refundable_credit_conversion.py`:**
- Defines `flat_refundable_credit` (TaxUnit, USD) as `taxpayers * per_taxpayer + ctc_kids * per_ctc_dependent + max(eitc_kids - ctc_kids, 0) * per_other_dependent + per_household + sum_over_persons(min(max(earnings, 0), cap) * rate)`. Each component is gated by its `use_*` toggle.
- Overrides `income_tax_refundable_credits` to append `flat_refundable_credit` to the existing federal refundable list (baseline EITC / refundable CTC / CDCC / AOTC / PTC / etc. continue to compute and contribute).
- `create_refundable_credit_conversion_reform` registers via the streamlined-EITC / ECPA pattern with a 5-year `in_effect` forward lookahead. The module-level alias uses `bypass=True` to make the bypass semantics explicit.

**Tests at `policyengine_us/tests/policy/contrib/refundable_credit_conversion/`:**
- `flat_credit.yaml` — credit math for single, MFJ, and MFJ-with-dependents shapes.
- `dependent_definition.yaml` — boolean-toggle behavior for CTC-qualifying vs other-dependent children.
- `earnings_credit.yaml` — per-worker cap arithmetic, SE income inclusion, boundary cases (zero, exactly at cap, negative SE), and dependents with earnings.
- `integration.yaml` — full reform end-to-end including an assertion on `income_tax_refundable_credits` (using a minimal household to avoid CI OOM), `in_effect: false` case, all four filing-status branches, and an all-toggles-on / all-amounts-zero sanity case.
- `test_refundable_credit_conversion_autoloader.py` — exercises the autoloader's `bypass=True` path and the default-inactive return-`None` path.

**Test isolation:** `test_batched.py` isolates `refundable_credit_conversion` into its own contrib batch because the `income_tax_refundable_credits` override pulls every federal refundable credit into the dependency graph and was OOM'ing the catch-all CI runner.

## Test plan

- [x] `make format`
- [x] All 23 CI checks pass
- [ ] `policyengine-core test policyengine_us/tests/policy/contrib/refundable_credit_conversion/ -c policyengine_us`
- [ ] `pytest policyengine_us/tests/microsimulation/test_microsim.py` (regression)
- [ ] Verify state EITC / state CTC / state standard deduction values are unchanged when the federal reform is active


DTrim99:
## Program Review

Automated `/review-program` pass over the diff: 1 reform-logic reviewer, 1 reference-quality reviewer, 1 code-pattern validator, 1 test-coverage analyzer. No PDF source applies (this is a contrib reform, not a parameter update from an official document), so PDF audit was skipped.

### Source Documents
- **PDF**: N/A — contrib reform, no external source document
- **Scope**: PR changes only

### Summary

The implementation is internally consistent and the math checks out — flat-credit formula, dependent-counting set difference, wage credit aggregation, 5-year lookahead, state neutrality, and variable reuse are all correct. CI is green across all 23 checks.

The single user-facing issue is that the **PR description on GitHub still describes the original (broader) scope** while the code has been deliberately narrowed to ship only the new credit. The reform's own docstring and the changelog fragment are accurate — but the GitHub description will mislead reviewers.

### Critical (Must Fix)

1. **PR description ↔ implementation drift.** The PR body still claims the reform "swaps the federal standard deduction, itemized deductions, above-the-line deductions, Head of Household filing status, CTC, CDCC, and EITC for a flat refundable credit" with `repeals/*.yaml` files, a `dependent_definition.yaml` int enum, a `repeals.yaml` test, and overrides for `standard_deduction` / `itemized_taxable_income_deductions` / `above_the_line_deductions` / `head_of_household_eligible` / `income_tax_non_refundable_credits`. **None of those are in the diff.** Recent commits (`fc51846591` "Narrow refundable credit conversion contrib to the new credit only", `30490b71` "Replace dependent_definition int with per-component boolean toggles", `f1909275bd` "Drop income_tax_refundable_credits assertion from RCC integration test") confirm the narrowing was intentional, and the reform docstring (`policyengine_us/reforms/refundable_credit_conversion/refundable_credit_conversion.py:25-29`) and changelog correctly say repeals are handled separately. Update the PR body to match the narrowed scope so reviewers and downstream consumers aren't surprised.

2. **The `income_tax_refundable_credits` override has no end-to-end assertion.** Commit `f1909275bd` removed the only test that exercised the override (for CI-OOM reasons) and nothing was added back. Every test asserts `flat_refundable_credit` directly — none verifies the splice into the federal refundable-credit list. If `update_variable(income_tax_refundable_credits)` silently broke (e.g., a typo in `"flat_refundable_credit"`), all existing tests would still pass. Add one targeted test: construct a household where every baseline refundable credit is zero (no earnings, no qualifying children, no AOTC/PTC eligibility) and assert `income_tax_refundable_credits == flat_refundable_credit == <expected>` for a chosen scenario. Or use the weaker `income_tax_refundable_credits >= flat_refundable_credit` invariant.

### Should Address

3. **`documentation` field on `flat_refundable_credit`.** Per `policyengine-variable-patterns`: use `reference` instead of `documentation`. `policyengine_us/reforms/refundable_credit_conversion/refundable_credit_conversion.py:37-41`. Either drop the field (the reform docstring already covers it) or replace with a `reference` link to PR #8219.

4. **"Wage credit" semantics vs. behavior.** The parameter labels and reform docstring call this a "wage" credit, but `earnings = person("earned_income", period)` (line 87) includes self-employment income (`earned_income` adds `employment_income + self_employment_income + sstb_self_employment_income`). Either rename to "earnings credit" / "earnings subsidy" to match behavior, or restrict the formula to `employment_income` to match the "wage" framing.

5. **5-year `in_effect` lookahead is untested.** No test exercises `create_refundable_credit_conversion_reform(parameters, period)` with `in_effect` true 3-4 years out or false through the whole window. YAML tests force-apply the reform via the `reforms:` directive, completely bypassing the lookahead. Add at least one boundary-case test (true within window, false beyond window).

6. **`integration.yaml` second case doesn't test what its name suggests.** It sets `in_effect: false` but explicitly passes `reforms: ...refundable_credit_conversion`, so the reform applies anyway — the test merely verifies the formula returns 0 because every `use_*` toggle defaults to false. It does not test the autoloader bypass path.

7. **Parameter labels lack a program prefix.** All amount/toggle parameters use generic labels like `"Flat credit amount per taxpayer"` and `"Apply per-CTC-dependent flat credit"` — these are not identifiable in the parameter explorer. Prefix with `"Refundable credit conversion: …"` to match the `in_effect.yaml` label (which already follows that convention). Compare AWRA's `"American Worker Tax Rebate amount (non-joint)"`.

8. **Filing-status coverage gap.** Only SINGLE and JOINT are tested. The reform's per-taxpayer count and dependent counts depend on filing status. Add cases for HEAD_OF_HOUSEHOLD, MFS, and (ideally) SURVIVING_SPOUSE.

9. **Wage credit boundary cases are thin.** Only one positive scenario (one earner under cap, one over). Missing: zero earnings, earnings exactly at the cap, negative earnings (the `max_(earnings, 0)` guard is never exercised), and dependents with earnings (current code awards them a wage credit — pin the behavior with a test even if it's intentional).

10. **Missing `reference` field on all 12 parameters.** Per `policyengine-parameter-patterns` the field is required. Contrib reforms with no external proposal sometimes omit it (cf. `gov/contrib/ctc/additional_bracket/in_effect.yaml`, `gov/contrib/joint_eitc/in_effect.yaml`), but adding a one-line link to PR #8219 — at minimum on `in_effect.yaml` as the entry point — costs nothing and improves traceability.

### Suggestions

11. **Module-level export should use `bypass=True`.** `policyengine_us/reforms/refundable_credit_conversion/refundable_credit_conversion.py:148` does `refundable_credit_conversion = create_refundable_credit_conversion()` (bypassing the lookahead implicitly). Compare `streamlined_eitc.py:69`: `streamlined_eitc = create_streamlined_eitc_reform(None, None, bypass=True)`. Match the established pattern so intent is explicit.

12. **Test number formatting.** `wage_credit.yaml` lines 7, 8, 29, 30 use `0.20` for the rate; the parameter-patterns convention drops trailing zeros (`0.2`).

13. **Hoist the duplicated `ctc_qualifying_children` lookup.** `refundable_credit_conversion.py:61` and `:74` both call `tax_unit("ctc_qualifying_children", period)` in different `if` branches — could be computed once. Purely cosmetic.

14. **Other-dependent count is a set-cardinality difference.** `max_(eitc_count - ctc_count, 0)` is correct in the typical case (CTC ⊆ EITC) but silently undercounts when the two sets are not nested (e.g., one CTC-qualifying child fails EITC ID requirements AND one EITC-qualifying child fails CTC). The inline comment acknowledges this. Worth a one-line note in `per_other_dependent.yaml` description that this is a population-aggregate difference, not a per-person classification.

15. **`absolute_error_margin` on wage_credit tests.** The wage credit produces a float (`0.2 * 8000 = 1600.0`); a future parameter change yielding `1599.99999` would silently fail. Add `absolute_error_margin: 0.01` for resilience.

16. **`in_effect.yaml` starts at `0000-01-01: false` rather than a real year.** Other reforms (e.g., `streamlined_eitc.in_effect`) use a real start year like `2020-01-01`. Either is fine, but a real year makes the parameter timeline less confusing in the UI.

17. **Toggle file naming has a redundant `_credit` suffix.** Files in `credit/` are named `use_taxpayer_credit.yaml`, `use_ctc_dependent_credit.yaml`, etc., yielding paths like `gov.contrib.refundable_credit_conversion.credit.use_taxpayer_credit` (double `credit`). Could be `credit/use_taxpayer.yaml` to mirror the amount file naming (`per_taxpayer.yaml`). Don't churn the diff if you'd rather keep the current names.

### Validation Summary

| Check | Result |
|-------|--------|
| Regulatory accuracy (reform-logic) | 1 critical (PR body drift), math/lookahead/state-neutrality all correct |
| Reference quality | 12/12 parameters have valid core metadata; `reference` field missing on all 12 (contrib-acceptable) |
| Code patterns | No hard-coded values; reuses `is_tax_unit_head_or_spouse`, `ctc_qualifying_children`, `eitc_child_count`, `earned_income`; mirrors ECPA / streamlined-EITC pattern correctly; 3 style/naming nits |
| Test coverage | 12 unit tests covering all 5 component toggles + `in_effect` gate; 1 critical gap (no end-to-end assertion on `income_tax_refundable_credits`), 4 should-address gaps (5-year lookahead, filing-status, wage boundaries, autoloader path) |
| CI status | All 23 checks passing |
| Changelog | Present at `changelog.d/refundable-credit-conversion.added.md`, accurately scoped |
| Test batching | `test_batched.py` correctly isolates `refundable_credit_conversion` into its own contrib batch |

### Review Severity: COMMENT

The implementation is correct and CI is green. The critical items are (a) updating the PR description to match the narrowed scope, and (b) adding one assertion that proves the `income_tax_refundable_credits` override actually fires. Neither blocks merge mechanically — but skipping them ships a misleading PR body and a structurally untested override.

### Next Steps
To auto-fix issues: `/fix-pr 8219`

PavelMakarchuk (commented):
## Program Review — PR #8219 (Contrib Reform: refundable_credit_conversion)

### Source Documents
None — contrib reform (configurable policy design primitive). No PDF audit applicable.

### Critical (Must Fix)
None.

### Should Address

1. **`in_effect.yaml` start date — `2024-01-01: false` is inconsistent with sibling reforms.** Pre-2024 callers rely on policyengine-core's implicit fallback. Sibling contrib reforms (`harris/lift/middle_class_tax_credit`, `reconciliation/ssn_for_llc_and_aoc`) use `0000-01-01: false`. Align with the dominant convention.

2. **PR description and module docstring overstate the baseline refundable list.** The docstring claims the override appends to "EITC / refundable CTC / CDCC / AOTC / PTC", but `gov/irs/credits/refundable.yaml` for periods >= 2022 does NOT include CDCC (was refundable only in 2021) or PTC (explicitly excluded by comment). The implementation is correct (it pulls dynamically from the parameter); only the docs are wrong. Either drop CDCC/PTC from the enumeration or use a generic phrase referencing `gov.irs.credits.refundable`.

3. **Consider a parametric reform on `gov.irs.credits.refundable` instead of overriding `income_tax_refundable_credits`.** Overriding the variable forces every federal refundable credit into the subprocess's variable graph, which is the proximate cause of the OOM-driven test isolation in `test_batched.py`. A parametric reform appending `flat_refundable_credit` to the parameter list would preserve the `adds` optimization and eliminate the batched-test workaround entirely. Not blocking — current approach works.

4. **`per_other_dependent.yaml` description is two sentences and uses no approved verb** (parameter-patterns skill expects one sentence with a verb from a fixed list). Same loose style applies across most parameter descriptions in this PR. Acceptable under contrib's relaxed conventions but worth tightening for consistency.

5. **Regulatory/explanatory comment is duplicated between the Python formula (lines 65-72) and the YAML description for `per_other_dependent`.** Pick one home (typically code comment) to avoid double maintenance when the explanation evolves.

6. **Per-worker cap with ALL workers above cap is not isolated in any test.** Current earnings_credit test has one worker under the cap and one over — passes for both correct `sum(min(earnings, cap))` and the typical bug `min(sum(earnings), cap)`. A test with two workers both well above the cap is the single most important regression-guard for the per-person cap semantics.

7. **`max_(eitc_kids - ctc_kids, 0)` clamp at line 75 is never exercised.** The guard exists for the rare configuration where a CTC-qualifying child fails EITC identification requirements (i.e., ctc_count > eitc_count). All current tests have eitc_count >= ctc_count. Add a case using `meets_eitc_identification_requirements: false` (or equivalent input) to drive the clamp.

8. **`in_effect: false` test does not verify the override behavior on `income_tax_refundable_credits`.** The case asserts only `flat_refundable_credit: 0`, which is also true when `in_effect: true` with all amounts zero. It does not distinguish "credit is zero" from "credit excluded from the refundable list". The autoloader's `None` return is exercised in Python tests, but the YAML `reforms:` directive force-applies the reform — so the zero value here actually comes from each `use_*` toggle defaulting to false, not from `in_effect: false`. Either tighten the assertion (compare `income_tax_refundable_credits` to baseline) or add a comment explaining the limited scope.

9. **Household credit (`use_household_credit`) is only tested in combination, never in isolation.** Other four toggles each have an isolation test. A regression that doubled the household credit or made it per-person could be partially masked by the combined cases. Adding an isolation case is cheap symmetry.

### Suggestions

1. **Earnings exactly at cap + 1.** Adds an off-by-one guard around the cap boundary.
2. **Combined W-2 + SE income on a single worker.** Verifies sum-then-cap ordering across the two income components (`earned_income` includes both plus SSTB).
3. **CTC age boundary 16 vs 17.** Pin down "under 17" cutoff explicitly; existing tests use 5 and 17 but never 16.
4. **Autoloader 5-year lookahead window untested.** The `for i in range(5)` lookahead in the autoloader has no direct test. Add cases at boundaries (`in_effect: true` in year+4 vs year+6).
5. **`bypass=True` with non-`None` parameters.** Current bypass test passes `None, None`; pin down that bypass truly ignores a parameters tree whose `in_effect` is false.
6. **Microsimulation smoke test marked `@pytest.mark.microsimulation`.** Per testing-patterns skill — the OOM concern in integration.yaml indicates an awareness of microsim risk; a marked test would catch regressions deliberately.
7. **Multi-tax-unit household.** Confirms credit aggregates per tax unit, not per household. Likely a non-issue given variable scope, but cheap symmetry.
8. **Split dense set-difference comment in formula** (lines 65-72) into a one-line summary plus a moved long-form note (paired with the duplication item).
9. **Style consistency — `taxpayer_amount = 0` scalars vs arrays.** When toggles are off, locals are Python scalars; broadcasting handles the final sum correctly. Cosmetic.
10. **Confirm `SURVIVING_SPOUSE` FilingStatus enum spelling** (visual check shows it's correct; flag is just defensive).

### PDF Audit Summary
Not applicable — contrib reform, no statute to encode.

### Validation Summary
| Check | Result |
|-------|--------|
| Regulatory Accuracy | N/A (contrib design primitive) |
| Reference Quality | N/A (no source documents) |
| Code Patterns | 0 critical + 5 should + 10 suggestions |
| Test Coverage | 0 critical + 8 should + 4 suggestions |
| PDF Value Audit | N/A |
| CI Status | 23/23 passing |

### Review Severity: APPROVE

Reason: This is a configurable contrib design primitive with no statutory mismatch to chase. Code patterns are clean (no hard-coded values, reuses existing variables like `earned_income` / `ctc_qualifying_children` / `eitc_child_count`, correctly appends rather than replaces in the override, sign-flip guard for negative SE income, autoloader pattern matches existing precedent). All 23 CI checks pass. Should-address items are polish — effective-date alignment, docs/description style, missing edge-case tests (cap behavior, eitc<ctc clamp, in_effect=false override-behavior assertion).

The architectural choice between variable override (current) and parametric reform on `gov.irs.credits.refundable` is worth a discussion — the override approach is what forced the `test_batched.py` isolation, and a parametric refactor would remove that wart. Not blocking.

### Next Steps
Optional follow-up to address SHOULD items.

DTrim99:
Addressed all of @PavelMakarchuk's review items. Summary:

**Should Address (9 items)**
- ✅ `in_effect.yaml` start date → `0000-01-01: false` (aligns with sibling contrib reforms)
- ✅ Docstring no longer enumerates CDCC/PTC — CDCC is refundable only in 2021 and PTC is explicitly excluded
- ❌ Parametric reform on `gov.irs.credits.refundable` — **tried, reverted**. Implemented `modify_parameters` snapshotting and re-emitting each transition with `flat_refundable_credit` appended. Empirically the per-test parameter-tree mutation is heavier than the variable override: CI showed shutdown-signal OOM *even inside the isolated `refundable_credit_conversion` batch*. Reverted to the variable override; kept the batch isolation in `test_batched.py`. Suggestion is sound in principle but doesn't pan out here.
- ✅ `per_other_dependent.yaml` collapsed to one sentence with the "Provides" verb
- ✅ Duplicated set-difference comment removed from YAML; lives in the Python formula only
- ✅ Both-workers-above-cap test case added to `earnings_credit.yaml` (regression guard for `sum(min(..., cap))` vs `min(sum(...), cap)` ordering)
- ✅ Eitc-clamp test using `meets_eitc_identification_requirements: false` added to `dependent_definition.yaml`
- ✅ `in_effect: false` integration test comment clarified — explicitly notes the YAML directive bypasses the autoloader
- ✅ Household-credit isolation test added to `flat_credit.yaml`

**Suggestions (10 items)**
- ✅ Earnings one dollar above cap boundary case
- ✅ Combined W-2 + SE income on one worker (sum-then-cap)
- ✅ CTC age 16 vs 17 boundary tests
- ✅ 5-year lookahead window tests (window start, year+4 inside, year+6 past) added to Python autoloader test
- ✅ `bypass=True` with non-`None` parameters test added
- ⏭ Microsim smoke test — skipped (requires dataset access)
- ⏭ Multi-tax-unit household — skipped (Pavel noted it's low-value symmetry)
- ✅ Long set-difference comment moved out of YAML
- ⏭ Scalar vs array style — cosmetic only
- ⏭ SURVIVING_SPOUSE enum spelling — defensive only

Commits: `e354e20050` (Pavel fixes), `a5f5ad62de` (restore isolation after parametric refactor OOM'd), `26a458caf8` (revert parametric refactor).

PavelMakarchuk (approved):

PR #8251, opened 2026-05-08
Fix KY family size tax credit at 1.33 boundary

DTrim99:
## Summary

KY's Schedule ITC publishes Family Size Tax Credit thresholds in whole dollars and rounds 0.5-cent fractions **up** to the next whole dollar. PE's `family_size/rate.yaml` encoded the breakpoints as exactly 1.27 and 1.33, leaving a sub-dollar window where MGI lands between the share threshold and KY's published cutoff — resulting in PE returning a 0% rate instead of 10% (or 20% instead of 30%) at the boundary.

For 2025 family of 4 (FPG = \$32,150):
- KY publishes "10% credit if MGI not over \$42,760"
- \$42,760 / \$32,150 = **1.330016**, not 1.33
- A household with true MGI = \$42,759.88 (e.g. issue [#852](https://github.com/PolicyEngine/policyengine-taxsim/issues/852)) gets share = 1.330012 in PE's float32, which exceeds the 1.33 bracket → 0% (TaxAct: 10%, \$91 credit)

Fix bumps the two affected thresholds to 1.2701 and 1.3301. The other brackets (1.04, 1.08, 1.12, 1.16, 1.20, 1.24, 1.30) land on whole dollars for all KY family sizes and are unchanged.

## Test plan
- [x] Existing 4 KY FSTC rate tests still pass
- [x] 3 new boundary tests (just below cutoff with cents, exactly at \$42,760, clearly above) pass
- [x] End-to-end check on issue 852 inputs: PE now returns FSTC = \$91.04 and KY tax = \$819.36, matching TaxAct (\$91 / \$819)
- [ ] CI green

Reported in PolicyEngine/policyengine-taxsim#852.


PavelMakarchuk (approved):
Approved.

Verified against the statute (KRS 141.066(3)(c)) and the 2025 Schedule ITC form:

- **Statute** uses strict percentage brackets (e.g., "10% if 130% < MGI/FPG ≤ 133%"). By that reading, MGI = $42,759.88, FPG = $32,150 gives ratio 1.33001 → strictly above 133% → 0% bracket.
- **Schedule ITC** publishes whole-dollar cutoffs: family-of-4 10% bracket is "is over $41,795 - is not over $42,760". KY arrives at $42,760 by rounding `1.33 × $32,150 = $42,759.50` up to a whole dollar. So MGI = $42,759.88 ≤ $42,760 → 10%.

The form is therefore strictly more generous than the statute in a sub-dollar window. This PR aligns PE with the form (which is what KY administers and what TaxAct follows). That's the right modeling choice.

Math also verified: thresholds `1.270032` and `1.330032` sit inside the safe interval `(1.27003195, 1.27009585)` and `(1.33003195, 1.33009585)` respectively (computed from the tightest family-of-1 case). Float32 preserves the comparison. Future-proof since `0.5 / FPG_min` shrinks as FPG grows.

One suggestion for a follow-up (not blocking): add a one-line note in the parameter YAML or changelog explicitly calling out the statute-vs-form tension. Something like *"this aligns PE with the Schedule ITC table KY publishes (whole-dollar cutoffs rounded up from `1.33 × FPG`); the strict statute is slightly less generous."* That single sentence converts the "magic number" question into a clear policy decision for future readers.

The failing `codecov/project` check is a coverage threshold — not relevant for a parameter-only change.

PR #8255, opened 2026-05-11
Model Medicare enrollment as eligible take-up

MaxGhenis:
## Summary
- Model `medicare_enrolled` as an eligibility-gated take-up variable, matching the Medicaid enrollment pattern.
- Add `takes_up_medicare_if_eligible`, defaulting to true for eligible people.
- Update Medicare cost, Part D surcharge, and ACA coverage-conflict fixtures to use take-up rather than directly forcing `medicare_enrolled`.

## Root Cause
Previously, `medicare_enrolled` had `defined_for = "is_medicare_eligible"` and `default_value = True`. When a dataset omitted `medicare_enrolled`, PolicyEngine could treat enrollment as a direct default input instead of a calculated eligibility-and-take-up result, which made downstream datasets vulnerable to overcounting enrollment.

## Validation
- `git diff --check`
- `uv run python -m py_compile policyengine_us/variables/gov/hhs/medicare/takes_up_medicare_if_eligible.py`
- `uv run python -m policyengine_core.scripts.policyengine_command test policyengine_us/tests/policy/baseline/gov/hhs/medicare policyengine_us/tests/policy/baseline/gov/aca/eligibility -c policyengine_us` (`171 passed, 1 warning`)

### DIFF
--- a/changelog.d/8255.fixed.md
+++ b/changelog.d/8255.fixed.md
@@ -1,0 +1,1 @@
+Fix Medicare enrollment defaults by modeling enrollment as eligibility-gated take-up.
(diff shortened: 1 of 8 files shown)


daphnehanse11 (approved):
codex and i say looks good!

PR #865, opened 2022-06-02
Massachusetts State income tax

nikhilwoodruff:
Fixes #845

PR #8284, opened 2026-05-13
Fix NJ property tax deduction at filing threshold (strict > should be >=)

### LINKED ISSUE #8278: NJ property tax deduction off-by-one at filing threshold (strict > should be >=)
## Summary

`nj_property_tax_deduction_eligible.py` line 16 uses strict `>` to gate the NJ property tax deduction on the NJ filing threshold. A joint filer with NJ AGI exactly equal to the filing threshold (\$20,000) is marked ineligible even though they file an NJ-1040 and qualify for the deduction.

## Affected file

[`policyengine_us/variables/gov/states/nj/tax/income/property_tax/nj_property_tax_deduction_eligible.py`](https://github.com/PolicyEngine/policyengine-us/blob/3225f95deb462b7e7fb9db4e827205ca21efc70d/policyengine_us/variables/gov/states/nj/tax/income/property_tax/nj_property_tax_deduction_eligible.py#L11-L21)

```python
    p = parameters(period).gov.states.nj.tax.income
    filing_status = tax_unit("filing_status", period)
    nj_agi = tax_unit("nj_agi", period)
    income_eligible = nj_agi > p.filing_threshold[filing_status]  # <-- here
    ...
```

## Concrete case (NJ joint, TY 2025)

Inputs: joint, both age 40, 2 dependents, spouse wages \$20,000, gross SS \$24,925, \$30,000 property tax, \$20,000 mortgage interest.

NJ AGI = \$20,000 (NJ excludes SS entirely; spouse wages are the only NJ-taxable income). Joint filing threshold = \$20,000.

| Variable | PE | Expected |
|---|---|---|
| `nj_property_tax_deduction_eligible` | False | True |
| `nj_property_tax_deduction` | \$0 | \$15,000 (capped) |
| `nj_taxable_income` | \$15,000 | \$0 |
| `nj_income_tax` (final, before refundable credits) | \$210 | \$0 |

## Suggested fix

Change line 16 to `>=` (or drop the threshold check entirely, since `nj_property_tax_deduction_eligible` is already gated on positive property-tax/rent payments — anyone with no income won't be in PE's NJ flow).

## Reference

Reported in PolicyEngine/policyengine-taxsim#865. Per the [2024 NJ-1040 instructions, page 25](https://www.state.nj.us/treasury/taxation/pdf/current/1040i.pdf), the deduction is available to any homeowner/tenant who files NJ-1040 — there's no separate income threshold for the deduction itself.

DTrim99:
## Summary

Strict `>` in `nj_property_tax_deduction_eligible` excluded households with NJ AGI exactly equal to the joint filing threshold (\$20,000) from the property tax deduction. The deduction is available to any taxpayer who files NJ-1040 (per the 2024 NJ-1040 instructions, page 25); the filing threshold is "must file if income exceeds threshold", and at exactly the threshold a household may still file voluntarily and qualify for the deduction.

Closes #8278. Reported in PolicyEngine/policyengine-taxsim#865.

## Changes

- `policyengine_us/variables/gov/states/nj/tax/income/property_tax/nj_property_tax_deduction_eligible.py` — change `>` to `>=` on the income eligibility check
- `policyengine_us/tests/.../nj_property_tax_deduction_eligible.yaml` — add 2 new boundary tests at exactly \$20,000 (joint) and \$10,000 (single)
- `changelog.d/fix-nj-property-tax-deduction-threshold.fixed.md` — changelog fragment

## Test plan
- [x] All 14 NJ property-tax-deduction tests pass (4 existing + 6 in deduction folder + 2 new boundary)
- [x] End-to-end verification: TAXSIM input `96640,2025,31,2,40,40,2,0,20000,...` now produces correct NJ property tax deduction = \$15,000 (capped) instead of \$0
- [ ] CI green

### DIFF
--- a/changelog.d/fix-nj-property-tax-deduction-threshold.fixed.md
+++ b/changelog.d/fix-nj-property-tax-deduction-threshold.fixed.md
@@ -1,0 +1,1 @@
+Fix New Jersey property tax deduction eligibility off-by-one at the filing threshold: change strict `>` to `>=` so households with NJ AGI exactly equal to the filing threshold qualify.
--- a/policyengine_us/tests/policy/baseline/gov/states/nj/tax/income/deductions/property_tax/nj_property_tax_deduction_eligible.yaml
+++ b/policyengine_us/tests/policy/baseline/gov/states/nj/tax/income/deductions/property_tax/nj_property_tax_deduction_eligible.yaml
@@ -45,3 +45,27 @@
     state_code: NJ
   output:
     nj_property_tax_deduction_eligible: true
+
+- name: 5 Income at exactly the joint filing threshold is eligible (boundary).
+  period: 2025
+  absolute_error_margin: 0
+  input:
+    filing_status: JOINT
+    nj_agi: 20_000
+    real_estate_taxes: 30_000
+    rents: false
+    state_code: NJ
+  output:
+    nj_property_tax_deduction_eligible: true
+
+- name: 6 Income at exactly the single filing threshold is eligible (boundary).
+  period: 2025
+  absolute_error_margin: 0
+  input:
+    filing_status: SINGLE
+    nj_agi: 10_000
+    real_estate_taxes: 30_000
+    rents: false
+    state_code: NJ
+  output:
+    nj_property_tax_deduction_eligible: true
--- a/policyengine_us/variables/gov/states/nj/tax/income/property_tax/nj_property_tax_deduction_eligible.py
+++ b/policyengine_us/variables/gov/states/nj/tax/income/property_tax/nj_property_tax_deduction_eligible.py
@@ -13,7 +13,7 @@ class nj_property_tax_deduction_eligible(Variable):
 
         filing_status = tax_unit("filing_status", period)
         nj_agi = tax_unit("nj_agi", period)
-        income_eligible = nj_agi > p.filing_threshold[filing_status]
+        income_eligible = nj_agi >= p.filing_threshold[filing_status]
 
         pays_ptax = add(tax_unit, period, ["real_estate_taxes"]) > 0
         pays_rent = tax_unit("rents", period)


PavelMakarchuk:
After digging into the official 2025 NJ-1040 instructions and the underlying statute, I think this change is going in the wrong direction. Three layers of evidence, all pointing the same way.

## Page 27 is the dispositive citation

The PR cites NJ-1040 page 25 (medical expenses), but the **actual eligibility list** for the Property Tax Deduction is on page 27. Bullet 5 reads:

> *"Your income on line 29 is **more than the filing threshold amount** for your filing status (see page 3). If you (and/or your spouse if filing jointly) were 65 or older or blind or disabled on the last day of the tax year and you were not required to file a return, see 'Seniors or Blind/Disabled Persons Not Required to File Form NJ-1040' below."*

This is unambiguous: deduction eligibility *itself* requires income **more than** the threshold — strict `>`, same wording as the filing-mandate rule. A household at exactly $20K joint **does not qualify for the property tax deduction** under NJ-1040.

The pre-PR code (`nj_agi > filing_threshold`) is exactly right. The PR flips it to `>=`, which makes the model contradict the form's eligibility wording.

## Filing-requirement wording (page 5) reinforces this

> *"You are required to file a return if... gross income was **more than** the filing threshold: $10,000 / $20,000."*

"More than $X" is strict `>` — a household at exactly $20,000 joint is not required to file.

## Statute (NJSA 54A:3A-17) has no income floor and a high-income ceiling

The [statute](https://law.justia.com/codes/new-jersey/title-54a/section-54a-3a-17/) has an income *ceiling* (deduction capped at $5K above $150K AGI; eliminated above $250K for non-senior/non-disabled filers) — not a floor. The form's "more than the filing threshold" eligibility test sits *on top of* the statutory rules.

## The PR's rationale doesn't generalize

The PR says "at exactly the threshold a household may still file voluntarily and qualify for the deduction." That argument applies equally at $19,999, $15,000, or $5,000 — voluntary filing for refunds/credits is a thing at any income. The current change flips a single income point, leaving everyone $1 below the threshold ineligible despite identical reasoning. Either it should be a broader rewrite, or the boundary should stay where the form says.

## Cause of the apparent ambiguity

`filing_threshold.yaml` description currently says *"requires filing income taxes if income is this amount or higher"* — that's actually **wrong** vs the official "more than" wording. The PR matches the (incorrect) parameter description rather than the (correct) statute and form. Worth fixing the description in a separate change.

## Recommendation

Close without merging and instead:

1. Fix `filing_threshold.yaml` description to match the official "more than" wording (small PR).
2. Take a closer look at the TAXSIM #865 reproducer — if TaxAct is producing the deduction at exactly $20K joint, that's actually a TaxAct vs. NJ-form discrepancy rather than a PE bug, since the form says "more than the filing threshold". Worth surfacing on the taxsim ticket.

## Two unrelated but real gaps I noticed while reviewing

These are out of scope for this PR but worth tracking:

1. **Missing senior/blind/disabled exception**: Per page 27, if 65+/blind/disabled and income at or below the filing threshold, the taxpayer can still claim via Form NJ-1040-HW. Not modeled.
2. **Missing high-income cap**: NJSA 54A:3A-17 reduces the deduction to $5,000 above $150K AGI and eliminates it above $250K (for non-senior/non-disabled filers). Currently appears to allow the full $15K cap regardless of income.

Both are bigger drivers of TAXSIM divergence than the boundary case here.

PavelMakarchuk (changes requested):
not sure if this is the right fix

DTrim99:
Thanks for the thorough review — you're right on all counts. I cited NJ-1040 page 25 in the PR body, but that's the Medical Expenses Deduction section; the actual Property Tax Deduction eligibility list is on **page 27** and uses "more than the filing threshold" wording. Strict `>` is correct.

I matched the existing `filing_threshold.yaml` description ("this amount or higher") rather than the form/statute — a mistake on my part. Closing this PR.

The two gaps you flagged are the more material divergences: senior/blind/disabled exception via NJ-1040-HW, and the high-income cap from NJSA 54A:3A-17 ($5K above $150K, eliminated above $250K for non-senior filers). Happy to scope each as a separate, focused PR. I'll also open a small PR fixing the `filing_threshold.yaml` description to match the official "more than" wording.

PR #8668, opened 2026-06-17
Add Nebraska to CCDF program coverage metadata

PavelMakarchuk:
## Summary

Nebraska's Child Care Subsidy is implemented in the model (`ne_child_care_subsidy`, under `gov/states/ne/dhhs/child_care_subsidy`, with parameters) but was missing from the CCDF entry in `programs.yaml`. As a result it doesn't appear on the model coverage page (`/us/model/rules/coverage`), which is driven by this metadata via `api.policyengine.org/<country>/metadata` → `modelled_policies`.

This adds the NE `state_implementation` and includes NE in the `coverage` list. CCDF now lists **31 jurisdictions**.

```yaml
- state: NE
  status: complete
  name: Nebraska CCS
  full_name: Nebraska Child Care Subsidy
  variable: ne_child_care_subsidy
  parameter_prefix: gov.states.ne.dhhs.child_care_subsidy
```

## Status note for reviewer

I marked NE `complete` because it produces an end-to-end subsidy amount plus eligibility variables. If the maintainers consider the Nebraska model still partial (e.g. missing copay/payment-rate detail comparable to other `complete` states), change `status` to `in_progress` or `partial`.

## Context

This is the data-side fix for the stale CCDF coverage shown on the model page. Note the live page is also lagging because the deployed API runs `policyengine-us` 1.715.2 while `main` is at 1.730.1 — the other recently-added CCDF states (AL, FL, HI, IA, ID, IN, KS, KY) are already correct in `main` and will appear once the API is redeployed on a newer release.

## Verification
- `programs.yaml` parses; CCDF now has 31 state implementations and 31 coverage entries.
- `ne_child_care_subsidy` variable and `gov/states/ne/dhhs/child_care_subsidy` parameter directory confirmed present on `main`.
- No tests reference `programs.yaml` structure; change is metadata-only.

### DIFF
--- a/changelog.d/add-ne-ccdf-coverage.changed.md
+++ b/changelog.d/add-ne-ccdf-coverage.changed.md
@@ -1,0 +1,1 @@
+Consolidated Nebraska Child Care Subsidy under the CCDF program coverage metadata, removing the duplicate standalone program entry.
--- a/policyengine_us/programs.yaml
+++ b/policyengine_us/programs.yaml
@@ -460,7 +460,7 @@ programs:
     category: Benefits
     agency: HHS
     status: partial
-    coverage: AK, AR, AL, AZ, CA, CO, CT, DE, DC, FL, GA, HI, IA, ID, IL, IN, KS, KY, LA, MA, MD, ME, MI, MN, MO, MS, NC, NH, NJ, PA, RI, SC, TX, VA, VT, WA, WV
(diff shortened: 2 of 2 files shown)


MaxGhenis (approved):
Verified the consolidation matches the CONTRIBUTING.md convention (state implementations under the parent federal entry), the coverage and state_implementations lists agree as exact 38-state sets with no stub entries, and nothing else references the removed ne_childcare id. Checked clean against both past programs.yaml corruption signatures. CI green.

PR #8691, opened 2026-06-19
Add dedicated South Carolina dependent exemption parameter

DTrim99:
## What this does

South Carolina's dependent exemption (§12-6-1140(13)) had **no parameter of its own** — both `sc_dependent_exemption` and `sc_young_child_deduction` read the young-child deduction's amount. This adds proper, **separate parameters** for the two distinct statutory provisions, each in its own folder:

- `gov.states.sc.tax.income.deductions.dependent_exemption.amount` — §12-6-1140(13), a deduction for **every** dependent (read by `sc_dependent_exemption`).
- `gov.states.sc.tax.income.deductions.young_child.amount` — §12-6-1160, the **additional** deduction for dependents under 6 (read by `sc_young_child_deduction`).

§12-6-1160 sets its amount equal to the §12-6-1140 amount, so both parameters carry identical values — but keeping them separate lets each be modelled/adjusted independently (e.g. in the Child Poverty Impact Dashboard).

## No behavior change

Pure restructure — values are identical. A dependent under 6 still correctly receives **both** deductions (~$8,220 combined), which is real SC policy. The existing SC baseline tests pass unchanged:
`sc_young_child_deduction.yaml` (4) + `sc_dependent_exemption.yaml` (2) — **6 passed**.

### DIFF
--- a/changelog.d/sc-dependent-exemption-parameter.added.md
+++ b/changelog.d/sc-dependent-exemption-parameter.added.md
@@ -1,0 +1,1 @@
+Added a dedicated South Carolina dependent exemption parameter (Section 12-6-1140), with the young child deduction (Section 12-6-1160) set equal to it; no change to calculated values.
--- a/policyengine_us/parameters/gov/states/sc/tax/income/deductions/dependent_exemption/amount.yaml
+++ b/policyengine_us/parameters/gov/states/sc/tax/income/deductions/dependent_exemption/amount.yaml
@@ -1,4 +1,4 @@
-description: South Carolina provides a young child deduction of this amount.
+description: South Carolina provides a dependent exemption of this amount per dependent. The young child deduction (Section 12-6-1160) is set equal to this amount.
 values:
   2019-01-01: 4_100 # Value in the legal code, last updated in 2019.
   2021-01-01: 4_300
@@ -14,15 +14,18 @@ metadata:
   rounding:
     type: downwards
     interval: 10
-  label: South Carolina young child deduction amount
+  label: South Carolina dependent exemption amount
   reference:
+    # 12-6-1140(13) defines the dependent exemption amount; 12-6-1160 sets the
+    # young child deduction equal to it.
+    - title: SC Legal Code | SECTION 12-6-1140
+      href: https://www.scstatehouse.gov/code/t12c006.php
+    - title: SC Legal Code | SECTION 12-6-1160
+      href: https://www.scstatehouse.gov/code/t12c006.php
     - title: SC 1040 tax form instructions (2025)
       href: https://dor.sc.gov/sites/dor/files/forms/SC1040Instr_2025.pdf#page=16
     - title: SC 1040 tax form instructions (2021)
       href: https://dor.sc.gov/forms-site/Forms/IITPacket_2021.pdf#page=19
-    # 12-6-1160 defines the program and points to the amount in 1140.
-    - title: SC Legal Code | SECTION 12-6-1160
-      href: https://www.scstatehouse.gov/code/t12c006.php
     - title: SC 1040 tax form (2022)
       href: https://dor.sc.gov/forms-site/Forms/SC1040_2022.pdf#page=2
     - title: SC 1040 tax form instructions (2023)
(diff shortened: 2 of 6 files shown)


PavelMakarchuk (changes requested):
## Program Review — Add dedicated South Carolina dependent exemption parameter

**Scope:** SC income tax parameter restructure (no-op), 2019–2025 · **CI:** ✅ 27/27 passing

Clean, well-justified restructure: it splits the shared parameter so `sc_dependent_exemption` (§12-6-1140(13)) and `sc_young_child_deduction` (§12-6-1160) each read their own dedicated parameter, with identical values and no calculated-value change. The statute is verified (§12-6-1140(13) is a genuine per-dependent exemption; §12-6-1160 sets the young-child deduction equal to it; the two stack), references corroborate every value 2019–2025, and the variable repoint is wired correctly with no dangling references. Requesting changes to add test coverage that locks in the separation this PR enables before merge.

### Requested changes

1. **Add a test that guards the split.** Because both parameters currently carry identical values, the existing `sc_dependent_exemption.yaml` (2 cases, both 2022) would pass even if the variable still read the old `young_child` parameter — so nothing actually pins the repoint. Since the whole point of the split is to let the two diverge (Child Poverty Impact Dashboard), please add a case that exercises the dependent exemption against its own parameter so future drift is caught.

2. **Test the stacking claim.** The PR states a dependent under 6 receives **both** deductions (~$8,220 combined), but no test combines them — each existing test exercises only one provision. Add a single case asserting both the dependent exemption and the young-child deduction fire for an under-6 dependent, to lock in the headline behavior.

3. **Add per-year symmetry (lower priority).** `sc_dependent_exemption` is tested only at 2022 while `young_child` covers 2022 + 2025. Now that the dependent exemption has its own parameter, add a 2025 dependent-exemption case for symmetry.

### Suggestions (non-blocking, pre-existing)
- The SC statute text reads *"four thousand one hundred ten"* ($4,110) for the base, while the 2019 parameter value is `4_100` (carried unchanged into the new file, not introduced here). Worth a one-time confirmation against the earliest SC1040 instructions.
- The 2022 reference points to the SC1040 *form* (p.2) rather than the instructions — confirm the value appears there.

### Validation Summary
| Check | Result |
|-------|--------|
| Regulatory Accuracy | ✅ Statute verified — §12-6-1140(13) per-dependent exemption; §12-6-1160 equal to it; the two stack |
| Reference Quality | ✅ Every value 2019–2025 corroborated; new file complete metadata + references; `young_child` edits reorder (don't weaken) citations |
| Code Patterns | ✅ Variable repointed correctly; no dangling references; values identical; changelog present; folder convention correct |
| Test Coverage | ⚠️ No test exercises the split independently; stacking case untested |
| CI Status | ✅ Passing (27/27) |

To auto-fix: `/fix-pr 8691`

DTrim99:
## Fixes Applied

Addressed the review (no critical issues; CI was green throughout). All changes are test additions — no model/parameter logic changed.

### Should-address
- **Repoint pin** — added a parameter-override test that sets `dependent_exemption.amount` and `young_child.amount` to distinct values and asserts `sc_dependent_exemption` follows the new `dependent_exemption` parameter (yields 18,000 = 2×9,000). Confirmed it would fail against the old `young_child` parameter (would produce 2,000), so the repoint is genuinely pinned.
- **Stacking** — added a test confirming an under-6 dependent receives both the dependent exemption and the young-child deduction independently for the same household.

### Suggestion
- Added a 2025 case for per-year symmetry.

### Out of scope (pre-existing, untouched)
- The $4,110-vs-$4,100 statute value and the 2022 form-vs-instructions reference are pre-existing and not introduced by this PR; left for a separate change.

### Verification
- `sc_dependent_exemption.yaml`: 5/5 pass; SC deductions+exemptions regression: 19/19 pass.
- No expected-value corrections were needed.
- `ruff format` + `ruff check`: clean.

PavelMakarchuk (approved):
Re-reviewed after the latest commit — all three requested test changes are addressed:

- **Case 3 (repoint pin)**: overrides both parameters to distinct values and asserts the variable reads `dependent_exemption` (2 × $9,000 = $18,000), so a future mis-pointing now fails.
- **Case 4 (stacking)**: an under-6 dependent gets both `sc_dependent_exemption` and `sc_young_child_deduction` ($4,930 each, 2025), proving §12-6-1140 / §12-6-1160 stack.
- **Case 5 (2025 symmetry)**: 3 dependents → $14,790.

CI is green. Approving. 🤖

Issue #8187, opened 2026-04-29
Deprecate legacy has_marketplace_health_coverage variable


MaxGhenis:
## Problem

`has_marketplace_health_coverage` is defined as a legacy modeled Marketplace coverage input, but it is easy to misread as reported or actual Marketplace coverage. In downstream prompt/evaluation work this created a contradiction: PolicyEngine could return a positive ACA Premium Tax Credit while the exposed household facts said `has_marketplace_health_coverage: false`.

From the current rules:

- `has_marketplace_health_coverage` is defined in `variables/household/expense/health/has_marketplace_health_coverage.py` with label "Person has modeled Marketplace health coverage" and `default_value = True`.
- `is_aca_ptc_eligible` does not use `has_marketplace_health_coverage`; it checks status, Medicaid/CHIP/ESI/Medicare eligibility, income eligibility, and age-based premium rules.
- Reported coverage logic uses `has_marketplace_health_coverage_at_interview` instead.
- `policyengine-us-data` currently populates `has_marketplace_health_coverage` as a legacy alias from `reported_has_marketplace_health_coverage_at_interview` for compatibility.

## Why this matters

The variable name suggests an ordinary coverage fact, but its rules-side meaning is legacy/model-facing. That makes it risky for public APIs, generated household prompts, data exports, and downstream users who do not know the historical distinction between modeled coverage and reported coverage at interview.

## Suggested path

- Deprecate `has_marketplace_health_coverage` in favor of explicit names.
- Keep `has_marketplace_health_coverage_at_interview` for reported survey coverage.
- If a modeled ACA take-up / enrollment concept is still needed, introduce a name that says so explicitly, for example `takes_up_aca_if_eligible`, `receives_marketplace_plan_if_eligible`, or another model-facing name consistent with existing ACA take-up variables.
- Add documentation or tests that `premium_tax_credit` / `is_aca_ptc_eligible` do not depend on reported Marketplace coverage at interview.
- Coordinate with `policyengine-us-data` to stop writing the legacy alias once compatibility permits, or mark it as deprecated there too.

## Downstream symptom

PolicyBench had to exclude `has_marketplace_health_coverage` from prompts because it leaked as a household fact and made no-tool models reasonably infer that a household with a listed Marketplace premium should receive no PTC. That was a prompt-interface issue rather than evidence that the PE PTC calculation was wrong.

PR #8693, opened 2026-06-19
Retire legacy Marketplace coverage input

### LINKED ISSUE #8674: Retire legacy Marketplace coverage input and canonicalize health benefit accounting
## Type

Pure rules / cleanup.

## Context

There are two related cleanup issues in the health-model surface:

1. `has_marketplace_health_coverage` is documented as a deprecated legacy modeled Marketplace coverage input, defaults to `True`, and says ACA PTC eligibility does not depend on it. The newer concepts are `has_marketplace_health_coverage_at_interview` for reported coverage and `takes_up_aca_if_eligible` for simulated take-up.

2. Health benefit accounting has overlapping paths: `household_health_benefits` uses the parameter list `medicaid`, `msp_cost`, `chip`, and monthly `premium_tax_credit`, while `healthcare_benefit_value` uses annual concepts such as `medicaid_cost`, `msp_cost`, `per_capita_chip`, and `assigned_aca_ptc`.

## Why this matters

Populace exports PE-US input variables. Deprecated input leaves with surprising defaults can become part of the dataset surface even when they no longer drive policy results. Separately, old monthly wrappers and newer annual health-value variables make it harder to know which output should be used for household resource analysis versus budget analysis.

## Suggested direction

- Remove `has_marketplace_health_coverage` from active model/export contracts if possible, or turn it into a clearly harmless compatibility alias.
- Add tests proving the deprecated variable does not affect ACA PTC, Marketplace premium, or health benefit outputs.
- Document/canonicalize which variables represent household health resource value and which represent government cost.
- Prefer one annual canonical ACA PTC variable in health accounting, or add tests proving the monthly wrapper conversion is intentional.

## Acceptance criteria

- Deprecated Marketplace coverage input no longer appears as an ambiguous active input in new datasets, or is explicitly guarded as compatibility-only.
- Health benefit accounting paths have documented, non-overlapping semantics.
- Tests prevent accidental double-counting or period-mismatch between `premium_tax_credit`, `assigned_aca_ptc`, `household_health_benefits`, and `healthcare_benefit_value`.

daphnehanse11:
Fixes #8674

## Summary

This PR retires the legacy Marketplace coverage input from the active input/export surface while preserving the variable name as a compatibility alias. It also canonicalizes household health benefit accounting on annual health-value variables, including `assigned_aca_ptc` rather than the monthly `premium_tax_credit` wrapper.

## API partner implications

This is intentionally API-surface-relevant:

- `has_marketplace_health_coverage` is no longer an active input variable discovered from PE-US metadata.
- The legacy variable remains calculable and now mirrors `has_marketplace_health_coverage_at_interview`.
- Partners exporting PE-US input variables should stop surfacing `has_marketplace_health_coverage`.
- Reported Marketplace coverage should use `has_marketplace_health_coverage_at_interview`.
- Simulated ACA take-up should use `takes_up_aca_if_eligible`.
- Annual health resource accounting should use `healthcare_benefit_value` / `assigned_aca_ptc`; `premium_tax_credit` remains a monthly wrapper.

## Health accounting semantics

- `healthcare_benefit_value`: annual canonical household resource value of health coverage.
- `household_health_benefits`: annual health value included in `household_benefits` only when `gov.simulation.include_health_benefits_in_net_income` is enabled.
- Program wrappers `medicaid`, `chip`, and monthly `premium_tax_credit` are no longer listed in household health benefits to avoid double-counting or period mismatch.

## Tests

- `uv run policyengine-core test policyengine_us/tests/policy/baseline/household/household_health_benefits.yaml policyengine_us/tests/policy/baseline/household/healthcare_benefit_value.yaml policyengine_us/tests/policy/baseline/household/expense/health/has_marketplace_health_coverage.yaml policyengine_us/tests/policy/baseline/gov/aca/ptc/premium_tax_credit.yaml policyengine_us/tests/policy/baseline/gov/aca/eligibility/is_aca_ptc_eligible.yaml policyengine_us/tests/policy/baseline/gov/aca/ptc/aca_ptc.yaml -c policyengine_us`
- `uv run pytest policyengine_us/tests/core/test_input_variable_definitions.py -q`
- `uv run policyengine-core test policyengine_us/tests/policy/baseline/household/household_health_benefits.yaml -c policyengine_us`
- `make format`
- `git diff --check`

## CI follow-up

An initial CI changelog check failed because the fragment used the old `changelog.d/fixed/8674.md` path. This branch now uses `changelog.d/8674.fixed.md`, matching the current PR workflow.


MaxGhenis (approved):
Approving — a clean soft-retirement plus health-accounting canonicalization, well-guarded by tests (agent-assisted review; I verified the load-bearing parts).

Nothing in the model reads `has_marketplace_health_coverage`: it keeps its name, drops `default_value = True`, and now aliases `has_marketplace_health_coverage_at_interview` via a formula, so it stops being an input variable but stays calculable (set-inputs still override). The effective default flips True→False, but with zero formula readers that's inert. #8187 already pulled it out of the PTC path; this finishes the retirement.

The `household_health_benefits` parameter lands on the same 7-variable list as `healthcare_benefit_value`: `medicaid`→`medicaid_cost` and `premium_tax_credit`→`assigned_aca_ptc` are value-neutral, and the real delta — picking up `co_omnisalud` and `or_healthier_oregon_cost` — only applies when `include_health_benefits_in_net_income` is on (default off). No double-count: Healthier Oregon is gated on federally-barred-non-citizen status (disjoint from `medicaid_cost`) and OmniSalud enrollees have no federal APTC (disjoint from `assigned_aca_ptc`). Case 4 nails the double-count guard. `fixed` is the right fragment type.

Two non-blocking notes:
- The canonical list now lives in two hand-kept places (`healthcare_benefit_value.adds` and the parameter). That's the drift #8674 wanted to kill — consider making `household_health_benefits` gate `healthcare_benefit_value` directly unless you're intentionally keeping the set reform-modifiable via the parameter.
- Downstream follow-up lives in populace: it still carries `has_marketplace_health_coverage` in `cps_carried.py` and asserts on it in `test_us_puf_support.py`. Nothing breaks, but the export surface should switch to the `_at_interview` variable to finish the retirement there.

PR #8724, opened 2026-06-22
Harden ACA ZIP3 parsing

### LINKED ISSUE #8675: Harden ACA SLCSP rating-area lookup for richer Populace geography
## Type

Mixed rules and data.

## Context

Populace can carry richer geography than the current ACA rating-area path appears to use. PE-US currently relies on `county_str` for the default SLCSP rating-area lookup, mutates the imported rating-area table's `rating_area` column to string, and silently fills missing rating areas with `1` before converting to int.

For Los Angeles County, PE-US uses a ZIP3-specific lookup, but `three_digit_zip_code` casts `zip_code` to int even though `zip_code` defaults to `UNKNOWN`. Missing or nonnumeric ZIPs can therefore be fragile in paths that require ZIP parsing.

## Why this matters

ACA PTC estimates depend on rating-area SLCSP. With Populace, we should be able to use stable geographic identifiers such as county FIPS, ZIP, ZCTA, or an explicit rating-area field. Silent fallback to rating area 1 can hide data or mapping failures and bias household-level credits and aggregate PTC estimates.

## Suggested direction

- Prefer a FIPS-based rating-area lookup where available.
- Consider accepting an explicit `aca_rating_area`/`slcsp_rating_area` input from Populace when source data can provide it.
- Replace silent missing-rating-area fallback with an explicit unknown sentinel or diagnostic.
- Make ZIP parsing robust to `UNKNOWN`, empty strings, leading-zero ZIPs, and nonnumeric values.
- Avoid mutating imported parameter/data tables inside variable formulas.

## Acceptance criteria

- Unknown county/rating-area mappings are visible in diagnostics or tests instead of silently becoming area 1.
- ZIP3 parsing handles missing and leading-zero ZIPs safely.
- Tests cover LA County ZIP lookup, non-LA county lookup, missing ZIP, missing county/rating area, and leading-zero ZIPs.
- Populace geography can be consumed through a documented PE-US contract.

daphnehanse11:
Partially fixes #8675.

## Summary
- Makes three_digit_zip_code parse ZIP codes as text before deriving ZIP3.
- Returns an empty ZIP3 for missing or nonnumeric ZIP codes instead of raising during ACA SLCSP paths.
- Adds tests for UNKNOWN, empty, nonnumeric, ZIP+4, and leading-zero ZIP values.

## Tests
- uv run pytest policyengine_us/tests/policy/baseline/household/demographic/geographic/test_three_digit_zip_code.py
- git diff --check


MaxGhenis (approved):
Approving — clean, strictly-improving hardening of the ZIP3 derivation.

Verified the behavior change is safe: for every pure-digit ZIP of length 1–5, the new `zfill(5)[:3]` is algebraically identical to the old `str(int(z)//100).zfill(3)`, so valid and leading-zero ZIPs (90019→900, 01234→012, 00601→006) are unchanged. The only outputs that move are ones that were already broken on main — ZIP+4 like "90019-1234" now resolves to "900" instead of failing the LA rating-area lookup, and missing/nonnumeric ZIPs return "" instead of the "UNKNOWN" passthrough. The "" sentinel is the right call: `slcsp_rating_area_la_county` already gates on `zip3 != ""`, and in `medicaid_rating_area` an empty string fails the merge exactly as "UNKNOWN" did, so no regression. Vectorized via pandas string ops on a cached YEAR variable.

I pushed one small commit to the branch: explicit tests for the two issue-named failure modes ("00601"→"006" and bare "1234"→"012", the leading-zero-dropped-by-int-parsing case) and a changelog wording fix — main already guards nonnumeric ZIPs, so the pre-fix failure was silent mis-parsing rather than a crash. All 9 tests in the file pass against this branch locally.

Correctly scoped as a partial fix for #8675 — the FIPS lookup, explicit rating-area input, and replacing the silent `fillna(1)` fallback are appropriately left for follow-up.

PR #8726, opened 2026-06-23
Cap DE non-refundable credits per spouse on separate/combined returns

### LINKED ISSUE #8710: DE separate/combined return: non-refundable credits capped against combined tax instead of each spouse's own tax
## Summary

On a Delaware separate/combined return (Filing Status 4, `de_files_separately = True`), PolicyEngine caps each non-refundable credit against the **combined** tax-unit liability instead of each spouse's **own** tax before credits. This lets one spouse's surplus personal credit offset the other spouse's tax, under-stating Delaware tax.

Per the PIT-RES form, on a separate/combined return each spouse computes tax and credits in their own column (A/B); a credit in one column can only offset that column's tax (Line 32 ≤ Line 26 per column).

## Reproduction

DE, MFJ, both age 20, 1 dependent, $21,790 taxable interest (split 50/50). Filed separate/combined.

| | Column A (spouse) | Column B (you) |
|---|---|---|
| Taxable income | $7,645 | $7,645 |
| Tax before credits | $168 | $168 |
| Personal credit (L27a) | $110 (1 exemption) | $220 (2 exemptions) |
| Credit used (L32) | $110 | $168 (capped at own tax) |
| Balance | **$58** | **$0** |

**Ground-truth DE tax (form & TAXSIM): $58.** Column B's $220 credit is limited to its own $168 of tax; the surplus $52 is lost.

**PolicyEngine: $8.33.** PE computes per-spouse tax before credits of $169.16 each ($338.33 combined), then applies all 3 exemptions × $110 = $330 against the combined $338.33:

```
de_files_separately            True
de_taxable_income_indv (per)   [7645.24, 7645.24, 0.0]
tax_before_nrc_indv (per)      [169.16, 169.16, 0.0]
tax_before_nrc_unit (combined) 338.33
de_personal_credit             330.0
de_non_refundable_credits      330.0
de_income_tax (PE final)       8.33
```

## Root cause

`de_non_refundable_credits.py` and `de_personal_credit.py` (and the other ordered credits) cap against `de_income_tax_before_non_refundable_credits_unit` — the combined liability:

```python
return ordered_capped_state_non_refundable_credits(
    tax_unit, period, ordered_credits,
    "de_income_tax_before_non_refundable_credits_unit",  # combined, not per-column
)
```

`de_income_tax_before_non_refundable_credits_unit` already selects between the per-spouse (`_indv`) and joint (`_joint`) computation via `de_files_separately`, but the **credit cap** is applied once against the pooled total, so surplus credit in one column spills onto the other column's tax.

Affected credits (`gov.states.de.tax.income.credits.non_refundable`): `de_personal_credit`, `de_aged_personal_credit`, `de_cdcc`, `de_non_refundable_eitc`.

## Proposed fix

When `de_files_separately` is True, cap each non-refundable credit at the **per-spouse** tax before credits (`de_income_tax_before_non_refundable_credits_indv`) and sum across spouses, rather than capping the pooled credit total against the combined liability. When filing jointly the current combined cap is correct.

Note: the dependent's exemption may be allocated to either column; for this household both allocations yield the same $58 total because both columns have equal $168 tax. A correct implementation should allocate to minimize wasted credit (matching the form/TAXSIM optimization).

## Context

Reported by Daniel Feenberg (NBER TAXSIM) in [PolicyEngine/policyengine-taxsim#1007](https://github.com/PolicyEngine/policyengine-taxsim/issues/1007). Also surfaced as a secondary finding while diagnosing [policyengine-taxsim#996](https://github.com/PolicyEngine/policyengine-taxsim/issues/996) (elderly separate/combined: PE $949.35 vs form $1,110).

cc @PavelMakarchuk

DTrim99:
Fixes #8710. **Draft for review** — flagging one TAXSIM-validated test change below.

## What

On a Delaware separate/combined return (Filing Status 4, `de_files_separately = True`), each spouse's non-refundable credits are limited to that spouse's **own** tax before credits (Form PIT-RES Line 32 ≤ Line 26, per column). PolicyEngine previously capped the *pooled* credits against the combined tax-unit liability, so one spouse's surplus personal credit could offset the other spouse's tax.

## Approach

- New person-level `de_non_refundable_credits_indv` holds each spouse's **column-tied** credits (own personal credit + own aged personal credit), which cannot move between columns.
- `de_non_refundable_credits`, when filing separately:
  1. applies each spouse's own credits against their own tax: `Σ min(own_i, tax_i)`;
  2. places the **allocable** credits (dependents' personal credits, child/dependent care credit, non-refundable EITC) in the column with the most remaining capacity, capped there — as a tax-minimizing filer would, since the form allows allocating exemptions between columns.
- Joint filers are unaffected (existing pooled cap retained).

## Results

| Case | Before | After |
|---|---|---|
| #1007 household (MFJ, 2 filers + 1 dep, equal-tax columns) | $8.33 | **$59.17** (≈ form's $58) |
| taxsimid 2682 (head $6.05 tax, spouse $3,225.65 tax, 1 dep + CDCC) | $2,429.20 | **$2,643.15** |

## ⚠️ Test change to review

This updates the existing `taxsimid 2682` integration test: `de_income_tax` **$2,429.20 → $2,643.15**. That prior value came from **TAXSIM-35, which pools the credits against combined tax** — exactly the behavior this PR fixes. In that household the head has only $6.05 of tax, so $213.95 of their own personal+aged credit is unusable; TAXSIM (and old PE) wrongly let it offset the spouse's tax. The new value is the correct per-column result, and is verified to still be lower than filing jointly ($2,705.30), so the couple's separate-filing choice remains valid. I don't have a filled DE form for this synthetic TAXSIM case, so this expected value is derived from the per-column logic confirmed by the #1007 form — please scrutinize.

## Tests

- Added a #1007 regression test (separate/combined, per-column cap → $59.17).
- Full DE suite passes (295 tests) with the 2682 update.

## Known limitation

Allocable credits are placed entirely in the single highest-capacity column. If allocable credits exceed that column's capacity, the remainder is not spilled to the other column (a rare edge case). Noted for reviewer.

## Context

Reported by Daniel Feenberg (NBER TAXSIM) in [PolicyEngine/policyengine-taxsim#1007](https://github.com/PolicyEngine/policyengine-taxsim/issues/1007).

### DIFF
--- a/changelog.d/de-separate-combined-credit-cap.fixed.md
+++ b/changelog.d/de-separate-combined-credit-cap.fixed.md
@@ -1,0 +1,1 @@
+Cap Delaware non-refundable credits at each spouse's own tax before credits on a separate/combined return rather than the combined tax-unit liability.
--- a/policyengine_us/tests/policy/baseline/gov/states/de/tax/income/integration.yaml
+++ b/policyengine_us/tests/policy/baseline/gov/states/de/tax/income/integration.yaml
@@ -67,6 +67,46 @@
     de_income_tax_before_non_refundable_credits_unit: 1_750.25
     de_income_tax_before_refundable_credits: 1_650.25
 
+# Regression test for policyengine-taxsim#1007: on a separate/combined return,
+# each spouse's non-refundable credits are capped at that spouse's own tax
+# before credits, so one spouse's surplus personal credit cannot offset the
+# other spouse's tax.
+- name: Separate/combined return caps each spouse's non-refundable credits at their own tax
+  period: 2025
+  absolute_error_margin: 0.01
+  input:
+    people:
+      head:
+        age: 20
+        taxable_interest_income: 10_895.24
+        is_tax_unit_head: true
+      spouse:
+        age: 20
+        taxable_interest_income: 10_895.24
+        is_tax_unit_spouse: true
+      child:
+        age: 10
+        is_tax_unit_dependent: true
+    tax_units:
+      tax_unit:
+        members: [head, spouse, child]
+        filing_status: JOINT
+    households:
+      household:
+        members: [head, spouse, child]
+        state_code: DE
+  output:
+    de_files_separately: true
+    # Per spouse: taxable income $7,645.24 -> tax $169.16.
+    de_taxable_income_indv: [7_645.24, 7_645.24, 0]
+    de_income_tax_before_non_refundable_credits_indv: [169.16, 169.16, 0]
+    # Personal credit: head column = self + dependent = 2 x $110 = $220;
+    # spouse column = 1 x $110 = $110. Each capped at its own $169.16 tax:
+    # min(220, 169.16) + min(110, 169.16) = 169.16 + 110 = 279.16.
+    # (Pre-fix PE pooled all $330 against the $338.33 combined tax -> $8.33.)
+    de_non_refundable_credits: 279.16
+    de_income_tax_before_refundable_credits: 59.17
+
 #new baseline integration test
 - name: baseline integration test separate
   period: 2022
@@ -224,8 +264,19 @@
       household:
         members: [person1, person2, person3]
         state_fips: 10  # DE
-  output:  # expected results from patched TAXSIM35 2024-04-04 version
-    de_income_tax: 2429.20
+  output:
+    # This couple files separately (Filing Status 4). Per-column tax before
+    # credits: person1 (head) $6.05, person2 (spouse) $3,225.65.
+    # Each spouse's own personal + aged credits ($220) are capped at their own
+    # tax: person1 uses $6.05 (the rest of their $220 is unusable against $6.05
+    # of tax), person2 uses $220. The allocable credits (dependent's $110
+    # personal credit + $252.50 child care credit = $362.50) are placed in
+    # person2's column, which has the remaining capacity, and are fully used.
+    # Total credits = 6.05 + 220 + 362.50 = 588.55; tax = 3,231.70 - 588.55.
+    # The prior expected value ($2,429.20) came from TAXSIM-35, which pools the
+    # credits against combined tax and so wrongly lets person1's unusable
+    # credits offset person2's tax (see policyengine-taxsim#1007).
+    de_income_tax: 2643.15
 
 - name: Tax unit with taxsimid 22359 in k21.its.csv and k21.ots.csv
   absolute_error_margin: 0.01
--- a/policyengine_us/variables/gov/states/de/tax/income/de_non_refundable_credits.py
+++ b/policyengine_us/variables/gov/states/de/tax/income/de_non_refundable_credits.py
@@ -16,9 +16,38 @@ class de_non_refundable_credits(Variable):
         ordered_credits = parameters(
             period
         ).gov.states.de.tax.income.credits.non_refundable
-        return ordered_capped_state_non_refundable_credits(
+        joint_credits = ordered_capped_state_non_refundable_credits(
             tax_unit,
             period,
             ordered_credits,
             "de_income_tax_before_non_refundable_credits_unit",
         )
+        # On a separate/combined return (Filing Status 4) each spouse's
+        # non-refundable credits are limited to that spouse's own tax before
+        # credits, so surplus credit in one column cannot offset the other
+        # spouse's tax.
+        person = tax_unit.members
+        own_credits = person("de_non_refundable_credits_indv", period)
+        indv_tax = person("de_income_tax_before_non_refundable_credits_indv", period)
+        # Each spouse first applies their own (column-tied) credits against
+        # their own tax.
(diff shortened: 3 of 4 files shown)


PavelMakarchuk (changes requested):
## Review: Cap DE non-refundable credits per spouse (FS4)

The **per-column cap concept is right and matches the form**, but the allocation of the "allocable" credits is regulatorily incorrect, so the headline `taxsimid 2682` value is wrong. Requesting changes. (You flagged the 2682 change for scrutiny — good call; it doesn't hold up.)

### Critical
1. **CDCC is routed to the wrong column.** The FS4 branch (`de_non_refundable_credits.py:43-50`) lumps dependents' personal credit + CDCC + non-refundable EITC into one bucket and places the whole bucket in the single **highest-remaining-capacity** column (`min_(allocable, tax_unit.max(remaining_capacity))`). But the 2024/2025 PIT-RES instructions are explicit: the **Child Care Credit (Line 31) may only be applied against the tax of the spouse with the LOWER taxable income (Line 23)**, and the **non-refundable EITC (Line 34) goes to the HIGHER-income spouse** — a deliberate asymmetry. Highest-capacity ≈ higher-income, so this routes the CDCC to the *opposite* column the form requires (and can't be right for both CDCC and EITC simultaneously). Only the Line-27a personal credits (incl. dependents') are genuinely allocable in $110 increments.
2. **`taxsimid 2682` expected value is wrong.** Head = lower-income ($6.05 pre-credit tax), spouse = higher ($3,225.65), CDCC = $252.50. The PR puts the CDCC on the spouse → fully used → $2,643.15. Form-correct: the CDCC is tied to the head, whose $6.05 is already exhausted by his own personal/aged credits, so the $252.50 CDCC is **wasted** → separate ≈ **$2,895.65**. That exceeds joint (~$2,705.30), so the couple files **joint** and the correct `de_income_tax` is **$2,705.30**, not $2,643.15. The prior $2,429.20 came from TAXSIM pooling credits (the very bug), but the replacement is also wrong (too low) — please don't lock it in.
3. **`tax_unit.max(remaining_capacity)` ranges over dependents** (`:29-31, :48-49`). `remaining_capacity` is Person-level over all members; a dependent with taxable income (own credits = 0) can become the `largest_remaining_capacity` and inflate the cap. Gate by `is_tax_unit_head_or_spouse`.

### Should
- The single lumped cap misapplies the per-column liability limit even apart from routing; credits should be routed per-credit and capped per-column (CDCC→lower, EITC→higher, personal→allocable), with spillover of the allocable personal portion.
- `de_non_refundable_credits.py` has no `reference` field despite adding statutory FS4 logic (the `_indv` sibling cites `#page=8`).
- Tests don't discriminate the bug: the #1007 case (→ $59.17) has no CDCC/EITC and symmetric columns (so "highest-capacity" == "lower-income" coincide); no test pins the CDCC to the lower-income column or the EITC to the higher. Add an asymmetric-column CDCC case.
- `de_cdcc_potential`/`de_non_refundable_eitc_potential` are the *correct* (pre-limitation) inputs — no change needed there; the defect is the routing, not the measure.

### Overlap with #7940 — recommend consolidating
This and **#7940** both fix DE FS4 credits and edit overlapping logic (`integration.yaml` textual conflict + a semantic conflict: #7940 restructures the FS4 path with per-column `_indv` credit variables — including a **column-tied `de_cdcc_indv`** — and rewires `de_income_tax_before_refundable_credits`, which would bypass/double-count this PR's branch). They **cannot both merge**. #7940's column-tied CDCC approach conforms to the PIT-RES lower-income rule and yields the correct $2,705.30 for 2682; this PR's highest-capacity approach does not. **Recommend adopting #7940's per-column structure** and folding in this PR's clean per-spouse-cap framing where useful, rather than landing #8726's allocation.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

DTrim99:
Closing in favor of #7940.

@PavelMakarchuk's review here was right, and I couldn't defend the headline value on re-examination. The highest-remaining-capacity allocation in this PR routes the CDCC and the non-refundable EITC to the wrong columns: per the PIT-RES rules the **Child Care Credit (Line 31) attaches to the lower-income spouse** and the **non-refundable EITC (Line 34) to the higher-income spouse** — a deliberate asymmetry that a single highest-capacity bucket can't honor. As a result the taxsimid 2682 value here ($2,643.15) is wrong; the form-correct result is **$2,705.30** (the CDCC is wasted on the head's exhausted $6.05 column, so separate ≈ $2,895.65 > joint, and the couple files joint).

#7940 implements the correct per-column structure (column-tied `de_cdcc_indv`, EITC to the higher-income column, the Line-27a personal-credit split in $110 increments, and a post-credit FS4-vs-joint election), and the two edits conflict, so only one should land. Deferring to #7940 — I've left a detailed cross-reference there: https://github.com/PolicyEngine/policyengine-us/pull/7940#issuecomment-4867572192

Issue #8682, opened 2026-06-18
Dashboard refundable-EITC reforms: fix MO/UT/OH + add SC full-refund contrib reform


DTrim99:
## Context

The [Child Poverty Impact Dashboard](https://github.com/PolicyEngine/child-poverty-impact-dashboard) is adding a "Make refundable" checkbox for states whose EITC is nonrefundable (SC, MO, OH, UT), wired to the `gov.contrib.states.{st}.child_poverty_impact_dashboard.eitc` reforms. Two upstream gaps currently block enabling that feature. The dashboard gates each state behind a `refundable_ready` flag (all `false` today) until this is resolved.

## Gap 1 — MO/OH/UT refundability reforms are broken

The three refundability reforms have two bugs:

**a) `add()` path-string crash.** Their `*_refundable_credits.formula` calls:

```python
add(tax_unit, period, "gov.states.mo.tax.income.credits.refundable")
```

`add()` expects a `List[str]` of variable names, not a parameter-path string. PE-core iterates the string character-by-character, `get_variable("g")` returns `None`, and the sim raises `AttributeError: 'NoneType' object has no attribute 'entity'`. Net effect: `household_net_income` silently collapses to `0` under the reform.

**b) Zero-refund bug.** They pay the post-cap *applied* credit (`mo_wftc` / `ut_eitc` / `oh_eitc`) instead of the `*_potential` (uncapped-at-liability) amount — so exactly the zero-liability filers refundability is meant to help get `$0`.

Files:
- `policyengine_us/reforms/states/mo/eitc/mo_refundable_eitc_reform.py` (~line 59)
- `policyengine_us/reforms/states/ut/child_poverty_eitc/ut_fully_refundable_eitc_reform.py`
- `policyengine_us/reforms/states/oh/eitc/oh_refundable_eitc_reform.py`

Fix PRs are already open and need review + merge + a release: **MO #8642, UT #8645, OH #8657**.

## Gap 2 — SC has no full-refundability contrib reform

`sc_eitc` is nonrefundable (in `gov.states.sc.tax.income.credits.non_refundable`, 2021+) — a 125% federal match capped at $200/filer from 2026. SC only has the unrelated **H.3492** partial reform (`gov.contrib.states.sc.h3492`, which refunds 25% of the excess over liability), not a full conversion.

**Request:** add a new contrib reform mirroring MO/UT:
[...]

PR #8645, opened 2026-06-16
Fix UT fully refundable EITC reform crash and zero-refund bug

PavelMakarchuk:
Fixes #8644

## Problem

Triggering the bundled Utah fully-refundable-EITC contrib reform (`gov.contrib.states.ut.child_poverty_impact_dashboard.eitc.in_effect`) crashed the simulation. After working around the crash, the reform also failed to actually pay the EITC as refundable for the low-liability filers refundability is meant to help — same dual-bug profile as the Missouri reform fixed in #8642.

## Three bugs fixed in `ut_fully_refundable_eitc_reform.py`

**1. `add()` received a parameter path string (crash).** `ut_non_refundable_credits.formula` passed the literal string `"gov.states.ut.tax.income.credits.non_refundable"` to `add()`, which iterates it character by character — `get_variable("g")` returns `None` → `AttributeError: 'NoneType' object has no attribute 'entity'`. Now resolves the parameter to its list of variable names first.

**2. Mixed computation modes (crash on core ≥3.26.8).** `ut_refundable_credits` redeclared a `formula` but inherited `adds` from the baseline variable, tripping the engine's strict mode check (`ValueError: ... mixes computation modes`). Now sets `adds = None` and `subtracts = None` explicitly.

**3. Refundable credit capped at tax liability (silent zero-refund).** `ut_fully_refundable_eitc` returned `ut_eitc` — the *applied* credit capped at UT tax liability — so a zero-liability filer received `$0` even though refundability should pay the full amount. Now uses `ut_eitc_potential` (uncapped at liability; the W-2 wages cap mandated by Utah Code § 59-10-1044 still applies as designed). Same functional fix the MO PR applied via `mo_wftc_potential`.

## Verification

Reproduced the issue's exact script — runs without raising. Verified the reform's effect against baseline using the YAML regression test below: a Utah filer with `eitc = 5_000`, `employment_income = 30_000`, and `ut_income_tax_before_non_refundable_credits = 0` (i.e. zero state liability) now receives the full `ut_eitc_potential = 1_000` as `ut_refundable_credits`, where the buggy reform returned `$0`.

## Tests

- New `tests/policy/reform/ut_fully_refundable_eitc.yaml`: verifies the full potential UT EITC is paid as refundable at zero liability, the nonrefundable portion is zeroed, and there's no double-counting when liability already absorbs the credit.
[...]

PR #8642, opened 2026-06-15
Fix MO refundable EITC reform crash and zero-refund bug

PavelMakarchuk:
Fixes #8640

## Problem

Triggering the bundled Missouri refundable-EITC contrib reform (`gov.contrib.states.mo.child_poverty_impact_dashboard.eitc.in_effect`) crashed the simulation. After working around the crash, the reform also failed to actually make the WFTC refundable for the low-liability filers refundability is meant to help.

## Three bugs fixed in `mo_refundable_eitc_reform.py`

**1. `add()` received a parameter path string (crash).** `mo_refundable_credits.formula` passed the literal string `"gov.states.mo.tax.income.credits.refundable"` to `add()`, which iterates it character by character — `get_variable("g")` returns `None` → `AttributeError: 'NoneType' object has no attribute 'entity'`. Now resolves the parameter to its list of variable names first.

**2. Mixed computation modes (crash on core ≥3.26.8).** `mo_refundable_credits` redeclared a `formula` but inherited `adds` from the baseline variable, tripping the engine's strict mode check (`ValueError: ... mixes computation modes`). Now sets `adds = None` and `subtracts = None` explicitly.

**3. Refundable credit capped at tax liability (silent zero-refund).** `mo_refundable_wftc` returned `mo_wftc` — the *applied* credit capped at MO tax liability — so a zero-liability filer received `$0` even though refundability should pay the full amount. Now uses `mo_wftc_potential` (the uncapped credit). This is the functional fix that makes the reform do what it claims.

## Verification

Reproduced the issue's exact script — runs without raising. Verified the reform's effect against baseline for a Missouri single parent, 2 kids, $20k earnings, 2026 (federal EITC $7,316 → potential WFTC $1,463):

| Variable | Baseline | Reform |
|---|---|---|
| `mo_refundable_credits` | $0 | $1,463.20 |
| `mo_income_tax` | $0 | −$1,463.20 (refund) |
| `household_net_income` | $36,103.81 | $37,567.01 |

Also confirmed no double-counting at positive liability: a moderate-income case yields a delta exactly equal to the previously-lost (nonrefundable-capped) portion of the credit.

## Tests

- New `tests/policy/reform/mo_refundable_eitc.yaml`: verifies the full potential WFTC is paid as refundable at zero liability, the nonrefundable portion is zeroed, and there's no double-counting when liability already absorbs the credit.
- All 361 Missouri baseline tests still pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #42, opened 2021-08-22
Add tax input and output variables

nikhilwoodruff:
Fixes #21 and #22 - adds the input and output variables to `tax-calculator`: no formulas, just scraped from [the documentation](http://taxcalc.pslmodels.org/guide/output_vars.html) (name, description and value type). In cases where the value type was integer and the description indicated "1 if ..., otherwise 0", I changed the value type to bool and 1/0 to true/false. The few variables we've got should need to be renamed/integrated with these, but this should make it so that we can replicate the formulas without needing to add variables now.

PR #8777, opened 2026-06-29
Fix create-state EITC reforms deleting baseline refundable credits (ID/AZ/WV/NC)

### LINKED ISSUE #8775: State EITC contrib reforms delete baseline refundable credits (ID, AZ, NC, WV)
## Summary

Several `child_poverty_impact_dashboard` state EITC contrib reforms don't simply *add* a refundable state EITC — they **override `{state}_refundable_credits` to return only the new EITC**, on an explicit "no baseline refundable credits" assumption that is **false** for some states. Where it's false, the reform **deletes the state's existing refundable credits**, producing a *negative* budgetary cost and *increasing* poverty/inequality for a reform that should only ever cost money and reduce poverty (static microsim).

Affected (negative aggregate cost when the EITC is switched on): **ID, AZ, NC, WV**.

## The pattern

`reforms/states/id/eitc/id_eitc_reform.py` (same shape in `az`, `wv`, and — with an extra twist — `nc`):

```python
class id_refundable_credits(Variable):
    # NOTE: ID currently has no baseline refundable credits.
    # If ID adds refundable credits in baseline, this formula must be updated.
        return tax_unit("id_eitc", period)   # REPLACES, does not add
```

When the state *does* have baseline refundable credits, this override wipes them.

## Evidence (policyengine-us 1.745.0, household sims)

Married couple, 2 children, reform = `gov.contrib.states.{st}.child_poverty_impact_dashboard.eitc.{in_effect=true, match=0.05}`:

| State | Finding |
|---|---|
| **ID** | `$60k` family: baseline `id_refundable_credits` = **$620** (grocery/food-tax credit) → reform = **$62**, **net −$558**. Baseline credit is wiped. |
| **AZ** | baseline `az_refundable_credits` = **$100** at $0–15k earnings; replaced by the (here ~$0) EITC. |
| **NC** | `gov.contrib.states.nc.eitc` reform also overrides **`nc_income_tax`** (subtracting the replaced `nc_refundable_credits`); `$90k` family nets **−$61** — a tax increase. |
| **WV** | Same `wv_refundable_credits` override; −$166M aggregate in a per-state sweep. |

For contrast, states that genuinely have no baseline refundable credits (GA, AL, MS, KY, AR, ND) behave correctly — positive cost, lower poverty.

These surfaced in a per-state budgetary sweep of the Child Poverty Impact Dashboard reforms: ID `−$271.0M` (child poverty **+0.08pp**), NC `−$186.5M`, WV `−$166.6M`, AZ `−$106.5M` — directionally impossible for a refundable EITC in a static microsimulation.

## Suggested fix

The `{state}_refundable_credits` override should **add** the new EITC to the existing baseline refundable credits rather than replace them, e.g.:

```python
    return baseline_refundable_credits + tax_unit("id_eitc", period)
```

(or add `{state}_eitc` into the state's existing refundable-credits aggregation). The "no baseline refundable credits" assumption in the NOTE is incorrect for ID (grocery credit), AZ, and likely WV; NC additionally needs its `nc_income_tax` override reviewed so it doesn't drop baseline refundable credits.

## Impact

Any tool that activates these contrib EITC reforms (e.g. the Child Poverty Impact Dashboard "create a state EITC" option) shows nonsensical results for these four states until fixed.

DTrim99:
## What this does

Fixes the create-state EITC contrib reforms that delete baseline refundable credits (sibling of the #8682 umbrella that fixed MO/OH/UT/SC). Fixes #8775.

A per-state budget sweep on 1.745.0 showed activating these refundable EITCs produced **negative cost and rising child poverty** — impossible for a refundable credit in a static microsim (ID −$271.0M, NC −$186.5M, WV −$166.6M, AZ −$106.5M).

## Root cause

- **ID/AZ/WV** override `{st}_refundable_credits` to return only `{st}_eitc`, replacing the baseline `adds` list and **deleting each state's real baseline refundable credits** (ID: grocery, aged/disabled; AZ: increased excise, property tax, families rebate; WV: SCTC, homestead excess property tax). Traced to commit 0ba00db243, which simplified the formulas on a "no baseline refundable credits" assumption that's false for these three.
- **NC** is a separate flavor: its `nc_income_tax` override added `nc_use_tax`, which is **already counted at the household level** via `state_use_tax` (`household_state_tax_before_refundable_credits` adds both `state_income_tax_before_refundable_credits` and `state_use_tax`). So it double-counted use tax, raising tax for every NC filer (−$61 even with no EITC).

## Fix

Following the MO #8642 / UT #8645 pattern:
- **ID/AZ/WV** `{st}_refundable_credits`: `adds = None` / `subtracts = None`; `formula = add(tax_unit, period, list(<baseline refundable param>)) + tax_unit("{st}_eitc", period)` — preserve the baseline list and add the EITC.
- **NC** `nc_income_tax`: mirror baseline (`nc_income_tax_before_credits` minus credits), no `nc_use_tax`.

## Open question — resolved

> Is `nc_use_tax` belonging in `nc_income_tax` a separate baseline question (is baseline `nc_income_tax` wrong)?

**No.** `nc_use_tax` is in `gov.states.household.state_use_tax`, so NC use tax is already counted at the household level via the separate `state_use_tax` path. Baseline `nc_income_tax` correctly **excludes** use tax. The reform was double-counting; baseline is right. No separate baseline bug.

## Tests

- Updated AZ/ID unit cases to reflect the preserved baseline credits (AZ +$25 excise → 125/525/25/25; ID +$120 grocery → 220/620/120/120). AZ values confirmed against the model (125/525/25/25).
- **Added regression/integration tests for all four states** that supply baseline refundable credits as inputs and assert they're preserved under the reform — and that NC income tax excludes use tax — so this bug class can't silently return.

## Verification notes

- Household sims (fix overlaid on 1.745.0): all deltas now ≥ 0 — ID @$60k −$558→+$62, AZ @$12k +$240 (excise preserved), WV @$60k +$62, NC @$90k −$61→$0, GA control +$360.
- The full PE-US suite couldn't be run in this environment (pinned core vs newer venv core → `social_security` mixed-mode error; and model runs were memory-bound), so the ID unit values and the new regression tests are computed by hand from the reform formulas and the baseline refundable lists rather than run locally — **CI is the authoritative check** for those.
- Dashboard side already handled: CPID PR #42 gates ID/AZ/WV/NC `in_development` until a fixed PE-US release ships.

### DIFF
--- a/changelog.d/fix-create-state-eitc-refundable-credits.fixed.md
+++ b/changelog.d/fix-create-state-eitc-refundable-credits.fixed.md
@@ -1,0 +1,1 @@
+Fixed create-state EITC contrib reforms (ID, AZ, WV) that replaced each state's refundable-credits list with only the new EITC, deleting baseline refundable credits, and the NC reform that double-counted use tax in income tax.
--- a/policyengine_us/reforms/states/az/eitc/az_eitc_reform.py
+++ b/policyengine_us/reforms/states/az/eitc/az_eitc_reform.py
@@ -19,17 +19,29 @@ def create_az_eitc() -> Reform:
             return federal_eitc * p.match
 
     class az_refundable_credits(Variable):
-        # NOTE: AZ currently has no baseline refundable credits.
-        # If AZ adds refundable credits in baseline, this formula must be updated.
         value_type = float
         entity = TaxUnit
         label = "Arizona refundable credits"
         unit = USD
         definition_period = YEAR
         defined_for = StateCode.AZ
+        # Baseline computes this via `adds`; replacing it with a
+        # formula requires clearing the inherited computation mode
+        # (the core engine rejects `formula` + `adds`/`subtracts`).
+        adds = None
+        subtracts = None
 
         def formula(tax_unit, period, parameters):
-            return tax_unit("az_eitc", period)
+            # Preserve the state's existing baseline refundable
+            # credits and ADD the new EITC. Returning only the EITC
+            # deletes the baseline refundable credits (this state
+            # has them), flipping the reform's sign. See #8775.
+            baseline_credits = parameters(
+                period
+            ).gov.states.az.tax.income.credits.refundable
+            return add(tax_unit, period, list(baseline_credits)) + tax_unit(
+                "az_eitc", period
+            )
 
     class reform(Reform):
         def apply(self):
--- a/policyengine_us/reforms/states/id/eitc/id_eitc_reform.py
+++ b/policyengine_us/reforms/states/id/eitc/id_eitc_reform.py
@@ -19,17 +19,29 @@ def create_id_eitc() -> Reform:
             return federal_eitc * p.match
 
     class id_refundable_credits(Variable):
-        # NOTE: ID currently has no baseline refundable credits.
-        # If ID adds refundable credits in baseline, this formula must be updated.
         value_type = float
         entity = TaxUnit
         label = "Idaho refundable credits"
         unit = USD
         definition_period = YEAR
         defined_for = StateCode.ID
+        # Baseline computes this via `adds`; replacing it with a
+        # formula requires clearing the inherited computation mode
+        # (the core engine rejects `formula` + `adds`/`subtracts`).
+        adds = None
+        subtracts = None
 
         def formula(tax_unit, period, parameters):
-            return tax_unit("id_eitc", period)
+            # Preserve the state's existing baseline refundable
+            # credits and ADD the new EITC. Returning only the EITC
+            # deletes the baseline refundable credits (this state
+            # has them), flipping the reform's sign. See #8775.
+            baseline_credits = parameters(
+                period
+            ).gov.states.id.tax.income.credits.refundable
+            return add(tax_unit, period, list(baseline_credits)) + tax_unit(
+                "id_eitc", period
+            )
 
     class reform(Reform):
         def apply(self):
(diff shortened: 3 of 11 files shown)


daphnehanse11:
## PolicyEngine Review: APPROVE

No blocking findings.

### Scope

Reviewed PolicyEngine/policyengine-us#8777 in a detached clean worktree at /private/tmp/policyengine-us-pr-8777-worktree.yqWwBW. The PR fixes contrib EITC reform behavior for AZ, ID, WV, and NC. It does not change statutory parameter values or references, so PDF and regulatory source audits were skipped as not applicable.

### Validation

Code review found the AZ, ID, and WV reform overrides use the correct pattern for replacing an adds-based baseline variable: clear adds/subtracts, resolve the baseline refundable credit list from parameters, sum it with add(), then add the new EITC variable. This preserves the existing refundable credits instead of replacing them.

NC review confirmed the reform now mirrors baseline nc_income_tax and no longer adds nc_use_tax inside the income tax variable. That is correct because nc_use_tax is already included through the state_use_tax path and household_state_tax_before_refundable_credits adds state_use_tax separately.

Test review found the new regression tests target the failure mode directly. AZ, ID, and WV tests supply baseline refundable credit variables as inputs and assert preservation; NC tests cover both zero-EITC and positive-EITC cases with a use-tax input.

### Verification

Local targeted run over the six changed YAML files collected 18 cases and printed 18 passed in 102.07s; the runner then hung in pytest cleanup and was terminated after the pass summary. I also reran the NC EITC folder through the repo batch helper, which exited 0: nc_eitc.yaml 4 passed and nc_eitc_excludes_use_tax.yaml 2 passed. GitHub CI for the PR reports all listed checks passing.

### Recommendation

APPROVE. No critical or should-address issues found.

daphnehanse11 (approved):
Approved. No blocking findings; see the posted PolicyEngine review report.

PR #8800, opened 2026-06-30
Add discrete per-filing-status AGI schedule to MD/OH dependent exemption contrib

DTrim99:
Resolves #8799.

## Problem
The dashboard drives most states' dependent exemption through the `gov.contrib.states.{st}.dependent_exemption` contrib (us#8696). That contrib offers a flat `amount` (or `-1` sentinel reading baseline), a single **linear** phase-out, and `age_limit` — which fits RI cleanly. **MD and OH don't fit**, because their dependent exemption is a **discrete, income-stepped, per-filing-status** schedule, so consumers had to reach into baseline brackets instead.

## Change
Give MD and OH a first-class `schedule` sub-tree in their contrib, mirroring the baseline, so the discrete step-down is adjustable entirely through the contrib (aligned with the rest of the us#8696 family):

- **MD** — per-filing-status bracket params `schedule/{single,separate,joint,head,surviving_spouse}.yaml`, mirroring the baseline §10-211 AGI step-down ($3,200 → $1,600 → $800 → $0 with filing-status-specific cutoffs).
- **OH** — `schedule/amount.yaml`, a MAGI-stepped bracket param mirroring the baseline §5747.025 schedule ($2,400 → $2,150 → $1,900 → $0).

The reform's dependent-exemption *maximum* now reads the stepped per-dependent amount from the contrib `schedule` rather than the baseline params. Defaults equal baseline, so:
- the `-1` `amount` sentinel remains a **no-op** at default (verified by the existing no-op tests),
- a non-negative flat `amount` still overrides the schedule,
- the linear `phaseout` still applies on top.

## Tests
Added to each state's contrib reform test file:
- **default schedule reproduces the AGI/MAGI step-down** — a higher-income unit lands on a lower step purely through the contrib (MD single AGI 130k → $800/exemption; OH MAGI 50k → $2,150/exemption).
- **schedule per-dependent amount is adjustable** — overriding `schedule.single[0].amount` (MD) / `schedule.amount[0].amount` (OH) changes only the dependent portion, leaving the baseline personal portion intact.

All existing no-op / flat-amount / phase-out / age-limit cases are unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #8816, opened 2026-07-01
Add dependent-exemption/credit age-gate contributed reforms for 13 states

DTrim99:
## What

Adds `gov.contrib.states.{st}.dependent_exemption` / `dependent_credit` contributed reforms so a state's per-dependent exemption or credit can be **adjusted, eliminated, or age-limited** — restricted to dependents under a chosen age. This lets tools (e.g. the child-poverty dashboard) model *eliminating a state's dependent exemption/credit only for young children and swapping it for a child allowance*.

Extends the existing us#8696 / #8800 dependent-exemption contrib family (HI, MD, OH, RI, DE, VA, MI, NE, OK, VT, WI, WV) to the remaining states that had a dependent exemption/credit but **no age gate**.

## New contribs

Each has `in_effect`, `amount` (default reproduces the baseline so activating at default is a **no-op** — a negative sentinel means "use the baseline schedule" for the stepped states), and `age_limit` (`in_effect` + `threshold`):

- **Separate-variable states:** NY, IL, MS, NJ, SC
- **Bundled (personal + dependent) carve-out:** GA, KS, MN — MN's AGI phase-out is preserved
- **Exemption-credit states:** CA, IA — the per-dependent portion is separated from the personal portion
- **Income/age-stepped:** AL (AGI-stepped; reads the baseline schedule via a `-1` sentinel), AZ (age-based schedule)

Also **adds an `age_limit` sub-tree to the existing AR `dependent_credit` contrib** and age-gates its person-level per-dependent amount (previously left un-age-gated).

## Behavior / tests

Every state has a YAML test with the same contract:
1. contrib in effect at the default amount → **reproduces baseline** (no-op),
2. `amount: 0` → **eliminates the dependent portion only** (personal exemptions / other credits preserved),
3. `age_limit` → applies the exemption/credit **only to dependents under the threshold** (older dependents keep the baseline).

All new tests pass against the current engine (verified locally via `policyengine-core test`). Bundled states additionally assert that over-age dependents fall back to the baseline personal treatment; NJ leaves its college-dependent exemption untouched; SC leaves its young-child deduction untouched.

## Also

Fixes `CLAUDE.md` to note the default branch is `main`, not `master` (it previously said "PRs targeting master branch").

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -53,6 +53,12 @@ make documentation
 ```
 
 ## GitHub Workflow
+- **Default branch is `main`, NOT `master`.** Base new work on `main`:
+  `git fetch upstream main && git checkout -b <branch> upstream/main`. A personal
+  fork's `origin/master` is often stale or absent — `git checkout master` can
+  silently land you on an ancient commit (e.g. a 1.44.x-era tree missing recent
+  contribs), so always branch from `upstream/main` (or `origin/main` when the fork
+  is current). Note the upstream remote uses `main` and has no `master` ref.
 - Checkout a PR: `gh pr checkout [PR-NUMBER]`
 - View PR list: `gh pr list`
 - View PR details: `gh pr view [PR-NUMBER]`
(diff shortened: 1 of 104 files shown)


MaxGhenis (approved):
Verified the no-op chain for all 13 states (baseline variable exists, reform override reproduces baseline value-for-value at default, in_effect defaults false, registration complete at all three reforms.py sites), with deep dives on CA/NY/IL/MN and the two-regime KS case. CI green on the exact head SHA including all contrib state shards. Remaining notes are cosmetic (in_effect sentinel-date inconsistency, changelog tense).

PR #8818, opened 2026-07-01
Fix WI retirement income exclusion applied after standard-deduction phaseout (#8817)

DTrim99:
Fixes #8817.

## Problem

Wisconsin's Act 15 retirement income exclusion (Schedule SB line 16, \$24,000/person for age 67+) was applied **after** the standard-deduction phaseout. On Form 1 the line-16 subtraction reduces WI income (line 7), and the standard deduction (line 8) is looked up on that reduced income — which is larger. Applying the subtraction after the phaseout kept the standard deduction too small and over-taxed elderly filers with retirement income.

## Root cause

`wi_retirement_income_exclusion_tax` computed `wi_taxable_income − line16`, but `wi_taxable_income` already baked in the standard deduction computed on the higher pre-subtraction WI income.

## Fix

Add back the extra standard deduction the line-16 subtraction unlocks — algebraically equivalent to recomputing taxable income on the post-subtraction WI income, while keeping `wi_taxable_income` as the base (so existing behaviour below the phaseout threshold is unchanged). Factored the standard-deduction phaseout into a reusable `wi_standard_deduction_for_income` helper so both the normal path (on `wi_agi`) and the exclusion path (on `wi_agi − line16`) share one implementation.

## Verification / tests

- New regression test (67+ MFJ, \$120k pension, exclusion \$48k): exclusion taxable income \$53,651, exclusion tax **\$2,184** (was ~\$2,602 pre-fix — the phaseout stayed keyed to \$120k).
- Updated the affected E2E case (single 45k, exclusion 20k): tax **474 → 390**, reflecting the recovered \$2,400 of standard deduction.
- The five low-income unit tests are unaffected (income below the phaseout threshold ⇒ no extra deduction), and the age-85 high-income integration case is unchanged (income stays fully phased out even after the subtraction).

## References
- 2025 Form 1 (lines 6-11) and Schedule SB line 16
- 2025 Schedule SB instructions

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #8679, opened 2026-06-17
Make CHIP health benefit value respect enrollment

daphnehanse11:
Fixes #8673.

## Summary

- Change `healthcare_benefit_value` to include CHIP through `chip`, which is gated by `chip_enrolled`.
- Document CHIP net/gross and eligibility/enrollment semantics on the existing CHIP value and cost variables.
- Add healthcare benefit value tests for CHIP enrolled, CHIP eligible but not enrolled, and CHIP ineligible cases.

## CHIP Semantics

- `chip`: enrolled, take-up-aware, net CHIP benefit value.
- `per_capita_chip`: eligibility-gated per-capita net CHIP payment; not a household resource accounting output by itself.
- `chip_gross` and `per_capita_chip_gross`: eligibility-gated gross value concepts, kept separate from household health benefit value.
- `chip_federal_cost` and `chip_state_cost`: enrollment-gated budget cost variables through `chip`.

## Validation

- `uv run policyengine-core test policyengine_us/tests/policy/baseline/household/healthcare_benefit_value.yaml -c policyengine_us`: 5 passed, 1 warning.
- `uv run policyengine-core test policyengine_us/tests/policy/baseline/gov/hhs/chip -c policyengine_us`: 70 passed, 1 warning.
- `uv run policyengine-core test policyengine_us/tests/policy/baseline/household/household_health_benefits.yaml -c policyengine_us`: 3 passed, 1 warning.
- `make format`: passed.
- `git diff --check`: passed.

## Known Failing Tests

None.

PR #8886, opened 2026-07-04
Speed up CI: fix NY tax-benefit-system cloning, run batches concurrently, rebalance to 14 runners

### LINKED ISSUE #8114: NY EITC/CTC formulas clone the full tax-benefit system on every call — CI memory blowup
## Summary

Three NY formulas clone the entire `tax_benefit_system` on every call to compute federal EITC/CTC with pinned-date parameters. Each clone is a deep copy of the full parameter tree and variable registry. Because the clone happens inside the formula body (not cached), every simulation that hits the branch pays the full cost, which drives CI memory pressure on the NY credits test batches.

Unlike #8108 (MD), these formulas are **correctness-safe** — they clone onto a simulation branch rather than mutating the shared baseline. The issue is purely memory cost.

## Affected files

| File | Pin date | Trigger |
|---|---|---|
| `policyengine_us/variables/gov/states/ny/tax/income/credits/ny_eitc.py` (L25-43) | 2020-01-01 (pre-ARPA) | `p.eitc.decoupled == True` (TY 2021 under NY decoupling) |
| `policyengine_us/variables/gov/states/ny/tax/income/credits/ctc/ny_ctc_pre_2024.py` (L29-72) | 2017-01-01 (pre-TCJA) | `p.pre_tcja == True` |
| `policyengine_us/variables/gov/states/ny/tax/income/credits/ctc/ny_ctc_pre_2024_eligible.py` (L28-49) | 2017-01-01 (pre-TCJA) | `p.pre_tcja == True` |

Pattern in all three:

```python
simulation = tax_unit.simulation
branch = simulation.get_branch("...")
branch.tax_benefit_system = simulation.tax_benefit_system.clone()   # full deep copy, per call
branch_params = branch.tax_benefit_system.parameters
for param in branch_params.gov.irs.credits.<...>.get_descendants():
    if isinstance(param, Parameter):
        param.update(start=..., stop=..., value=param(<pin_date>))
for variable in branch.tax_benefit_system.variables:
    if "<program>" in variable:
        branch.delete_arrays(variable)
federal_<...> = branch.tax_unit("<program>", period)
```

## Why the MD fix (#8113) does not apply here

MD (#8108) only needed to skip one boolean gate (`eitc_demographic_eligible`) — we recomposed `eitc_phased_in`, `eitc_maximum`, `eitc_reduction` inline. That works because the non-demographic federal gates are already well-isolated.

NY needs the **full federal EITC / CTC computation** to run with a different parameter vintage: phase-in, phase-out, max, investment-income test, SSN, filer status, refundable vs non-refundable split, ACTC earned-income floor, etc. Recomposing all that inside the NY formula would be a large correctness risk and would couple NY tightly to every future federal-formula refactor.

## Proposed fix — module-level cache of the pinned tax-benefit system

Build each pinned TBS **once** and share it across all calls. The cache is keyed on the baseline TBS object (weak ref, GC'd with the baseline — matches the test runner's pattern at `policyengine_core/tools/test_runner.py:71`).

Expected impact: one clone per process per pin date, instead of one clone per simulation that hits the branch. No change in values returned.

Sketch (ny_eitc):

```python
import weakref
_PRE_ARPA_TBS_CACHE: "weakref.WeakValueDictionary" = weakref.WeakValueDictionary()

def _pre_arpa_eitc_tbs(base_tbs):
    cached = _PRE_ARPA_TBS_CACHE.get(id(base_tbs))
    if cached is not None:
        return cached
    tbs = base_tbs.clone()
    pin = instant("2020-01-01")
    start = instant("2021-01-01")
    stop = instant("2021-12-31")
    for param in tbs.parameters.gov.irs.credits.eitc.get_descendants():
        if isinstance(param, Parameter):
            try:
                param.update(start=start, stop=stop, value=param(pin))
            except Exception:
                pass
    _PRE_ARPA_TBS_CACHE[id(base_tbs)] = tbs
    return tbs
```

The two `ny_ctc_pre_2024*.py` files would share a single `_PRE_TCJA_TBS_CACHE`. Alternative: use `policyengine_core.reforms.Reform` (a structured parameter override spec) with the same module-level cache — semantically cleaner and the cached object is explicitly meant to be read-only.

## Caveats for reviewers

- The pinned TBS must be treated as read-only by all callers; otherwise one caller's mutation leaks into another's results. The `Reform`-based alternative avoids that class of bug.
- `ny_ctc_pre_2024.py:51-55` calls `branch.set_input("ctc_individual_maximum", ...)`. Verify this still writes to branch holders (not the shared TBS) once the TBS is cached — quick manual check before rollout.
- These formulas do not hit the MD correctness bug because they do not mutate `simulation.tax_benefit_system` in place. Fixing this is an optimization, not a correctness fix.

## Related

- #8108 / #8113 — the MD variant of the shared-TBS pattern (correctness bug, fixed by recomposing the formula).

hua7450:
## Summary

CI took ~34 min wall-clock across 17 runners, sat at the memory cliff (single batches measured at 12-14.6 GB on 16 GB runners — the source of the historical intermittent "runner lost communication" failures), and needed hand-rearranging as tests grew. This PR fixes the worst per-case cost bug, re-batches the suite from **measured** per-batch peak RSS, and adds permanent timing/memory telemetry to every CI log.

Final state (validated by 3 full runs on this PR): **all green, zero OOMs, ~26 min wall on 23 full-suite runners, every subprocess ≤ ~8.5 GB peak** — half of each runner's memory left as headroom so newly added heavy tests don't require re-tuning.

Fixes #8114

### 1. NY #8114 fix (`policyengine_us/tools/pinned_tbs.py` + 3 NY credit variables)
The NY EITC / pre-2024 CTC formulas deep-cloned the entire tax-benefit system per formula call. Each pinned system (pre-ARPA 2020 EITC, pre-TCJA 2017 CTC) is now built once per process in a module-level cache keyed on the baseline system's identity (weakref — layered reforms rebuild against their own base). Measured: `ny_ctc_pre_2024.yaml` 373s → 57s, `ny_eitc.yaml` 106s → 53s, peak RSS 4.8 → 2.8 GB, outputs identical. The dedicated `states-ny` CI runner is retired; NY runs in the ordinary states shards.

### 2. Permanent CI telemetry
- Every job sets `PYTEST_ADDOPTS="--durations=25 -p no:unraisableexception -p no:threadexception"` — each log ends with the 25 slowest test cases, and subprocesses skip pytest 8.4's ~11s full-heap gc sweep at exit.
- `test_batched.py` reports each batch subprocess's true peak RSS (`VmHWM` from `/proc`) in its footer and the summary. Heavy new tests become visible in green runs, not red ones.
- (`test_batched.py` also gained an opt-in `--workers N` pool; CI uses 2 only for the Partners job, whose batches max at 2.5 GB. Everything else runs single-worker after measurement showed reform-heavy batches co-scheduled 2-wide exceed 16 GB.)

### 3. Memory-headroom re-batching (measured, not estimated)
Per-batch peak RSS from this PR's own runs drove the layout: states `--batches 16` across 4 shards, household `--batches 4`, `gov/simulation` and `baseline/contrib/states` per-file, `oh` joins `ri` in contrib per-file isolation, congress/irs/ssa heavy batches run strictly solo. Two files were too big for any batching and are split by reform-parameter combination with byte-identical cases: `contrib/states/ri/ctc_reform_test.yaml` (**14.6 GB alone**) → 3 files; `contrib/crfb/tax_employer_payroll_tax_percentage.yaml` (9.6 GB) → 2 files.

### 4. Process-lifetime memory bugs in test runners
- The combined `make test-other` pytest process OOM'd at 94%: microsimulation tests now run as their own CI job, and `test_microsim.py` builds one Microsimulation per dataset (module-scoped fixture; 2 builds instead of 4 — it was the real ~17-min hotspot of the old Rest job). The `RUN_HEAVY_TESTS` LSR/CG tests (skipped on CI) now subsample and share a baseline.
- `run_selective_tests.py` (Quick Feedback) packed every selected location into one process — reuniting the very files split above and OOMing under coverage. It now runs one location per subprocess (`coverage run -a` appends across invocations), bounding memory at the heaviest single location.

### 5. Job layout: 17 → 23 runners
Baseline: states-shard-1..4, irs, household, ssa-usda, rest-a, rest-b, contrib-hhs, reform. Contrib: states-shard-1..4, other-shard-1, 2a, 2b, 3, congress. Plus Partners, Rest (Python + variables), Microsimulation. Total runner-minutes are roughly unchanged — the compute is spread wider and shorter, with growth headroom on every runner.

## Measured results (run on commit 1d4e2b9)

All 35 checks green. Wall 26m27s (was ~33.7 min). Old→new highlights: Baseline rest 33.7 → 12.0∥10.6; Rest+microsim 29.8 → 11.8∥11.7; irs-household 23.6 → 14.5∥10.9; Partners 29.8 → 22.2; contrib other max 26.4 → 14.1; reform 9.6 → 9.4. Worst job: contrib states-shard-4 (26m27s — carries the RI per-file batches).

## Notes for maintainers

- **Required status checks need updating**: deleted `Baseline (states-ny/states-non-ny-*/rest/contrib/ssa/irs-household)`, `Reform (per-file)`, `Contrib (other-shard-1/2/3 as previously named)`; new names per the layout above.
- A natural follow-up (not in this PR): pack the states shards from a measured-duration cost ledger instead of alphabetical stride — contrib states-shard-4 carries a disproportionate content draw and sets the wall; balanced packing projects ~22-23 min walls.
- Every log now shows per-batch peak RSS: if any batch trends toward ~10 GB in future green runs, split it then — no more surprise OOMs.

## Test plan
- [x] 3 full CI runs on this PR; final run entirely green (35/35 checks)
- [x] Per-batch peak RSS verified ≤ ~8.5 GB across all jobs
- [x] NY fix outputs verified byte-identical before/after
- [x] Split test files verified case-count-identical (RI 5+4+2=11; CRFB 3+3=6)


hua7450:
**Round 2 pushed** — re-batched from the measured per-batch peak RSS that round 1's new logging produced.

Round 1 result: 9/15 heavy jobs passed with large wins (rest 33.7→10.4 min, Partners 29.8→18.0, irs-household 23.6→13.9), but 2-wide batch co-scheduling OOM'd four jobs: the RSS telemetry revealed single batches already peak at 12-14.6 GB on Linux (`ri/ctc_reform_test.yaml` = **14.6 GB as one file**; `contrib/states/oh` 12.4 GB; `gov/simulation` 12.1 GB) — the pre-existing layout was living at the cliff, which explains the historical intermittent "runner lost communication" failures.

Round 2 is headroom-first, per maintainer direction (long-term robustness to new tests > minimum runner count):
- `--workers 1` everywhere except Partners (max batch 2.5 GB measured)
- Finer batches so every subprocess stays ≤ ~8 GB: states `--batches 16` × 4 shards, household `--batches 4`, `gov/simulation` + `baseline/contrib/states` per-file, `oh` joins `ri` in per-file isolation
- The two files no batching can save are split by reform combo (cases relocated byte-identically): RI CTC → 3 files, CRFB payroll sweep → 2 files
- Microsimulation pytest gets its own job (the combined `test-other` process was OOM-killed at 94%)
- Quick Feedback selective cap 250→100 files (over-cap safely defers to full suites)

Now 23 full-suite runners (was 17 before this PR), each with ≥ ~8 GB of slack for future heavy tests, and every log permanently reports per-batch peak RSS + the 25 slowest cases so drift is visible before it breaks.

MaxGhenis:
Data point for this rebalance from tonight's merge queue: "Full Suite - Contrib (states-shard-2)" was killed (exit 143 / runner-canceled, ~17-26 min in) on at least nine separate PR runs across #8679, #8818, #8901 (twice), #8902, #8903 (twice), #8907, and #8908 — every one passed unchanged on rerun, so it's load/memory, not tests. That shard looks like the first candidate for whatever split this PR lands on.

MaxGhenis (approved):
Fable review + instrumented verification: the pinned-TBS cache replaces per-call full-system clones in the NY pre-TCJA CTC and pre-ARPA EITC formulas (12→2 clones, peak shard RSS 7.30→3.72 GB, byte-identical outputs), with weakref identity keying so reformed systems rebuild rather than leak baseline parameters. Shard rebalance proven exhaustive: 4,280/4,280 test files covered, zero orphans/dups. This addresses the exit-143 kills that hit ~20 runs in the last day. CI green.

PR #8895, opened 2026-07-05
Fix CTC value and EITC disabled dependent child count

### LINKED ISSUE #8893: CTC value can exceed actual allowed credit when nonrefundable and refundable portions bind differently
## Summary

`ctc_value` is documented as the actual value of the Child Tax Credit, but under current law it returns `min(ctc, ctc_phase_in)` when the CTC is not fully refundable. That can overstate or understate the actual credit value because it ignores the nonrefundable credit allowed by tax liability and the separately computed refundable CTC.

## Evidence

A Populace/Axiom oracle comparison found cases where Axiom and PSL Tax-Calculator agree, while PolicyEngine-US reports a different `ctc_value`:

- Two qualifying children, `ctc = 4,400`, `refundable_ctc = 3,400`, `ctc_limiting_tax_liability = 853.67`: actual value is `4,253.67`, but PE reports `4,400`.
- One other dependent credit case, `ctc = 500`, `ctc_phase_in = 0`, `refundable_ctc = 0`, large limiting tax liability: actual value is `500`, but PE reports `0`.

The current formula uses `ctc_phase_in` as the value cap. A better actual-value formula is `min(ctc, ctc_limiting_tax_liability + refundable_ctc)`.

## Impact

This primarily affects the reported `ctc_value` variable. Net `income_tax` currently uses `non_refundable_ctc` and `refundable_ctc` through the credit aggregates rather than `ctc_value` directly, so some cases have wrong reported CTC value while net tax is still correct.

### LINKED ISSUE #8894: EITC child count excludes permanently disabled adult dependents
## Summary

`eitc_child_count` counts `is_qualifying_child_dependent`, which intentionally only applies the age/student test under IRC 152(c)(3)(A). It does not include the IRC 152(c)(3)(B) exception for individuals who are permanently and totally disabled.

PolicyEngine-US already has `is_child_dependent`, whose documentation and formula include the permanently disabled pathway, but `eitc_child_count` does not use that path.

## Evidence

A Populace/Axiom oracle comparison found disabled adult dependent cases where Axiom and PSL Tax-Calculator count the dependent for EITC and PolicyEngine-US does not. In a representative case:

- dependent age: 29
- `is_tax_unit_dependent = true`
- `is_permanently_and_totally_disabled = true`
- `meets_eitc_identification_requirements = true`

PolicyEngine-US produces `is_child_dependent = true`, but `is_qualifying_child_dependent = false`, so `eitc_child_count = 0`. The resulting `eitc` and `income_tax` are too low/high respectively.

## Impact

This affects `eitc_child_count`, `eitc`, and downstream `income_tax` because `eitc` is included in `income_tax_refundable_credits`.

MaxGhenis:
## Summary

- calculate `ctc_value` as the actual allowed CTC value capped by tax liability plus refundable CTC
- include permanently and totally disabled tax-unit dependents in `eitc_child_count`
- add regression cases for the Axiom/TaxCalc oracle mismatches
- repin 41 downstream `ctc_value` assertions that encoded the old `min(ctc, ctc_phase_in)` behavior (analytics-coverage signatures for CO/IL/MA/NC/TX, EITC/CDCC/composition edge cases, OK integration tests); the OK integration case now matches `ok_federal_ctc` (93.26), which already modeled the liability-capped federal CTC

Closes #8893
Closes #8894

## Tests

- `uv run python -m policyengine_core.scripts.policyengine_command test policyengine_us/tests/policy/baseline/gov/irs/credits/ctc/refundable/ctc_value.yaml -c policyengine_us`
- `uv run python -m policyengine_core.scripts.policyengine_command test policyengine_us/tests/policy/baseline/gov/irs/credits/earned_income/eitc_child_count.yaml -c policyengine_us`
- ran all 26 test files that reference `ctc_value` or set `is_permanently_and_totally_disabled` (partners analytics coverage, my_friend_ben, NY/OK/KS/PA state tests): 152 passed
- `uv run ruff format` / `uv run ruff check` on the changed variables
- `git diff --check`

### DIFF
--- a/policyengine_us/variables/gov/irs/credits/ctc/ctc_value.py
+++ b/policyengine_us/variables/gov/irs/credits/ctc/ctc_value.py
@@ -11,8 +11,6 @@ class ctc_value(Variable):
 
     def formula(tax_unit, period, parameters):
         ctc = tax_unit("ctc", period)
-        p = parameters(period).gov.irs.credits.ctc.refundable
-        if not p.fully_refundable:
-            phase_in = tax_unit("ctc_phase_in", period)
-            return min_(ctc, phase_in)
-        return ctc
+        limiting_tax = tax_unit("ctc_limiting_tax_liability", period)
+        refundable_ctc = tax_unit("refundable_ctc", period)
+        return min_(ctc, limiting_tax + refundable_ctc)
--- a/changelog.d/fix-ctc-value-eitc-disabled.fixed.md
+++ b/changelog.d/fix-ctc-value-eitc-disabled.fixed.md
@@ -1,0 +1,1 @@
+Fixed the reported federal Child Tax Credit value and EITC child count for permanently disabled adult dependents.
--- a/policyengine_us/tests/policy/baseline/gov/irs/credits/ctc/refundable/ctc_value.yaml
+++ b/policyengine_us/tests/policy/baseline/gov/irs/credits/ctc/refundable/ctc_value.yaml
@@ -82,3 +82,25 @@
     refundable_ctc: 4_000
     non_refundable_ctc: 0
     ctc_value: 4_000
+
+- name: Case 5, actual CTC value is capped by liability plus refundable credit.
+  period: 2026
+  absolute_error_margin: 0.01
+  input:
+    ctc: 4_400
+    ctc_limiting_tax_liability: 853.67
+    ctc_phase_in: 5_735
+    refundable_ctc: 3_400
+  output:
+    ctc_value: 4_253.67
+
+- name: Case 6, other dependent credit is valuable even with no CTC phase-in.
+  period: 2026
+  absolute_error_margin: 0.01
+  input:
+    ctc: 500
+    ctc_limiting_tax_liability: 10_133.92
+    ctc_phase_in: 0
+    refundable_ctc: 0
+  output:
+    ctc_value: 500
--- a/policyengine_us/tests/policy/baseline/gov/irs/credits/earned_income/eitc_child_count.yaml
+++ b/policyengine_us/tests/policy/baseline/gov/irs/credits/earned_income/eitc_child_count.yaml
@@ -51,3 +51,37 @@
         members: [person1, person2, person3]
   output:
     eitc_child_count: 1
+
+- name: Case 4, disabled adult dependent meets the EITC child count.
+  period: 2026
+  input:
+    people:
+      person1:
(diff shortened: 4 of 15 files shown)


policyengine_us/variables/gov/irs/credits/ctc/ctc_value.py
```diff
@@ -11,8 +11,6 @@ class ctc_value(Variable):
 
     def formula(tax_unit, period, parameters):
         ctc = tax_unit("ctc", period)
-        p = parameters(period).gov.irs.credits.ctc.refundable
-        if not p.fully_refundable:
-            phase_in = tax_unit("ctc_phase_in", period)
-            return min_(ctc, phase_in)
-        return ctc
+        limiting_tax = tax_unit("ctc_limiting_tax_liability", period)
```
PavelMakarchuk, line 14: @MaxGhenis was this change intended? Does not meet PR scope and this might throw off any refundable CTC reform options

MaxGhenis: Intended — it's the core of #8893 rather than scope creep: `ctc_value` is documented as the *actual* value of the credit, and the old `min(ctc, ctc_phase_in)` misreported it in both directions (e.g., $4,400 for a zero-liability family whose true value is $3,400; $0 for an ODC-only filer with ample liability whose true value is $500). The new formula is algebraically `min(non_refundable_ctc, liability) + refundable_ctc`, since `non_refundable_ctc = ctc − refundable_ctc`.

On fully-refundable reform options, behavior is unchanged where it should be and more correct where it differs:

- Under a fully-refundable regime with the refundable maximum at the full credit (ARPA baseline and the standard reform), `refundable_ctc = ctc`, so `min(ctc, liability + ctc) = ctc` — identical to the old branch. Verified on current main for 2021 at $0/$10k/$40k earnings: `ctc = refundable_ctc = ctc_value = 7,200` in all three.
- Under a hybrid reform (`fully_refundable` on but a lower refundable cap), the old branch reported the full `ctc` as the value even when the nonrefundable remainder exceeded liability — an overstatement; the new formula reports what the family actually receives. That helps reform analysis rather than throwing it off.

The Reform (per-file) CI shard, which runs the refundable-CTC reform suites, passed on this PR, and `ctc_value` has no formula consumers (it's a reporting variable for the household app), so nothing downstream shifts besides the displayed value.

PR #9043, opened 2026-07-15
[2/4] Count reported receipt (receives_X) in state and local receipt-proxy rules

hua7450:
## Summary

PR 2 of the 4-PR receipt-vs-eligibility series (see #9041 for the contract). Wires the `receives_ssi` / `receives_tanf` / `receives_snap` reported-receipt booleans into the **44 remaining receipt-proxy consumers** — every rule that keys on benefit *receipt* (`X > 0` or `X == 0`) rather than benefit *amount* — so someone who reports receiving a program the simulation computes as $0 is still treated as a recipient.

> #9041 has merged and `upstream/main` is pulled into this branch, so the diff now shows only this PR's own changes.

## Commits (one per consumer family)

1. **Federal** — SNAP student exemption (reported TANF satisfies the statutory exception); Medicaid community engagement pass-through (reported SNAP/TANF).
2. **State SSI supplements** — KS SSPP, WA SSP, GA SSP, MI SSP, IN RCAP + SAPN, FL OSS, MD PAA, NE AABD, ID AABD, MN MSA resource test.
3. **TANF assistance-unit exclusions** — the exclusionary direction: reported SSI receipt now removes people from TX TANF categorical eligibility, NC TANF household sizing, and MT/IL TANF payment eligibility, matching how these rules treat computed SSI.
4. **State tax credits, childcare, county GA** — ID grocery credit + permanent building fund, IL I-PASS Assist, AZ property tax credit, NM CDCC, LA CCAP (age + special-needs rules), MO CCS special needs, MS CCPP minimum-fee category, WA Birth-to-Three ECEAP, and the California county General Assistance programs (Alameda, Contra Costa, Marin, Orange, Riverside, Santa Clara, San Francisco, San Mateo).
5. **County GA income, DC/MD/MI childcare, and DC tax** — Santa Clara GA countable-income exclusion, DC CCSP qualified need, DC disabled exclusion subtraction, MD CCS eligibility + weekly copay, MI CCAP income waiver.
6. **Review response** — Marin GR test, three missed TANF consumers, federal `tanf` in external consumers (details below).

Every site keeps PR 1's shape: `receives_X` is OR'd with (or negated against) the computed signal — it can only add receipt, never remove it; switching receipt off remains the take-up flags' job. All defaults are unchanged (`receives_X` defaults to false).

## Response to review (2026-07-21)

[...]

PR #9041, opened 2026-07-15
[1/4] Add X_if_takes_up outputs and receives_X inputs for receipt-vs-eligibility handling (SSI, TANF, SNAP, WIC)

hua7450:
## Summary

Implements the consistent reported-vs-eligible receipt contract requested by an API partner, one commit per program (SSI → TANF → SNAP → WIC), plus refinement commits.

**This is PR 1 of a 4-PR series:**
1. **This PR** — core contract: `X_if_takes_up` outputs, take-up through the API, `receives_X` inputs, federal categorical consumers.
2. **#9043** — state/local `receives_X` sweep — wire reported receipt into the ~25 remaining receipt-proxy consumers (state SSPs, county GA, TANF assistance-unit rules, SNAP student exemption).
3. `applicable_ssi` deprecation (PR not yet opened) — migrate its consumers to read `ssi` directly, retire `ssi_reported`/`use_reported_ssi`, and fold the Medicaid SSI-category reads into the same single reported-receipt mechanism (requires partner coordination).
4. **#9044** — `is_tanf_enrolled` → `receives_tanf` consolidation — derives `is_tanf_enrolled` from `receives_tanf` by default, one canonical "currently on TANF" input across its ~70 consumers.

For any program X that feeds other programs' calculations:
- **Reports receiving X, amount known** → send it as the `X` input (annual inputs on monthly variables spread automatically). Flows to all downstream consumers automatically.
- **Reports receiving X, amount unknown** → leave `X` computed (today's default).
- **Reports not receiving X** → send `takes_up_X_if_eligible: false`; the simulated amount stops counting as income downstream and receipt-conferred categorical eligibility switches off. Flows to all downstream consumers automatically.
- **Read `X_if_takes_up` in the same request** for the computed entitlement — what they'd be eligible for if they applied. This is the output partners should request when they want our computed value while overriding receipt.
- **`receives_X` booleans** cover the case where receipt and the model disagree: someone reports receiving X but the simulation computes $0, and categorical eligibility should still fire.

## Per-program changes

| Program | `X_if_takes_up` | Take-up through API | `receives_X` input | Categorical consumers wired |
|---|---|---|---|---|
| SSI | `ssi_if_takes_up` (new) | already worked (`takes_up_ssi_if_eligible`) | `receives_ssi` (new) | SNAP categorical, Head Start, Medicaid 1634/SSI-criteria + 209(b) |
[...]

PR #9044, opened 2026-07-15
[4/4] Derive is_tanf_enrolled from receives_tanf by default

hua7450:
## Summary

PR 4 of the 4-PR receipt-vs-eligibility series (see #9041 for the contract). Consolidates the two overlapping "currently on TANF" booleans so one input drives everything.

> **Stacked on #9041** (which introduces `receives_tanf`) — the diff shows PR 1's commits until it merges, then this reduces to a single commit. Independent of #9043; can merge in either order after PR 1.

## The problem

`is_tanf_enrolled` (used by ~70 state TANF/CCAP applicant-vs-recipient rules and the SNAP work-registration exemption) and `receives_tanf` (the partner-facing reported-receipt input from #9041) represent the same fact. As two independent inputs they allow inconsistent states — a household marked as receiving TANF but not enrolled.

## The fix

`is_tanf_enrolled` gets a one-line formula defaulting to `receives_tanf`. Consequences:

- **Partners set one boolean.** `receives_tanf: true` now drives both cross-program categorical eligibility and recipient-tier rules (income limits, disregards, copays) in state programs.
- **Full backward compatibility.** Anyone sending `is_tanf_enrolled` directly (including all existing YAML tests and partner requests) overrides the derivation — verified: `receives_tanf: true` + `is_tanf_enrolled: false` still yields false.
- **No default change.** Both variables default to false; zero hits for `is_tanf_enrolled` in policyengine-us-data, so microsimulation is unaffected.

None of `is_tanf_enrolled`'s ~70 consumers are touched.

## Verification

- [x] Household simulations: default false / derived true / explicit override wins
- [ ] CI passes
- [ ] Rebase onto main after #9041 merges, mark ready for review


PavelMakarchuk:
# Program Review — PR #9044 (automated /review-program)

## Scope

- **Series context:** PR 4 of a 4-PR stacked series. The GitHub diff is stacked and includes the earlier parts, which were reviewed on PRs #9041 and #9043 — those findings are not re-reported here. This review covers only the **#9044 delta**:
  - `is_tanf_enrolled.py` gains a formula (`return spm_unit("receives_tanf", period)`)
  - Six YEAR-defined state consumers migrated to `period.first_month` reads: `ca_tanf_financial_eligible`, `co_tanf_countable_earned_income_need`, `co_tanf_countable_gross_unearned_income`, `nc_tanf_countable_gross_unearned_income`, `nh_fanf_resources_eligible`, `tx_ceap_eligible`
  - New test `is_tanf_enrolled.yaml`; changelog fragment `tanf-enrolled-consolidation.changed.md`
- **Notably, this PR resolves the `receives_tanf` / `is_tanf_enrolled` duplication flagged in the PR #9041 review** — the enrollment flag now derives from reported receipt by default while remaining overridable as an input.
- **PDF audit:** skipped (no parameter changes)
- **CI:** all 33 checks passing
- **Partner tests:** no files under `tests/policy/baseline/partners/` touched; the one partner test that uses `is_tanf_enrolled` (`edge_cases/federal/cash/tanf/federal.yaml`) supplies it as an input, and the input-override behavior is preserved and verified — no partner-facing risk.

## Critical

None. Three independent validators (regulatory/semantics, code patterns, test coverage) each returned zero critical findings.

## Should Address

1. **Confirm the policyengine-us-data side of the consolidation.** The new formula only changes results when a dataset sets `receives_tanf` but **not** `is_tanf_enrolled` (a dataset-supplied `is_tanf_enrolled` still overrides the formula, and `receives_tanf` is brand-new so no existing dataset populates it — meaning no microsim change today). Confirm with the companion policyengine-us-data change which of the two variables the enhanced CPS will populate, so the derivation does not silently flip enrollment for any SPM unit later. A microsim smoke run (CalWORKs, CO/NC TANF income, NH FANF, TX CEAP) alongside the data-repo update would close this. Not checkable from this repo.

2. **The derived path (`receives_tanf` → `is_tanf_enrolled` → state consumer) is never exercised end-to-end.** `is_tanf_enrolled.yaml` tests the formula in isolation, but all six consumer test files set `is_tanf_enrolled` directly as an input, which overrides the formula — so no test proves a YEAR consumer responds to `receives_tanf`. Risk is low (one-line pass-through formula), but one consumer-level case (e.g., in `nc_tanf_countable_gross_unearned_income.yaml` or `co_tanf_countable_gross_unearned_income.yaml`: set `receives_tanf: true` with no `is_tanf_enrolled` input at a YEAR period and assert child support is excluded) would pin the integration.

## Suggestions

1. **`ok_stc` any-month vs first-month semantics.** `ok_stc.py:70` is a seventh YEAR-defined consumer, intentionally left unchanged because `add(tax_unit, period, ["is_tanf_enrolled"]) > 0` aggregates safely over the 12 months (no crash). It now carries "enrolled in any month" semantics while the six converted consumers use "enrolled in January." Both are defensible; a one-line confirmation that this divergence is intended for the OK sales tax credit would be worthwhile.
2. **Add a YEAR-level input-override test case.** The new test proves the override at MONTH (2025-01). A case supplying `is_tanf_enrolled` at a YEAR period would lock in the dispatch+override behavior the six YEAR consumers actually rely on. Optional polish: a reverse-override case (`is_tanf_enrolled: false`, `receives_tanf: true` → `false`) proving the input wins outright rather than OR-ing with the formula.
3. **Document the January approximation in the consumers.** For a household enrolled Feb–Dec but not January, the six annual variables now treat the whole year as not enrolled. The YAML framework cannot express partial-year enrollment (only `20xx` / `20xx-01` periods), so this edge is effectively untestable at unit level — a one-line code comment in each consumer noting "annual enrollment approximated by January status" would document the choice.
4. **`documentation` vs `reference` field.** `is_tanf_enrolled` (re)uses `documentation`; the variable-patterns skill prefers `reference` with a URL. Acceptable for an internal derived flag with no statutory citation — noted only because the PR touches that line.
5. **`tx_ceap_eligible` sibling reads (future cleanup).** Within the same formula, `is_snap_eligible` and `is_ssi_eligible` (also MONTH-defined) are still read at plain `period` while `is_tanf_enrolled` now uses `period.first_month`. Pre-existing inconsistency, not introduced by this PR; those siblings have no formula-vs-input transition forcing a change.

## Verified Correct

- **Scope completeness (enumerated, not sampled).** All 65 references to `is_tanf_enrolled` were enumerated. Exactly 6 are YEAR-defined variables reading it directly at plain `period` — precisely the 6 the PR converts. The other 58 direct consumers are MONTH-defined (including all `person.spm_unit(...)` readers: ak_atap, dc_tanf, in_tanf ×2, in_ccdf, nd_ccap, id_iccp, medicaid_community_engagement) and need no change. No YEAR direct reader was missed; the conversion is also *required* (a YEAR read of a MONTH-with-formula variable at plain `period` would now raise), not merely stylistic.
- **Input-override contract preserved.** `is_tanf_enrolled` is a bool → STOCK → `set_input_dispatch_by_period`, chosen independently of whether a formula exists (verified empirically in policyengine-core). A supplied input at MONTH or YEAR dispatches to all 12 months and wins over the formula. Proven by new-test Case 3 (`is_tanf_enrolled: true`, `receives_tanf: false` → `true`) and by `build_from_dict` holder inspection for YEAR inputs. The docs notebook and the partner federal.yaml, both of which supply `is_tanf_enrolled` as input, keep working unchanged.
- **No circular dependency.** Chain is `tanf` → state TANF → `is_tanf_enrolled` → `receives_tanf` (leaf input, no formula). No back-edge.
- **`period.first_month` is the codebase-standard idiom** for YEAR formulas reading MONTH variables (per period-patterns; 14 existing usages, e.g., `meets_snap_asset_test.py`, `meets_wic_income_test.py`).
- **Regression evidence.** All six consumer test files are unchanged and continue to pass: a directly-set MONTH bool input resolves identically at `period.first_month` as under the old YEAR read, so the period switch is behavior-preserving for every scenario those tests cover.
- **No behavior drift when neither variable is set** (both default false), and no microsim change until the data repo populates `receives_tanf`.
- **Hygiene:** no hard-coded values; entity/period alignment clean (SPMUnit/MONTH → SPMUnit/MONTH); no `adds`/formula mixing; naming consistent; no references in `reforms/` or `programs.yaml`; changelog fragment valid (`changed`, top-level, matches branch).
- **Resolves the `receives_tanf`/`is_tanf_enrolled` split flagged on PR #9041.**

## Validation Summary

| Check | Result |
|---|---|
| Regulatory/semantics validator | PASS — 0 critical; override mechanics verified empirically; no circularity |
| Code-pattern validator | PASS — 0 critical; 6-consumer scope proven exactly complete via enumeration of all 65 references |
| Test-coverage validator | PASS — 0 critical; new test covers default/derivation/override; unchanged consumer tests are valid regression evidence |
| Partner contract tests | Untouched; input-override path preserved for `federal.yaml` |
| PDF audit | Skipped (no parameter changes) |
| Changelog fragment | Valid (`tanf-enrolled-consolidation.changed.md`) |
| CI | All 33 checks passing |

## Review Severity: COMMENT

No critical or blocking findings — the delta is correct, complete in scope, and well-tested at the unit level. COMMENT rather than APPROVE because the two should-address items merit author action around merge time: the data-repo coordination (item 1) determines whether the consolidation ever changes microsim results and should be confirmed before or alongside merge, and the end-to-end derived-path consumer test (item 2) is a cheap addition that closes the only real coverage gap. Neither requires code changes to this PR's logic.

## Next Steps

To auto-fix issues: `/fix-pr 9044`

PR #9113, opened 2026-07-21
Apply basic income disability supplement to blind people

### LINKED ISSUE #3523: Disability UBI supplement not applying to household simulations
For example, this scenario has a $6,186 disability UBI, but the blind individual only receives the adult amount @nikhilwoodruff https://policyengine.org/us/household?reform=42434&focus=householdOutput.netIncome&region=enhanced_us&timePeriod=2023&baseline=2&household=38407

DTrim99:
Fixes #3523.

The basic income disability supplement in `basic_income_before_phase_out` keyed off `is_ssi_disabled`, which does not consider blindness. Blindness is only captured in the microdata (set directly in `cps.py`), so in **household** simulations a blind person was denied the disability supplement — as reported in #3523.

**Change:** base the supplement count on `is_ssi_disabled | is_blind`. `is_blind` is a settable person-level input, so it now works directly on the household page (no `ssi_reported > 0` workaround needed).

**Tests** (appended to `basic_income_before_phase_out.yaml`):
- a blind, non-`ssi_disabled` adult receives the disability supplement;
- a blind **and** disabled adult is not double-counted (the OR yields one supplement).

Verified locally: 9/9 in the file and 17/17 across the `ubi_center` tests pass.

### DIFF
--- a/changelog.d/ubi-disability-supplement-blind.fixed.md
+++ b/changelog.d/ubi-disability-supplement-blind.fixed.md
@@ -1,0 +1,1 @@
+Applied the basic income disability supplement to blind people in household simulations by including is_blind alongside is_ssi_disabled.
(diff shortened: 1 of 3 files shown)


hua7450:
The fix looks right for this variable. More broadly, #3523's root cause is a long-known issue on our radar — it can exist anywhere a formula uses `is_disabled`/`is_ssi_disabled` for a program whose statute also covers blindness.

hua7450 (approved):

PR #8961, opened 2026-07-08
Make SNAP work requirement disqualification person-level

daphnehanse11:
Fixes #8862

## Summary

Issue #8862 identified two misreadings of SNAP work requirement law. The first — household-wide disqualification when any member fails — was largely resolved by #8372, which made `snap_unit_size` remove failing individuals, added income proration, and gave the household-level variable any-member semantics. This PR completes the issue's scope by fixing the second misreading: the general work requirement treated working 30+ hours/week as an affirmative requirement, disqualifying every non-exempt adult who simply was not working. It also corrects the Medicaid community engagement pass-through, whose SNAP prong the interim final rule defines as a status test rather than a compliance test.

## Regulatory authority

Verified against primary sources (eCFR, U.S. Code, enrolled H.R. 1, Federal Register):

- **7 U.S.C. 2015(d)(1)(A), (o)(2); 7 CFR 273.7(f)(1), 273.24(b)**: work-requirement ineligibility attaches to the *individual*, who "will be considered an ineligible household member." The only household-wide sanction is a state option applying solely when the *head of household* fails, capped at 180 days (7 U.S.C. 2015(d)(1)(B); 7 CFR 273.7(f)(5)); no household-wide sanction exists for the ABAWD time limit.
- **7 CFR 273.7(b)(1)(vii)**: working 30+ hours/week is an *exemption from work registration*, not an affirmative requirement. A non-exempt registrant who is not working remains eligible unless they affirmatively fail to comply without good cause — refusing suitable employment, E&T noncompliance, or voluntary quit/reduction of work effort under 273.7(j). These "without good cause" state-agency determinations are unobservable in survey data.
- **7 CFR 273.11(c)**: (c)(1) counts work-sanctioned members' income in full; (c)(2) prorates ABAWD time-limit-ineligible members' income. Both remove the member from household size for the benefit calculation.
- **SSA § 1902(xx)(9)(A)(ii)(VI)(bb)** (added by H.R. 1 § 71119) excludes from Medicaid community engagement an individual who "is a member of a household that receives [SNAP] benefits ... and is not exempt from a work requirement under such Act." **42 CFR 435.554(c)(7)** (CMS-2454-IFC, 91 FR, June 3, 2026) confirms: "States only need to determine that the individual is not exempt from SNAP work requirements and is in a household that receives SNAP benefits; the State does not need to confirm that the individual is in fact compliant with SNAP work requirements" — an explicit asymmetry with the TANF prong, which does require compliance.

## Changes

[...]

PR #9118, opened 2026-07-22
Fix SNAP eligibility conflating student and immigration filters across members

### LINKED ISSUE #8133: SNAP eligibility incorrectly passes when student and immigration filters are satisfied by different household members
## Summary

`is_snap_eligible` splits the "not an ineligible student" and "immigration-eligible" per-person filters into two independent `spm_unit.any(...)` calls. This allows a household to be flagged eligible when *different* members satisfy each filter, even though *no single person* satisfies both.

Per **7 CFR 273.1(b)**, **273.4**, and **273.5**, each per-person exclusion must be evaluated in conjunction — a household needs at least one member who passes *all* per-person filters simultaneously.

## Location

`policyengine_us/variables/gov/usda/snap/eligibility/is_snap_eligible.py` (lines 24–35)

```python
eligible_person_present = spm_unit.any(
    ~person("is_snap_ineligible_student", period)
)
...
immigration_eligible_member_present = spm_unit.any(
    person("is_snap_immigration_status_eligible", period)
)
return (
    (normal_eligibility | categorical_eligibility)
    & eligible_person_present
    & work_requirements_eligibility
    & immigration_eligible_member_present
)
```

## Reproducing scenario

A 2-person SPM unit that passes income/asset tests:

- **Person A**: ineligible student (e.g., half-time+ college, no exemption), US citizen
- **Person B**: non-student adult, non-qualified noncitizen

Current logic:

| Check                                                               | Result        | Satisfied by |
| ------------------------------------------------------------------- | ------------- | ------------ |
| `eligible_person_present` (`any(~ineligible_student)`)              | `True`        | Person B     |
| `immigration_eligible_member_present` (`any(immigration_eligible)`) | `True`        | Person A     |
| Final `is_snap_eligible`                                            | **`True`** ❌ | —            |

No single member passes both per-person filters, so the correct result is **`False`**.

## Expected behavior

Household eligible **only if at least one person individually passes every per-person exclusion filter**:

```python
has_eligible_member = spm_unit.any(
    ~person("is_snap_ineligible_student", period)
    & person("is_snap_immigration_status_eligible", period)
)
return (
    (normal_eligibility | categorical_eligibility)
    & has_eligible_member
    & work_requirements_eligibility
)
```

## Regulatory references

- **7 CFR 273.1(b)** — household composition; excluded members are not treated as household members for benefit purposes.
- **7 CFR 273.4** — ineligible noncitizens (per-person exclusion).
- **7 CFR 273.5** — ineligible students (per-person exclusion).

## Impact

Overstates SNAP eligibility for mixed-composition households — most notably mixed-status families with a citizen who is an ineligible student and a non-citizen non-student adult. Affects both single-household `Simulation` queries and microsimulation estimates of SNAP caseload and cost.

## Suggested tests

Add YAML tests under `policyengine_us/tests/policy/baseline/gov/usda/snap/`:

1. **Should be ineligible**: ineligible-student citizen + non-qualified-noncitizen non-student, otherwise meeting income/asset tests — assert `is_snap_eligible: false`.
2. **Should be eligible** (regression guard): ineligible-student citizen + qualified-immigrant non-student, otherwise meeting tests — assert `is_snap_eligible: true`.
3. **Should be eligible** (regression guard): non-student citizen alone, otherwise meeting tests — assert `is_snap_eligible: true`.

## Related

- #4962 — Exclude students from the SNAP unit if they are students (addresses student exclusion generally, but not the cross-filter conjunction).
- #6805 — Separate vectorization-bug pattern (`np.any()` missing `axis`); structurally distinct from this logical-conjunction bug but thematically related as an eligibility-aggregation correctness issue.

## Labels

`bug`, `snap`

DTrim99:
Fixes #8133.

`is_snap_eligible` tested the student rule (`~is_snap_ineligible_student`) and the immigration rule (`is_snap_immigration_status_eligible`) with **two independent** `spm_unit.any()` reductions, then AND-ed the results. This incorrectly passed a SNAP unit when **member A** satisfied the student rule and a **different member B** satisfied the immigration rule — even though no single member satisfied both.

**Fix:** fuse into one member-level `any()` of the conjunction:
```python
eligible_member_present = spm_unit.any(
    ~person("is_snap_ineligible_student", period)
    & person("is_snap_immigration_status_eligible", period)
)
```

**Test:** added a split case (student-eligible-but-undocumented member + citizen-but-ineligible-student member → `is_snap_eligible: false`) and a happy-path case (one member is both → `true`). SNAP dir: 472/472 pass.


hua7450 (approved):
Verified the fix and the two new tests against 7 USC 2015(e)/(f) — fusing the student and immigration filters into a single member-level `any()` is correct, and the split-member regression case reproduces #8133 exactly.

Pushed one small commit on top:
- Read `is_snap_ineligible_student` (YEAR-defined) with `period.this_year`, matching all six other call sites of this variable.
- Cited 7 USC 2015(e)/(f) in the same-member comment.
- Merged current upstream `main` into the branch.

One related issue found during review, deliberately **not** addressed here: `meets_snap_work_requirements` is still a separate `spm_unit.any()`, so the member passing the fused student+immigration check need not be the member meeting work requirements — the same conflation class this PR fixes (e.g., an immigration-ineligible full-time worker + a citizen ABAWD failing the time limit still passes, and a fully-excluded unit of size 0 still draws the minimum allotment). The clean fix is `spm_unit.any(~is_snap_excluded_member)`, which already encodes the three-way per-person conjunction, but since #8961 is actively rewriting work-requirement semantics, that belongs in a follow-up after it lands.

DTrim99:
Thanks @hua7450 — appreciate the careful review and the commit on top. All confirmed on my end:

- **`period.this_year` on `is_snap_ineligible_student`** — agreed, that's the right convention. It's YEAR-defined and read inside a MONTH-period formula, so `period.this_year` matches the six other call sites (including `is_snap_excluded_member` itself). Good catch.
- **`7 USC 2015(e)/(f)` citation** on the fused-member comment — 👍.

On the deferred **work-requirement conflation**: fully agree, and it's a real one — the same cross-member class, one filter over. Since `meets_snap_work_requirements` is a separate `spm_unit.any()`, the member passing student + immigration need not be the member meeting work requirements. And as you note, `is_snap_excluded_member` already encodes the full three-way per-person disjunction, so the clean fix collapses to a single `spm_unit.any(~is_snap_excluded_member)` — simpler than the current two-`any()` structure. Deferring behind #8961 (which is rewriting work-requirement semantics) is the right call; swapping now would risk colliding with that.

I've captured it as a tracked follow-up so it isn't lost: **#9145** (blocked on #8961, with the failure cases and acceptance criteria, credited to you).

This PR is good to merge as-is once CI settles.

Issue #626, opened 2022-03-17
98% tax MTR for single parent of one with income between $19,700 and $22,800


MaxGhenis:
https://policyengine.org/us/household, single with one child. Seems wrong.
<img width="842" alt="image" src="https://user-images.githubusercontent.com/6076111/158733322-07709d02-1a08-4258-8f03-9e5bbb26ddf4.png">

For comparison, Tax-Cruncher shows more like a 42% MTR at this range for the same household type: https://compute.studio/PSLmodels/Tax-Cruncher/871/
<img width="630" alt="image" src="https://user-images.githubusercontent.com/6076111/158735045-ef602099-3742-4c8b-b52c-b75b184e51de.png">

MaxGhenis:
This was also the AMT bug, fixed now![Screenshot_20220317-065638.png](https://user-images.githubusercontent.com/6076111/158824751-50293c93-7959-4283-bb6d-24ca79ed6431.png)

PR #9261, opened 2026-08-10
Default weekly_hours_worked_before_lsr to 0 to harden SNAP work requirements (#9254)

### LINKED ISSUE #9254: weekly_hours_worked_before_lsr default of 40 silently passes all SNAP work requirements when hours data is missing
## Summary

`weekly_hours_worked_before_lsr` is a pure input with `default_value = 40` and no fallback formula (`policyengine_us/variables/household/income/person/weekly_hours_worked.py`). Forty hours passes both the SNAP 20-hour ABAWD test (`meets_snap_abawd_work_requirements.py`) and the 30-hour general work-requirement test, so **any dataset or household that fails to populate this input silently behaves as if everyone works full-time and all hours-conditioned work requirements pass** — a false negative that looks like a populated model rather than an error.

This is the same failure mode behind the populace constant-40 landmine (populace #613, #626, #242/#248): whenever the input is absent or defaulted, SNAP work-requirement reform exposure collapses to ~zero with no visible signal.

## Why 40 is the worst possible default

The real population distribution (CPS/ACS) is mean ~19 weekly hours with ~50% zeros. A default of 40 is not a neutral placeholder — it sits above every work-requirement threshold, so missing data maps precisely onto "exempt / compliant." We repeatedly hit this: a fixture and multiple dataset builds all showed `meets_abawd_work_requirement = TRUE` for 0-hour adults purely because the input defaulted to 40.

## Proposed hardening (pick one)

1. **Change `default_value` to 0.** Missing data then reads as "no work hours" — a visible, conservative failure that surfaces immediately in any work-requirement result, rather than silently exempting everyone. Datasets that carry real hours are unaffected.
2. **Add a fallback formula** deriving `weekly_hours_worked_before_lsr` from `hours_worked_last_week` (already a populated input on the affected builds) when the usual-weekly input is absent. Inputs still override the formula, so real data is untouched; only the missing-data case changes.

Option 1 is the smaller, safer change and is my recommendation; option 2 preserves more signal where `hours_worked_last_week` exists.

## Guard (optional, complementary)

A test asserting the model does not treat a fully-defaulted hours column as universal WR compliance would catch regressions of this class.

Refs: populace #613, #626, #242, #248.

DTrim99:
Closes #9254.

## Problem
`weekly_hours_worked_before_lsr` defaulted to **40** (full-time). Any household or dataset that failed to populate it therefore read as working full-time and **silently cleared both** the 20-hour SNAP ABAWD test and the 30-hour general work-requirement test. That turns work-requirement reforms into a no-op whenever the hours column is missing or degenerate — the failure mode that hid a constant-40 hours column in a populace build (PolicyEngine/populace#613, #626) and made a hand-built fixture read `meets_abawd_work_requirement = true` for 0-hour adults.

Because 40 sits above every hours threshold, missing data was indistinguishable from "compliant" — it failed silently.

## Change
Default `weekly_hours_worked_before_lsr` to **0**. Missing hours now read as "no hours" and fail the work-hours test loudly instead of passing.

- Datasets and tests that provide the input are **unaffected** (input overrides the default), so the microsimulation on the populated national dataset does not change.
- Only situations/tests that omit hours change — they now correctly show a non-working person.

## Tests added
- `weekly_hours_worked.yaml`: `weekly_hours_worked_before_lsr` defaults to 0 when unspecified.
- `meets_snap_abawd_work_requirements.yaml` (Case 59): a working-age, non-exempt adult with no hours input now **fails** the ABAWD hours test.

## Draft — why
I could not run the full YAML suite locally (environment constraint). Opening as draft so CI can surface any existing unit tests that implicitly relied on the old 40-hour default (i.e. tests with working-age adults in hours-conditioned programs that omit hours). I'll update those to set hours explicitly once CI reports. Flagging the reviewer decision: default `0` (this PR, per the issue's recommendation) vs. a fallback formula from `hours_worked_last_week` — happy to switch approaches.

### DIFF
--- a/changelog.d/harden-snap-wr-hours-default.fixed.md
+++ b/changelog.d/harden-snap-wr-hours-default.fixed.md
@@ -1,0 +1,1 @@
+Default weekly_hours_worked_before_lsr to 0 rather than 40 so that missing hours data fails the SNAP work-requirement hours tests loudly instead of silently satisfying them.
--- a/policyengine_us/tests/policy/baseline/gov/hhs/tanf/non_cash/meets_tanf_non_cash_gross_income_test.yaml
+++ b/policyengine_us/tests/policy/baseline/gov/hhs/tanf/non_cash/meets_tanf_non_cash_gross_income_test.yaml
@@ -20,6 +20,7 @@
   absolute_error_margin: 0.05
   input:
     employment_income: 30_600 # 2550 monthly
+    weekly_hours_worked_before_lsr: 40
     state_code: NC
   output:
     snap_gross_income: 2_550
--- a/policyengine_us/tests/policy/baseline/gov/states/ak/dpa/ccap/benefit/ak_ccap.yaml
+++ b/policyengine_us/tests/policy/baseline/gov/states/ak/dpa/ccap/benefit/ak_ccap.yaml
@@ -169,6 +169,7 @@
     people:
       parent:
         age: 30
+        weekly_hours_worked_before_lsr: 40
         immigration_status: CITIZEN
         # A small AK fishing operation has a bad season — net loss of $30K.
         self_employment_income: -30_000
--- a/policyengine_us/tests/policy/baseline/gov/states/az/hhs/ccap/az_ccap.yaml
+++ b/policyengine_us/tests/policy/baseline/gov/states/az/hhs/ccap/az_ccap.yaml
@@ -5,6 +5,7 @@
     people:
       person1:
         age: 30
+        weekly_hours_worked_before_lsr: 40
       person2:
         age: 0
         childcare_attending_days_per_month: 20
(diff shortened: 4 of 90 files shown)


hua7450:
Pushed fbdcea6 removing `weekly_hours_worked_before_lsr: 40` from three dependent children that the blanket test migration had swept up (NJ CCAP edge cases 19–20, WV CCAP edge case 14). Work-hours inputs should sit on adults — the tax unit head or spouse — not on dependents. The removals are output-neutral: `nj_ccap_eligible_child` never reads hours, and `wv_ccap_activity_eligible` applies its hours test only to head/spouse members; both files pass locally (88/88).

I audited all 562 hours additions in this PR for the same issue: no person under 18 received hours anywhere, and these three (ages 18–19, `is_tax_unit_dependent: true`) were the only dependent assignments. The expected outputs are byte-identical across all 88 test files, including the four partner contract fixtures.

This PR looks good to me. Holding merge until Thursday (Aug 13).

hua7450 (approved):

Issue #9304, opened 2026-08-18
NIIT source list omits passive partnership/S-corp and estate income (section 1411(c)(1)(A)(ii))


MaxGhenis:
Found while preparing the business-income crosswalk memo for the TAXSIM collaboration (Microcosm #530 arc), verified against the formulas on current main.

`net_investment_income` adds the parameterized source list `gov.irs.investment.income.sources`: taxable_interest_income, dividend_income, rental_income, loss_limited_net_capital_gains, non_sch_d_capital_gains. Rental income is therefore the only business-income input in the NIIT base.

Per section 1411(c)(1)(A)(ii) and Form 8960 line 4a-4b, net investment income also includes income from passive trades or businesses — passive partnership/S-corp interests and estate/trust income retaining that character. The engine has no passive/active split on `partnership_income`/`s_corp_income`/`estate_income` today, so a full fix needs either a passivity input or a documented assumption; either way the current silent exclusion understates the NIIT base for pass-through-heavy filers.

Flagged in the crosswalk memo to Sabelhaus/Feenberg as an engine-scope note ('we flag rather than paper over it').

🤖 Generated with [Claude Code](https://claude.com/claude-code)

MaxGhenis:
Scope adjudication: #9306 implements the passive partnership/S-corp leg (new `passive_partnership_s_corp_income` input, §469 passive subset of income already in AGI, added to the NIIT source list from 2013; default zero until data supplies it).

The **estate leg stays open and blocked**: `estate_income` is not in `gov.irs.gross_income.sources` at all today (the file carries placeholder comments where it would go), so adding it to the NIIT base before it enters AGI would tax income the model never counts as income. Sequencing: (1) estate income into gross income (a revenue-affecting change deserving its own PR and validation), then (2) its NIIT inclusion — this issue stays open to track that second leg.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #530, opened 2022-01-30
Add GH Action to run Jupyterbook notebooks and commit back to the PR


MaxGhenis:

nikhilwoodruff:
This'd add a commit to every other commit, which I think would be over-complicated and error-prone.

PR #9306, opened 2026-08-18
Add passive partnership/S-corp income input to the NIIT base

MaxGhenis:
The NIIT source list (`gov.irs.investment.income.sources`) included rental income but omitted the section 1411(c)(1)(A)(ii) passive-business leg entirely, so passive pass-through income was never taxed under NIIT.

This adds a `passive_partnership_s_corp_income` input — the section 469 passive subset of `partnership_s_corp_income`, which is already in gross income, so including the subset in the NIIT base adds no income to AGI — and appends it to the source list from 2013 (NIIT has included passive business income since inception). It defaults to zero, so behavior is unchanged until a dataset supplies it; the Microcosm data side can populate it from the JCT/SOI passivity evidence gathered in Microcosm #530.

**Deliberately excluded**: the estate-income leg of #9304. `estate_income` is not in `gov.irs.gross_income.sources` today (placeholder comments mark the spot), and taxing income under NIIT before it enters AGI would be incoherent — that leg is blocked on gross-income inclusion first, re-scoped on the issue.

Tests: above-threshold NIIT computation with the new input (3.8% × min(NII, excess AGI)), below-threshold zero, and a default-zero regression case; also deduplicates two identically-named existing test cases.

Refs #9304

### DIFF
--- a/changelog.d/niit-passive-partnership-s-corp-income.added.md
+++ b/changelog.d/niit-passive-partnership-s-corp-income.added.md
@@ -1,0 +1,1 @@
+Passive partnership and S-corporation income input, included in the net investment income tax base.
--- a/policyengine_us/parameters/gov/irs/investment/income/sources.yaml
+++ b/policyengine_us/parameters/gov/irs/investment/income/sources.yaml
@@ -4,6 +4,7 @@ values:
     - taxable_interest_income
     - dividend_income
     - rental_income
+    - passive_partnership_s_corp_income
     - loss_limited_net_capital_gains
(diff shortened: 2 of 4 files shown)


MaxGhenis:
Data-side population tracked in PolicyEngine/microcosm#722 (assignment stage: SCF non-actively-managed-business shape, SOI/Form 8960 level) with the anchor ingestion in PolicyEngine/chronicle#179. Until that lands this input stays at its zero default, so merging this PR changes no results — it makes the NIIT base able to receive the passive leg.

PR #4130, opened 2024-03-11
Rename WIDOW to SURVIVING_SPOUSE filing status

JudyZ13:
Fixes #4094

PR #3807, opened 2024-02-07
Update Arkansas parameters based on 2023 tax form

Kiki-67:
Fixes #3804

PR #9321, opened 2026-08-20
Fix Arkansas inflation relief credit for combined filers

### LINKED ISSUE #9320: Arkansas inflation relief credit: combined filers use the joint table instead of the per-spouse single-filer table
## Summary

For a married couple filing a **combined** Arkansas return (filing status 4, `ar_files_separately = True`), `ar_inflation_relief_credit_person` looks up the credit in the **joint** table (max $300, phase-out from $174,000) and splits it 50/50. It should instead look up each spouse **separately in the single-filer table** (max $150, phase-out from $87,000, gone at $101,000), per the Arkansas Inflationary Relief Income-Tax Credit Worksheet, Filing Status 4.

## The worksheet

Arkansas Inflationary Relief Income-Tax Credit Worksheet, **Filing Status 4** (married filing combined/separately on the same return):

> Filing Status 4 — Primary / Spouse
> 1. Enter amount from line 28 of your AR1000F — 1A / 1B
> 2. Find income range in table located on next page. Enter corresponding credit for each spouse — 2A / 2B
> 3. Add primary and spouse columns from line 2 above. Enter amount here and on line 7 of form AR1000TC.

Each spouse is looked up in the single-filer table (the "$1–$87,000 → $150 … $101,001 and up → $0" schedule). The $300 / $174,000 schedule applies only to Filing Status 2 (a true joint return on one combined income).

## Failing case (taxsim [#1143](https://github.com/PolicyEngine/policyengine-taxsim/issues/1143), NBER out2psl/1142)

Joint couple, TY2022, $88,000 + $162,381 wages, $10,294 interest, 1 dependent — files combined (cheaper). Per-spouse net income (AR1000F line 28): primary **$90,877**, spouse **$165,258** (PolicyEngine computes the same split).

TaxAct worksheet (Filing Status 4, single-filer table per spouse):
- Primary $90,877 → **$110** (row $90,001–$91,000)
- Spouse $165,258 → **$0** (over $101,000)
- Total = **$110** (AR1000TC line 7; AR1000F line 36 = 110, net tax line 38 = 11,925)

PolicyEngine (joint table, split 50/50):
- Primary: max($300 − 0, 0) / 2 = **$150**
- Spouse: max($300 − 0, 0) / 2 = **$150**
- Total = **$300** → AR net tax **$11,739**, about $186 low (the $190 over-credit net of a ~$4 rounding difference in tax before credits).

## Root cause

`ar_inflation_relief_credit_person` keys the table parameters off `filing_status`, which is `JOINT` for any married couple regardless of the combined-vs-joint method:

```python
income = where(filing_separately, indiv_income, joint_income)
...
max_amount = p.max_amount[filing_status]          # JOINT -> $300
reduction_start = p.reduction.start[filing_status] # JOINT -> $174,000
...
divisor = where(joint, 2, 1)
credit = max_(max_amount - total_reduction_amount, 0) / divisor
```

When `ar_files_separately` is true, the per-spouse income is used (good) but the **joint** table and the /2 split are still applied (wrong). It should use the single-filer (SEPARATE) table per spouse and no /2.

## Suggested fix

Use the SEPARATE table when the couple files combined, and reserve the /2 split for a true joint return:

```python
credit_status = where(filing_separately, statuses.SEPARATE, filing_status)
max_amount = p.max_amount[credit_status]
reduction_start = p.reduction.start[credit_status]
increment = p.reduction.increment[credit_status]
reduction_per_increment = p.reduction.amount[credit_status]
...
true_joint = (filing_status == statuses.JOINT) & ~filing_separately
divisor = where(true_joint, 2, 1)
```

This changes only the combined-married case; true-joint, single, and head-of-household cases are unchanged.

## Related

The credit now differs by filing method ($110 combined vs $0 joint here), so ideally the combined-vs-joint selection in `ar_files_separately` should weigh it, as it already does for the additional tax credit for qualified individuals (#9299/#9300). Out of scope for the credit-value fix but noted for follow-up.

## Runnable TAXSIM input

```
taxsimid,year,state,mstat,page,sage,depx,pwages,psemp,swages,ssemp,dividends,intrec,stcg,ltcg,otherprop,nonprop,pensions,gssi,pui,sui,transfers,rentpaid,proptax,otheritem,childcare,mortgage,scorp,idtl,opt1,opt1v
2567204,2022,4,2,38,31,1,88000,0,162380.95,0,0,10294.458,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,5,30,1
```

DTrim99:
Fixes #9320.

## What

For a married couple filing a **combined** Arkansas return (filing status 4, `ar_files_separately = True`), the inflationary relief credit is looked up **per spouse in the single-filer table** (max $150, phased out from $87,000, gone at $101,000), per the Arkansas Inflationary Relief Income-Tax Credit Worksheet, Filing Status 4. The joint table (max $300, phased out from $174,000) applies only to a true joint return (status 2) on one combined income.

`ar_inflation_relief_credit_person` keyed the table off `filing_status`, which is `JOINT` for any married couple, so combined filers wrongly used the joint table split 50/50. This PR uses the single-filer (SEPARATE) table per spouse for combined filers, reserving the joint table and the halving split for a true joint return.

## Failing case (taxsim [#1143](https://github.com/PolicyEngine/policyengine-taxsim/issues/1143))

Joint couple, TY2022, $88,000 + $162,381 wages, $10,294 interest, 1 dependent — files combined. Per-spouse net income (AR1000F line 28): primary $90,877, spouse $165,258 (PolicyEngine computes the same split).

- TaxAct worksheet (single-filer table per spouse): $110 + $0 = **$110** (AR1000TC line 7; net tax 11,925).
- PolicyEngine before (joint table, 50/50): $150 + $150 = **$300** (net tax 11,739, ~$186 low).
- PolicyEngine after: **$110**. ✓

## Impact

Only the combined-married case changes. True-joint, single, head-of-household, surviving-spouse, and separate cases are unchanged (verified against the existing tests).

## Tests

- Added the #1143 combined-filing case ([110, 0]) and a true-joint contrast on the same incomes ([0, 0], combined income $256,135 > $174,000).
- Full Arkansas suite passes (469 tests).

## Related

The credit now differs by filing method ($110 combined vs $0 joint here), so the combined-vs-joint selection in `ar_files_separately` should ideally weigh it, as it already does for the additional tax credit for qualified individuals (#9300). Left as a follow-up to keep this change focused and avoid a circular dependency.

### DIFF
--- a/changelog.d/fix-ar-inflation-relief-combined-table.fixed.md
+++ b/changelog.d/fix-ar-inflation-relief-combined-table.fixed.md
@@ -1,0 +1,1 @@
+Fixed the Arkansas inflation relief income-tax credit for married couples filing combined (status 4), which now uses the per-spouse single-filer table (maximum $150, phased out from $87,000) instead of the joint table (maximum $300, phased out from $174,000) split between spouses.
--- a/policyengine_us/tests/policy/baseline/gov/states/ar/tax/income/credits/ar_inflation_relief_credit_person.yaml
+++ b/policyengine_us/tests/policy/baseline/gov/states/ar/tax/income/credits/ar_inflation_relief_credit_person.yaml
@@ -151,3 +151,51 @@
         state_code: AR
   output:
     ar_inflation_relief_credit_person: [150, 150, 0]
+
+# Married couple filing combined (status 4): each spouse uses the single-filer
+# table on their own net income, not the joint table split in half (taxsim #1143).
+- name: Married couple filing combined uses the per-spouse single-filer table
+  period: 2022
+  input:
+    people:
+      person1:
+        ar_taxable_income_indiv: 90_877
+      person2:
+        ar_taxable_income_indiv: 165_258
+    tax_units:
+      tax_unit:
+        members: [person1, person2]
+        ar_files_separately: true
+        filing_status: JOINT
+    households:
+      household:
+        members: [person1, person2]
+        state_code: AR
+  output:
+    # person1: $150 - ceil(3,877 / 1,000) * $10 = $150 - $40 = $110
+    # person2: $165,258 > $101,000 -> $0
+    ar_inflation_relief_credit_person: [110, 0]
+
+# The same couple filing a true joint return uses the joint table on combined
+# income ($256,135 > $174,000 -> $0), split in half.
+- name: Married couple filing a true joint return uses the joint table on combined income
+  period: 2022
+  input:
+    people:
+      person1:
+        ar_taxable_income_indiv: 0
+        ar_net_taxable_income_joint: 90_877
+      person2:
+        ar_taxable_income_indiv: 0
+        ar_net_taxable_income_joint: 165_258
+    tax_units:
+      tax_unit:
+        members: [person1, person2]
+        ar_files_separately: false
+        filing_status: JOINT
+    households:
+      household:
+        members: [person1, person2]
+        state_code: AR
+  output:
+    ar_inflation_relief_credit_person: [0, 0]
--- a/policyengine_us/variables/gov/states/ar/tax/income/credits/ar_inflation_relief_credit_person.py
+++ b/policyengine_us/variables/gov/states/ar/tax/income/credits/ar_inflation_relief_credit_person.py
@@ -22,16 +22,27 @@ class ar_inflation_relief_credit_person(Variable):
         income = where(filing_separately, indiv_income, joint_income)
         p = parameters(period).gov.states.ar.tax.income.credits.inflationary_relief
         filing_status = person.tax_unit("filing_status", period)
-        max_amount = p.max_amount[filing_status]
-        reduction_start = p.reduction.start[filing_status]
-        increment = p.reduction.increment[filing_status]
-        reduction_per_increment = p.reduction.amount[filing_status]
+        statuses = filing_status.possible_values
+        # A married couple filing combined (status 4) computes the credit for
+        # each spouse separately in the single-filer table (max $150, phased
+        # out from $87,000), per the Inflationary Relief Credit Worksheet.
+        # Only a true joint return (status 2) uses the joint table (max $300,
+        # phased out from $174,000) on combined income and splits it in half.
+        credit_status = statuses.encode(
+            where(filing_separately, statuses.SEPARATE, filing_status.decode())
+        )
+        max_amount = p.max_amount[credit_status]
+        reduction_start = p.reduction.start[credit_status]
+        increment = p.reduction.increment[credit_status]
+        reduction_per_increment = p.reduction.amount[credit_status]
         excess = max_(income - reduction_start, 0)
         increments = np.ceil(excess / increment)
         total_reduction_amount = increments * reduction_per_increment
-        # Attribute the maximum amount to each spouse equally when married filing jointly
-        joint = filing_status == filing_status.possible_values.JOINT
-        divisor = where(joint, 2, 1)
+        # Attribute the maximum amount to each spouse equally only when the
+        # couple files a true joint return (combined filers already use their
+        # own individual income above).
+        true_joint = (filing_status == statuses.JOINT) & ~filing_separately
+        divisor = where(true_joint, 2, 1)
         credit = max_(max_amount - total_reduction_amount, 0) / divisor
         # The credit is only allocated to the head and spouse, not dependents.
         head_or_spouse = person("is_tax_unit_head_or_spouse", period)


hua7450 (approved):
Reviewed with a source-verification pass against the official documents (2022 AR1000F/AR1000NR booklet, the standalone 2022 and 2023 Inflationary Relief worksheets, and the DFA HB1002 impact statement). **The fix is correct — approving.** Also surfacing some pre-existing issues found along the way; @DTrim99 could you take a look and see which are worth fixing (follow-up or here)?

## Verified

- **The status-4 remap matches the law, triple-corroborated**: [Act 2 of 2022, 3rd Ex. Sess., § 6(b)(1)(B)](https://www.arkleg.state.ar.us/Home/FTPDocument?path=/ACTS/2022S3/Public/ACT2.pdf#page=18) ("Spouses filing separately on the same income-tax return may each claim one (1) credit"), the 2022 worksheet (the $150 table is captioned "Filing Status 1,3,**4**,5, and 6", and the Status 4 block looks up each spouse's own line-28 income), and the [DFA impact statement](https://www.arkleg.state.ar.us/Home/FTPDocument?path=%2FAssembly%2F2021%2F2022S3%2FFiscal+Impacts%2FHB1002-DFA1.pdf#page=2). The true-joint path correctly keeps the $300 table on combined income.
- All 17 audited 2022 parameter values match the worksheet exactly (joint phase-out ends at $202,000).
- Both new tests' expectations hand-verified against the PDF brackets: $90,877 → $110 ($90,001–$91,000 row); $165,258 > $101,000 → $0; joint $256,135 > $202,000 → $0.
- Code patterns clean: fully parameterized, vectorized encode/decode round-trip, correct changelog fragment. CI green.

## Non-blocking suggestions for this PR

1. `ar_files_separately`'s cheaper-path comparison (and its "other credits are the same for both paths" comment) now ignores a credit that differs by path — a couple with one spouse above $101,000 gets $150×1 combined vs $300 joint, so the optimizer can pick a path that loses up to $300. You already flagged this in the PR description as a follow-up; the comment at `ar_files_separately.py:28-31` should at least be corrected now so it doesn't state something false.
2. The changed variable has no `reference` — and it now implements exactly the rule under review. The act PDFs above are the durable cite (there's no permanent statute: both enactments are "DO NOT CODIFY" temporary language; Ark. Code Ann. § 26-51-516 is actually the stillborn-child credit).
3. Test gaps in the changed direction: no case where the fix *reduces* the credit (e.g., combined [$30,000, $100,000] → new $150 + $20 = $170 vs old $300), and no exact boundary cases ($87,000 → $150, $101,000 → $10, joint $174,000 / $202,000). Also nothing asserts the credit flowing into `ar_income_tax` (the taxsim case's net-tax figure is only a YAML comment).
4. Minor: the new code comment and changelog state the 2022 amounts ($150/$87,000/$300/$174,000) as if constant — the 2023 tables differ; and consider `where(filing_separately & (filing_status == statuses.JOINT), statuses.SEPARATE, filing_status.decode())` so a directly-input flag on a non-married status can't silently reroute if the single-type columns ever diverge.

## Pre-existing issues (not this PR — worth fixing?)

1. **The 2023 parameter values contradict the official 2023 worksheet by up to 3×.** The [2023 worksheet](https://www.dfa.arkansas.gov/wp-content/uploads/2023_Final_InflationaryReliefIncomeTaxCredit_FI.pdf) keeps $150/$300 maximums (phase-outs $89,600→$103,600 and $179,200→$207,200), but `inflationary_relief/max_amount.yaml` stores 2023 maximums of **$50/$100**, and `reduction/start.yaml`'s JOINT block has **no 2023 row** (inherits $174,000). Origins traced: the $50/$100 matches the *State Political Contributions Credit* caps on Line 1 of the 2023 AR1000TC instructions (wrong credit), and the JOINT `2023-01-01: 179_200` row **was** added by #3807 but accidentally deleted by the WIDOW→SURVIVING_SPOUSE rename in #4130. Fixing just those two files reproduces the official 2023 tables exactly (`increment`/`amount` are already right). Verified by code-path trace and 600 DPI page reading.
2. **The credit's parameters outlive its sunset.** The credit doesn't exist for TY2024 (absent from the [2024 booklet's AR1000TC credit list](https://www.dfa.arkansas.gov/wp-content/uploads/2024_AR1000F_and_AR1000NR_Instructions.pdf#page=15)). Behavior is already right — `non_refundable.yaml` drops the credit at 2024-01-01 — but `max_amount.yaml` carries $50/$100 forward indefinitely, so the person-level variable still pays if queried directly in 2024+.
3. **3 of 4 reference URLs on all four `inflationary_relief` parameter files are dead** — DFA moved from `/images/uploads/incomeTaxOffice/` to `/wp-content/uploads/` (same filenames). While refreshing: the booklet worksheet is at `#page=25` in the current copy (cited as 26), and two entries share the identical title "Inflationary relief income-tax credit worksheet" (2022 vs 2023).
4. **Scalar-masked doubling in the existing HOH/surviving-spouse tests**: they put two adults in the unit and assert a scalar, so the broadcast gives the unit $60/$300 — double those statuses' $150 per-return max (the divisor only halves true JOINT). Artificial situations, but worth rewriting as single-adult cases with explicit arrays. Two tests are also both named "Single household with $100,000 taxable income" though the second is a SEPARATE couple.

Items 1–3 all touch the same four parameter files, so they'd bundle naturally into one follow-up PR.

PR #3903, opened 2024-02-16
Decompose snap_utility_allowance into three variables for SUA, LUA, and IUA

yy3187:
Fixes #3869

PR #9344, opened 2026-08-25
Filter SNAP individual utility allowance to incurred utility expenses

### LINKED ISSUE #9343: SNAP individual utility allowance sums all utility-type standards regardless of incurred expenses (regression of #3853)
## Description

`snap_individual_utility_allowance` sums the individual standards for **all** utility types in `utility.single.utility_types`, without checking whether the household actually incurs each expense.

Example (MO, FY2026): a household whose only utility bill is a telephone receives $891/month (5 × $162 for electricity, gas/fuel, water, sewage, trash + $81 phone) instead of the $81 telephone standard.

## Federal rule

7 CFR 273.9(d)(6)(iii)(A) authorizes "an individual standard for each type of utility expense" — the standard substitutes for the specific expense the household incurs.

## History

- #3853 reported exactly this defect; fixed by #3867 (merged 2024-02-13) with per-expense filtering.
- #3903 (merged 2024-02-28) decomposed `snap_utility_allowance` into SUA/LUA/IUA variables and reverted to the unconditional sum, with no review discussion of dropping the filter.
- The current test file enshrines the regressed behavior (e.g. the AK_C trash-only test asserts the sum of all six standards).

## Impact

- Affects every state with individual utility standards, for households on the IUA path (one non-heating utility bill in LUA states; any non-heating bills in states without a LUA).
- Produces an ordering inversion: a 1-bill household can receive more ($891 in MO) than a 2-bill LUA household ($373) or a heating household's full SUA ($508).

hua7450:
## Summary

Restores the individual utility allowance (IUA) expense filtering that #3867 added and the #3903 SUA/LUA/IUA decomposition silently reverted two weeks later. Households on the IUA path now receive an individual standard only for each utility expense they actually incur, per 7 CFR 273.9(d)(6)(iii)(A) ("an individual standard for each type of utility expense").

Fixes #9343

## The defect

`snap_individual_utility_allowance` summed the standards for **all** types in `utility.single.utility_types` regardless of the household's actual bills. Example (MO, FY2026): a phone-only household received $891/month (5 × $162 electricity/gas/water/sewage/trash + $81 phone) instead of the $81 telephone standard — more than MO's 2-bill LUA ($373) or full heating SUA ($508), an ordering inversion where fewer utilities produced a larger deduction.

## Changes

- `snap_individual_utility_allowance.py`: each utility-type standard now counts only when the household's corresponding expense is positive. Electricity checks `pre_subsidy_electricity_expense` to avoid the SNAP-enrollment circularity, matching `count_distinct_utility_expenses`. The HI/GU household-size handling is unchanged, just gated by expense.
- Tests: the AK_C trash-only cases now assert the single trash standard ($30 FY22 / $36 FY24) instead of the all-six sum; HI/GU household-size cases supply all six expenses (same totals, now legitimately); new regression cases for MO phone-only ($81 FY26), MO electricity-only ($162), AK water+sewage stacking ($111 = 56 + 55), and no-expense ($0).

Multi-utility stacking for incurred expenses is preserved — that part is correct under the federal rule. Partner contract tests are unaffected: they exercise only the SUA (heating/cooling) and LUA (two-bill) paths.

## Impact

- **Population microsimulation: no change.** The utility expense input variables are unpopulated in the certified datasets (empty groups in `enhanced_cps_2024.h5`; absent from `populace_us_2024.h5`), so `count_distinct_utility_expenses` is 0 everywhere and no unit routes to the IUA branch — measured SNAP aggregate delta is exactly $0, with 0 weighted units changed. The aggregate is nil *because* the data is unpopulated, not because the effect is small; if utility expenses are ever imputed into the datasets, this fix lowers SNAP for IUA households (the allowance is strictly non-increasing under the filter).
- **Household calculator / API: real, and can be large.** Any API consumer supplying utility expense inputs for an IUA-path household will see a smaller allowance. Example (MO, single person, $1,000/month rent, phone bill only, 2026): `snap_individual_utility_allowance` drops 891 → 81 and `snap_excess_shelter_expense_deduction` drops 744 (pinned at the shelter cap) → 81, up to ≈$199/month less SNAP at the 30% benefit-reduction rate, and possible loss of eligibility via the net income test. Mandatory-SUA states (e.g. CA, MA, WA, CO) and any household on the SUA/LUA paths are unaffected.

## Test plan

- [x] 14 IUA unit tests pass
- [x] Full SNAP baseline suite passes (605 tests)
- [x] CI passes (33/33 checks)

### DIFF
--- a/changelog.d/fix-snap-iua-filter-incurred.fixed.md
+++ b/changelog.d/fix-snap-iua-filter-incurred.fixed.md
@@ -1,0 +1,1 @@
+Filter the SNAP individual utility allowance to the utility expenses the household actually incurs, restoring the fix from #3867 that was reverted during the #3903 decomposition.
--- a/policyengine_us/tests/policy/baseline/gov/usda/snap/income/deductions/snap_individual_utility_allowance.yaml
+++ b/policyengine_us/tests/policy/baseline/gov/usda/snap/income/deductions/snap_individual_utility_allowance.yaml
@@ -67,8 +67,24 @@
     snap_utility_allowance_type: IUA
     snap_utility_region_str: MO
     pre_subsidy_electricity_expense: 1_200
+    # Overriding electricity_expense to 0 pins that the formula reads the
+    # pre-subsidy variable. No naturally occurring household can distinguish
+    # the two by value: the only subsidies (ca_care, ca_fera, ca_la_ez_save)
+    # are California-only and CA's electricity standard is 0.
+    electricity_expense: 0
   output:
     # MO FY26 electricity standard
+    snap_individual_utility_allowance: 162
+
+- name: MO gas-only household receives the gas standard in 2026
+  period: 2026-01
+  input:
+    snap_utility_allowance_type: IUA
+    snap_utility_region_str: MO
+    gas_expense: 2_400
+  output:
+    # MO FY26 gas standard, read via the gas_and_fuel_expense -> gas_expense
+    # variable mapping
     snap_individual_utility_allowance: 162
 
 # HI individual utility allowances vary by household size
@@ -169,5 +185,37 @@
   input:
     snap_utility_allowance_type: IUA
     snap_utility_region_str: MO
+  output:
+    snap_individual_utility_allowance: 0
+
+# The individual allowance is 0 on the non-IUA paths even when the household
+# incurs utility expenses; snap_utility_allowance sums the SUA, LUA, and IUA
+# components, so a nonzero value here would double count.
+- name: SUA household receives no individual allowance despite utility bills
+  period: 2026-01
+  input:
+    snap_utility_allowance_type: SUA
+    snap_utility_region_str: MO
+    heating_cooling_expense: 2_400
+    phone_expense: 600
+  output:
+    snap_individual_utility_allowance: 0
+
+- name: LUA household receives no individual allowance despite utility bills
+  period: 2026-01
+  input:
+    snap_utility_allowance_type: LUA
+    snap_utility_region_str: MO
+    water_expense: 480
+    trash_expense: 360
+  output:
+    snap_individual_utility_allowance: 0
+
+- name: NONE-type household receives no individual allowance despite a utility bill
+  period: 2026-01
+  input:
+    snap_utility_allowance_type: NONE
+    snap_utility_region_str: MO
+    phone_expense: 600
   output:
     snap_individual_utility_allowance: 0
--- a/policyengine_us/tests/policy/baseline/gov/usda/snap/income/deductions/snap_individual_utility_allowance.yaml
+++ b/policyengine_us/tests/policy/baseline/gov/usda/snap/income/deductions/snap_individual_utility_allowance.yaml
@@ -1,3 +1,5 @@
+# Households receive an individual standard only for each utility expense
+# they actually incur (7 CFR 273.9(d)(6)(iii)(A)).
 - name: Test in 2022 with phone expenses
   period: 2022-01
   absolute_error_margin: 1
@@ -18,68 +20,136 @@
   output:
     snap_individual_utility_allowance: 19
 
-- name: Test in 2022 with trash expenses
+- name: Trash-only household receives only the trash standard in 2022
   period: 2022-01
   input:
     snap_utility_allowance_type: IUA
     snap_utility_region_str: AK_C
     trash_expense: 20
   output:
-    snap_individual_utility_allowance: 380
-    # AK_C FY22: elec=110 + gas=114 + water=56 + sewer=55 + trash=30 + phone=15
+    # AK_C FY22 trash standard only, not the sum of all six standards
+    snap_individual_utility_allowance: 30
 
-- name: Test in 2024 with trash expenses
+- name: Trash-only household receives only the trash standard in 2024
   period: 2024-01
   input:
     snap_utility_allowance_type: IUA
     snap_utility_region_str: AK_C
     trash_expense: 20
   output:
-    snap_individual_utility_allowance: 638
+    # AK_C FY24 trash standard only
+    snap_individual_utility_allowance: 36
+
+- name: Water and sewage household receives both incurred standards in 2022
+  period: 2022-01
+  input:
+    snap_utility_allowance_type: IUA
+    snap_utility_region_str: AK_C
+    water_expense: 240
+    sewage_expense: 240
+  output:
+    # AK_C FY22: water=56 + sewage=55
+    snap_individual_utility_allowance: 111
+
+- name: MO phone-only household receives the telephone standard in 2026
+  period: 2026-01
+  input:
+    snap_utility_allowance_type: IUA
+    snap_utility_region_str: MO
+    phone_expense: 972
+  output:
+    # MO FY26 phone standard, not the $891 sum of all six standards
+    snap_individual_utility_allowance: 81
+
+- name: MO electricity-only household receives the electricity standard in 2026
+  period: 2026-01
+  input:
+    snap_utility_allowance_type: IUA
+    snap_utility_region_str: MO
+    pre_subsidy_electricity_expense: 1_200
+  output:
+    # MO FY26 electricity standard
+    snap_individual_utility_allowance: 162
 
 # HI individual utility allowances vary by household size
 # HH1 FY2025: elec=291 + gas=291 + water=65 + sewage=97 + trash=97 + phone=42 = 883
-- name: HI IUA HH size 1 in FY2025
+- name: HI IUA HH size 1 with all utility expenses in FY2025
   period: 2025-01
   absolute_error_margin: 1
   input:
     snap_utility_allowance_type: IUA
     snap_utility_region_str: HI
     spm_unit_size: 1
+    pre_subsidy_electricity_expense: 1_200
+    gas_expense: 240
+    water_expense: 240
+    sewage_expense: 240
+    phone_expense: 240
+    trash_expense: 240
   output:
     snap_individual_utility_allowance: 883
 
 # HH3 FY2025: elec=366 + gas=366 + water=81 + sewage=97 + trash=97 + phone=42 = 1049
-- name: HI IUA HH size 3 in FY2025
+- name: HI IUA HH size 3 with all utility expenses in FY2025
   period: 2025-01
   absolute_error_margin: 1
   input:
     snap_utility_allowance_type: IUA
     snap_utility_region_str: HI
     spm_unit_size: 3
+    pre_subsidy_electricity_expense: 1_200
+    gas_expense: 240
+    water_expense: 240
+    sewage_expense: 240
+    phone_expense: 240
+    trash_expense: 240
   output:
     snap_individual_utility_allowance: 1_049
 
+- name: HI IUA HH size 3 with only electricity expenses in FY2025
+  period: 2025-01
+  absolute_error_margin: 1
+  input:
+    snap_utility_allowance_type: IUA
+    snap_utility_region_str: HI
+    spm_unit_size: 3
+    pre_subsidy_electricity_expense: 1_200
+  output:
+    # HI HH3 FY2025 electricity standard only
+    snap_individual_utility_allowance: 366
+
 # GU individual utility allowances vary by household size
 # HH1 FY2025: elec=188 + gas=36 + water=38 + sewage/trash/phone uprated from FY2024
-- name: GU IUA HH size 1 in FY2025
+- name: GU IUA HH size 1 with all utility expenses in FY2025
   period: 2025-01
   absolute_error_margin: 3
   input:
     snap_utility_allowance_type: IUA
     snap_utility_region_str: GU
     spm_unit_size: 1
+    pre_subsidy_electricity_expense: 1_200
+    gas_expense: 240
+    water_expense: 240
+    sewage_expense: 240
+    phone_expense: 240
+    trash_expense: 240
   output:
     snap_individual_utility_allowance: 350
 
 # HH2 FY2025: elec=218 + gas=36 + water=50 + sewage/trash/phone uprated from FY2024
-- name: GU IUA HH size 2 in FY2025
+- name: GU IUA HH size 2 with all utility expenses in FY2025
   period: 2025-01
   absolute_error_margin: 3
   input:
     snap_utility_allowance_type: IUA
     snap_utility_region_str: GU
     spm_unit_size: 2
+    pre_subsidy_electricity_expense: 1_200
+    gas_expense: 240
+    water_expense: 240
+    sewage_expense: 240
+    phone_expense: 240
+    trash_expense: 240
   output:
     snap_individual_utility_allowance: 392
 
@@ -90,5 +160,14 @@
     snap_utility_allowance_type: IUA
     snap_utility_region_str: NC
     spm_unit_size: 1
+    phone_expense: 240
   output:
     snap_individual_utility_allowance: 41
+
+- name: Household with no utility expenses receives no individual allowance
+  period: 2025-01
+  input:
+    snap_utility_allowance_type: IUA
+    snap_utility_region_str: MO
+  output:
+    snap_individual_utility_allowance: 0
--- a/policyengine_us/variables/gov/usda/snap/income/deductions/shelter/snap_individual_utility_allowance.py
+++ b/policyengine_us/variables/gov/usda/snap/income/deductions/shelter/snap_individual_utility_allowance.py
@@ -17,6 +17,9 @@ class snap_individual_utility_allowance(Variable):
     unit = USD
     documentation = "The individual utility allowance deduction for SNAP"
     definition_period = MONTH
+    reference = (
+        "https://www.ecfr.gov/current/title-7/section-273.9#p-273.9(d)(6)(iii)(A)"
+    )
 
     def formula(spm_unit, period, parameters):
         utility = parameters(period).gov.usda.snap.income.deductions.utility
@@ -36,8 +39,6 @@ class snap_individual_utility_allowance(Variable):
 
         hh_size_utilities = {"electricity", "gas_and_fuel", "water"}
 
-        # Households receive an individual standard only for each utility
-        # expense they actually incur (7 CFR 273.9(d)(6)(iii)(A)).
         sum_of_individual_allowances = 0
         for expense in expense_types:
             util_name = expense.replace("_expense", "")
--- a/policyengine_us/variables/gov/usda/snap/income/deductions/shelter/snap_individual_utility_allowance.py
+++ b/policyengine_us/variables/gov/usda/snap/income/deductions/shelter/snap_individual_utility_allowance.py
@@ -1,5 +1,14 @@
 from policyengine_us.model_api import *
 
+# Expense input variables that differ from the <type>_expense naming of the
+# utility_types parameter entries. Electricity uses pre-subsidy expenses to
+# avoid circular references since electricity subsidies depend on SNAP
+# enrollment.
+EXPENSE_VARIABLE_OVERRIDES = {
+    "electricity_expense": "pre_subsidy_electricity_expense",
+    "gas_and_fuel_expense": "gas_expense",
+}
+
 
 class snap_individual_utility_allowance(Variable):
     value_type = float
@@ -26,23 +35,14 @@ class snap_individual_utility_allowance(Variable):
         safe_region = where(is_hh_size_state, region, "HI")
 
         hh_size_utilities = {"electricity", "gas_and_fuel", "water"}
+
         # Households receive an individual standard only for each utility
         # expense they actually incur (7 CFR 273.9(d)(6)(iii)(A)).
-        # Use pre-subsidy electricity expenses to avoid circular references
-        # since electricity subsidies depend on SNAP enrollment.
-        expense_variables = {
-            "electricity": "pre_subsidy_electricity_expense",
-            "gas_and_fuel": "gas_expense",
-            "water": "water_expense",
-            "sewage": "sewage_expense",
-            "phone": "phone_expense",
-            "trash": "trash_expense",
-        }
-
         sum_of_individual_allowances = 0
         for expense in expense_types:
             util_name = expense.replace("_expense", "")
-            incurs_expense = spm_unit(expense_variables[util_name], period) > 0
+            expense_variable = EXPENSE_VARIABLE_OVERRIDES.get(expense, expense)
+            incurs_expense = spm_unit(expense_variable, period) > 0
             flat_val = utility.single[util_name][region]
             if util_name in hh_size_utilities:
                 hh_val = utility.single.by_household_size[util_name][safe_region][
--- a/policyengine_us/variables/gov/usda/snap/income/deductions/shelter/snap_individual_utility_allowance.py
+++ b/policyengine_us/variables/gov/usda/snap/income/deductions/shelter/snap_individual_utility_allowance.py
@@ -26,20 +26,32 @@ class snap_individual_utility_allowance(Variable):
         safe_region = where(is_hh_size_state, region, "HI")
 
         hh_size_utilities = {"electricity", "gas_and_fuel", "water"}
+        # Households receive an individual standard only for each utility
+        # expense they actually incur (7 CFR 273.9(d)(6)(iii)(A)).
+        # Use pre-subsidy electricity expenses to avoid circular references
+        # since electricity subsidies depend on SNAP enrollment.
+        expense_variables = {
+            "electricity": "pre_subsidy_electricity_expense",
+            "gas_and_fuel": "gas_expense",
+            "water": "water_expense",
+            "sewage": "sewage_expense",
+            "phone": "phone_expense",
+            "trash": "trash_expense",
+        }
 
         sum_of_individual_allowances = 0
         for expense in expense_types:
             util_name = expense.replace("_expense", "")
+            incurs_expense = spm_unit(expense_variables[util_name], period) > 0
             flat_val = utility.single[util_name][region]
             if util_name in hh_size_utilities:
                 hh_val = utility.single.by_household_size[util_name][safe_region][
                     capped_size
                 ]
-                sum_of_individual_allowances += where(
-                    is_hh_size_state, hh_val, flat_val
-                )
+                allowance = where(is_hh_size_state, hh_val, flat_val)
             else:
-                sum_of_individual_allowances += flat_val
+                allowance = flat_val
+            sum_of_individual_allowances += incurs_expense * allowance
 
         return where(
             allowance_type == allowance_types.IUA,


PavelMakarchuk (approved):
# PR 9344 — Filter SNAP individual utility allowance to incurred utility expenses

## Program Review

### Source Documents
- **PDF**: None — no source documents cited; regulation verified from govinfo 7 CFR 273.9/273.10 XML
- **Year**: FY22 / FY24 / FY25 / FY26 test periods; formula change applies to all years
- **Scope**: PR changes only
- **Reviewed head SHA**: dca7d83a97b1d648012e1ece602e31cc28894c67
- **Mode**: full

---

### Verdict in one paragraph

The formula is a faithful encoding of 7 CFR 273.9(d)(6)(iii)(A) — "an individual standard for
each type of utility expense", to be used "in place of actual costs". Before this PR,
`snap_individual_utility_allowance` paid a household the sum of **all six** individual standards
regardless of which bills it actually had; the PR gates each standard on `expense > 0`. All three
reviewer roles (regulatory, code-pattern, test-coverage) independently recomputed every one of the
14 test expectations from `parameters/gov/usda/snap/income/deductions/utility/single/**` and agree
exactly — three-way agreement, **zero CRITICAL findings**. The change is also an *improvement* on
the version it restores: #3867 used `select([...six conditions...], ...)`, which returns only the
**first** match, so an LUA-inactive region household with both water and sewage bills got one
standard instead of two; this PR's additive form is correct, and the new AK_C 2022 water+sewage
case (56 + 55 = 111) is precisely the case #3867 got wrong.

The most important thing for a maintainer to know is the impact story, which **changed during the
review** — see A1.

---

### Critical (Must Fix)

**None.** All three roles reported 0 CRITICAL.

Specifically, the two CRITICAL triggers in scope both clear:
- **The regression is pinned.** Six of the new/changed cases fail under the old sum-all-six
  formula (trash-only AK_C 30 vs 380 and 36 vs 638; water+sewage 111 vs 380; MO phone-only 81 vs
  891; MO electricity-only 162 vs 891; HI HH3 electricity-only 366 vs 1,049).
- **No non-functional tests.** No boolean outputs, so no `absolute_error_margin >= 1` on a
  boolean; all `period:` values are legal `YYYY-01`; all nine input variables exist and are
  SPMUnit/YEAR, matching what the formula reads (notably `pre_subsidy_electricity_expense`, not
  `electricity_expense`, and `gas_expense`, not the parameter key `gas_and_fuel_expense`).

---

### Should Address

#### A1. Impact disclosure: the change is **inert on the certified datasets**, but materially changes household-calculator results — the PR body says neither
`/Users/pavelmakarchuk/policyengine-us/policyengine_us/variables/gov/usda/snap/income/deductions/shelter/snap_individual_utility_allowance.py:45,54`
— 7 CFR 273.9(d)(6)(iii)(A)

*This supersedes an earlier prediction made during this review.* The context-analysis phase
reasoned a priori that the PR would cut SNAP at the population level, since the allowance is
**strictly non-increasing** (every term goes from `allowance` to `incurs × allowance`, all
standards ≥ 0) and lower IUA → lower `snap_excess_shelter_expense_deduction` → higher
`snap_net_income` → higher `snap_expected_contribution` → lower `snap_normal_allotment`, and it
can flip `meets_snap_net_income_test` for units near 100% FPG. The mechanical direction is indeed
unambiguously downward. But the regulatory role then **measured** the aggregate and found it is
exactly zero:

- `Microsimulation` over `enhanced_cps_2024` at 2026: `snap_utility_allowance_type` is
  SUA 37.1% / NONE 62.9% / **LUA 0% / IUA 0%**. Baseline vs. a reform applying this PR's formula:
  SNAP delta **$0.0000B (0.000%)**, 0 weighted units changed, max allowance increase 0.
- Root cause verified directly in the HDF5, not inferred: in `enhanced_cps_2024.h5`, every one of
  `heating_cooling_expense`, `pre_subsidy_electricity_expense`, `electricity_expense`,
  `gas_expense`, `water_expense`, `sewage_expense`, `phone_expense`, `trash_expense`, `phone_cost`
  is an **empty group** (no stored values ⇒ default 0). The certified `populace_us_2024.h5`
  carries no utility columns at all. So `count_distinct_utility_expenses` = 0 everywhere and
  **the IUA branch is unreachable in population runs.**

Direction and magnitude, reported without prescribing a remedy:
- **Aggregate microsim impact: nil**, and nil *because* utility expenses are unpopulated — not
  because the effect is small. The fix will only bite once utility expenses are imputed into the
  datasets, at which point the downward direction above becomes live.
- **Household-calculator / API impact: real and large.** Measured (MO, single person, $12k rent,
  phone-only, `Simulation` at `2026-01`, baseline working tree vs. this PR):
  `snap_individual_utility_allowance` **891 → 81** and `snap_excess_shelter_expense_deduction`
  **744 → 81**. The old value was pinned at the FY2026 non-elderly shelter cap
  (`excess_shelter_expense/cap.yaml`, CONTIGUOUS_US = 744), so the deduction falls $663/month; at
  the 30% expected-contribution rate that is up to **≈$199/month (≈$2.4k/year) less SNAP** for
  such a household, plus possible loss of eligibility via the net-income test.

The PR body currently frames this as a bug fix with no impact statement. Point (a) — nil aggregate
because of unpopulated data — is the load-bearing one for anyone triaging this against partner/API
expectations, since a partner calling the household API *will* see the change while a microsim
consumer will not.

A corroborating sanity check on the direction of the fix: the old MO FY2026 IUA of $891 exceeded
both the LUA ($373, `limited/main.yaml` MO 2025-10-01) and the SUA ($508, `standard/main.yaml`) —
a one-bill household was receiving more than a heating household, which is impossible under
(d)(6)(iii)(A) read with (E).

#### A2. Nothing pins the non-IUA branch — delete the outer `where` and the entire suite still passes
`/Users/pavelmakarchuk/policyengine-us/policyengine_us/tests/policy/baseline/gov/usda/snap/income/deductions/snap_individual_utility_allowance.yaml`
(all 14 cases) vs `snap_individual_utility_allowance.py:56-60`

Every case sets `snap_utility_allowance_type: IUA`, so
`where(allowance_type == allowance_types.IUA, sum_of_individual_allowances, 0)` is completely
unverified. If that line were removed, `snap_utility_allowance` (`adds = [standard, limited,
individual]`) would start **double-counting** for every SUA and LUA household, with no test
failure. This is a pre-existing gap, but it is exactly the line that carries the PR's own claim
that SUA/LUA households are unaffected, and it is the same class of "the tests do not pin the
invariant" weakness that permitted the #3903 regression. Three cheap cases close it (SUA with a
bill → 0, LUA with bills → 0, NONE with a bill → 0).

#### A3. The utility-type → expense-variable mapping is now duplicated in a third place
`snap_individual_utility_allowance.py:3-10,26` vs
`/Users/pavelmakarchuk/policyengine-us/policyengine_us/variables/household/expense/utilities/count_distinct_utility_expenses.py:12-22`
and `/Users/pavelmakarchuk/policyengine-us/policyengine_us/variables/household/expense/utilities/utility_expense.py:12-20`

Three lists now encode "which variables are the utility bills"; two of them independently encode
the same `electricity → pre_subsidy_electricity_expense` / `gas_and_fuel → gas_expense` mapping
(`utility_expense.py` deliberately differs, using post-subsidy `electricity_expense`). The
coupling is load-bearing: `count_distinct_utility_expenses` routes a household **into** the IUA
branch (`snap_utility_allowance_type.py:21,31`) while this new code decides **which standards** it
then receives. If the lists desync, a household is routed to `IUA` on the strength of a bill this
formula does not count and lands on a **$0 allowance with a non-`NONE` allowance type** — a silent
benefit loss with no test and no type error to catch it.

The two lists agree today; both were checked element by element. Suggested direction: hold the
mapping once — either move the expense-variable names into the parameter tree next to
`utility/single/utility_types.yaml` (the repo already supports a parameter whose value is a list of
variable names, e.g. `adds = "gov.irs.credits.refundable"`), or have both variables import a single
shared constant. Note the module-level `UPPER_CASE` dict itself is **not** the finding —
`EXPENSE_VARIABLE_OVERRIDES` follows an established repo pattern
(`SPM_TENURE_TYPE_TO_GEOADJ_KEY`, `AK_COUNTY_TO_REGION`, `CIRCULAR_SOURCES`). The finding is that
the mapping is *copied*.

#### A4. The `pre_subsidy_electricity_expense` override is not proven load-bearing by any test
`snap_individual_utility_allowance.yaml` (both electricity cases: MO 2026, HI HH3 FY2025)

Both cases set only `pre_subsidy_electricity_expense: 1_200`. With no subsidy inputs,
`electricity_expense = max_(1_200 - 0, 0) = 1_200 > 0` as well — so swapping the override back to
`electricity_expense` leaves both cases green while re-introducing the circular reference the code
comment warns about. That cycle is real and was traced, not assumed:
`electricity_expense → ca_care → snap → snap_utility_allowance → snap_individual_utility_allowance
→ electricity_expense` (via `parameters/household/expense/utilities/subsidies.yaml` =
`[ca_care, ca_fera, ca_la_ez_save]` and `ca_care` categorical eligibility listing `snap`).

Pin it by overriding the post-subsidy variable directly (it is input-overridable, so its formula
will not run): same case plus `electricity_expense: 0`, still expecting 162. Worth a line in the
test file explaining why: all three real subsidies are California-only and CA's electricity
standard is 0, so **no naturally occurring household can distinguish the two variables by value** —
an input override is the only way to pin this.

#### A5. `.get(expense, expense)` silently couples parameter keys to the variable namespace
`snap_individual_utility_allowance.py:44-45`

```python
expense_variable = EXPENSE_VARIABLE_OVERRIDES.get(expense, expense)
incurs_expense = spm_unit(expense_variable, period) > 0
```

Before this PR the keys in `utility/single/utility_types.yaml` indexed **only** the parameter tree,
so adding a utility type required only a matching parameter node. After this PR every key must
**also** name a real SPMUnit variable (or appear in the overrides dict), or the model raises
`VariableNotFound` at runtime for every household. That is a new failure mode created by an
implicit convention with no validation and no warning comment. Either make the mapping total
(explicit entry per key, `[]` lookup instead of `.get`, so a missing entry fails loudly at the
source) or document the requirement on `utility_types.yaml`.

#### A6. No `reference` attribute on the Variable, although a CFR citation is the whole point of the PR
`snap_individual_utility_allowance.py:13-19` (class body); citation currently lives only as a `#`
comment at `:39-40` and a YAML header comment at `snap_individual_utility_allowance.yaml:1-2`

Sibling variables in the same directory use the idiomatic slot:
`/Users/pavelmakarchuk/policyengine-us/policyengine_us/variables/gov/usda/snap/income/deductions/snap_net_income_pre_shelter.py:11`
(bare string) and `.../shelter/snap_excess_shelter_expense_deduction.py:12` (tuple). The
model-development guidance is `reference`, never `documentation`, so the citation reaches the API
and docs. Suggested:

```python
reference = "https://www.ecfr.gov/current/title-7/section-273.9#p-273.9(d)(6)(iii)(A)"
```

Applying this also makes the in-formula comment at `:39-40` redundant (it restates the rule the
code implements), satisfying the style rule against such comments.

#### A7. Regions whose only nonzero individual standard is telephone now yield $0, with no actual-cost fallback
`snap_individual_utility_allowance.py:46,54`; parameters
`utility/single/{electricity,gas_and_fuel,water,sewage,trash}.yaml` — 7 CFR 273.9(d)(6)(iii)(D)

Several regions carry a telephone standard only, with 0 for the other five (NC is clearest:
electricity/gas_and_fuel/water/sewage/trash = 0 at every date, phone = 41 in FY2025; CA is the
same shape). Under the old code an NC water-only household received $41 — the phone standard, to
which it was plainly not entitled. Under the fix it receives **$0**.

$0 is the correct *standard*, but 273.9(d)(6)(iii)(D) gives such a household the right to claim
"verified actual utility costs for any allowable expense identified in paragraph (d)(6)(ii)(C)" —
and the model has **no actual-cost path for utilities at all** (`snap_utility_allowance` is
`adds = [SUA, LUA, IUA]` only). The fix is right; it makes a pre-existing modeling gap bite for
the first time. Worth an explicit note in the PR body or a follow-up issue so it is not later
mistaken for a regression of *this* PR. Scope note: NC's `limited/active.yaml` is `True`, so NC
two-bill households route to LUA and are unaffected — the exposure is single-bill households in
telephone-only-standard regions.

The test-side corollary: the three phone cases (CA 2022-01, CA 2024-01, NC 2025-01) return
19/19/41 under **both** the old and the new formula, so they carry no regression value. One case
making the zero-standard behaviour explicit would (CA 2024-01, `pre_subsidy_electricity_expense`
plus `phone_expense`, expecting 19 — electricity standard 0, phone 19).

#### A8. `gas_and_fuel` is the only utility type with no isolating test case
`snap_individual_utility_allowance.yaml` — gas appears only inside the two HI/GU all-six bundles

Every other type has (or shares) a case where it is the sole contributor. A gas-only case in a
flat (non-HH-size) region is also the cleanest proof that the `gas_and_fuel_expense → gas_expense`
override reads the right variable via the `where(is_hh_size_state, hh_val, flat_val)` → `flat_val`
branch. Suggested: MO 2026-01, `gas_expense: 2_400`, expecting **162**. (The override cannot
silently regress — `gas_and_fuel_expense` does not exist as a variable anywhere in the repo, so
dropping the mapping raises `VariableNotFound` on every case — but the value-level pin is still
missing.)

#### A9. No end-to-end test: the routing chain and the downstream benefit effect are both unpinned
`snap_individual_utility_allowance.yaml` (type is an input in all 14 cases);
`/Users/pavelmakarchuk/policyengine-us/policyengine_us/tests/policy/baseline/gov/usda/snap/income/deductions/snap_excess_shelter_expense_deduction.yaml:13,23,86`
injects `snap_utility_allowance` as an input;
`/Users/pavelmakarchuk/policyengine-us/policyengine_us/tests/policy/baseline/gov/usda/snap/integration.yaml`
contains no utility case at all

No test anywhere runs `expense inputs → count_distinct_utility_expenses →
snap_utility_allowance_type → snap_individual_utility_allowance → snap_utility_allowance →
snap_excess_shelter_expense_deduction → snap`. Given A1 (the household-level drop is real and can
reach the full shelter cap), that downstream drop is entirely unasserted. One integration case
(MO trash-only household → `count_distinct_utility_expenses: 1`, type `IUA`, allowance 162 not
891, plus the computed `snap`) plus its mirror image (same household with `heating_cooling_expense`
added → type `SUA`, individual allowance 0) would pin both the routing and the untouched-SUA
claim.

---

### Suggestions

#### S1. `> 0` boundary, negative, and tiny-positive cases are unpinned
`snap_individual_utility_allowance.py:45`; test file has only the defaults-based zero case

All were run against the snapshot during review and behave correctly; none are in the shipped
file: negative `trash_expense: -240` → **0** (and `count_distinct_utility_expenses` uses the
identical `> 0` test, so routing and amount agree — no divergence); explicit `0` → 0;
`trash_expense: 0.12` ($0.01/month) → the **full** monthly standard of 36. That last one is a real
cliff — any positive amount, however tiny, buys the whole standard — and is how standard
allowances are meant to work, but it deserves a pinned case. The negative case matters most: a
future `!= 0` or `abs()` refactor would currently slip through.

#### S2. `absolute_error_margin` drifts from the currency convention, and one margin is masking a real gap
`snap_individual_utility_allowance.yaml:5,18` (margin 1, expected 19 — exact); `:78,95` (margin 1);
`:111` (margin 1, expected 366 — exact, and a **new** case); `:125,142` (margin 3); margins absent
entirely on the new cases at `:23,33,43,54,64,167`

Convention for currency is `0.01`. Margin 1 on an expected 19 or 366 would hide a whole-dollar
error. The new HI electricity-only case at `:109-119` resolves to a flat parameter value of exactly
366 and should be `0.01`. The GU margin of 3 is doing **real work**: hand-summing FY2025 GU HH1
from the explicitly dated values gives 188+36+38+28+30+28 = **348** vs the expected **350**
(HH2: 390 vs 392), the $2 gap coming from uprating filling missing 2024-10-01 entries for
sewage/trash/phone. Either pin the actual uprated values and tighten to 0.01, or state the uprated
per-utility numbers in the comment so a future reader can reproduce 350/392. Also pick one
convention — present on some cases, absent on others.

#### S3. Case naming and readability
`snap_individual_utility_allowance.yaml:3,13,23,33,43,54,64,76,93,109,123,140,156,167`

- `:156` **"NC TUA HH size 1 in FY2025"** — the input is `snap_utility_allowance_type: IUA`, and
  `TUA` is not a `SNAPUtilityAllowanceType` member. Pre-existing, but the PR edited the adjacent
  line (`:163`), so it is cheap to fix in passing.
- "Test in 2022 with phone expenses" / "Test in 2024 with phone expenses" were left un-renamed
  while their neighbours were rewritten, and neither names the region (CA) that makes the expected
  19 meaningful.
- The repo convention is `Case N, description.`; none of the file (old or new) follows it. The new
  descriptive names *are* a readability improvement over the old ones and match the file's own
  style, so this is a whole-file convention question, not a regression introduced here.
- MO phone-only uses `phone_expense: 972` → 972/12 = **81/month**, numerically identical to the
  expected allowance of 81. Coincidence, but it invites a reader to think the allowance tracks the
  expense; e.g. `600` (50/month) with expected 81 removes the ambiguity.

#### S4. `MAX_SPM_SIZE = 10` is a hard-coded literal and its clamp is untested
`snap_individual_utility_allowance.py:32-33`; HI/GU tested only at sizes 1, 2, 3

A Category-1 pattern violation (a literal other than 0/1/2 in a formula). It is a structural bound
on the parameter table rather than a policy amount and does match the table
(`utility/single/by_household_size/electricity.yaml` tops out at key `10`), but it will silently
truncate if USDA publishes an 11-person column. **The line is untouched by this diff** — the
code-pattern role rated it SHOULD ADDRESS while explicitly scoping it as "not a blocker for PR
9344"; it is recorded here as a suggestion because it falls outside the PR-changes-only review
scope. The `max_(1, min_(10, spm_size))` clamp is likewise unexercised in either direction; a
size-12 HI case (all six FY2025 size-10 standards: 609+609+131+97+97+42 = 1,585) and a
`spm_unit_size: 0` case expecting the size-1 result would close both ends.

#### S5. No isolating case for a household-size-varying **non-electricity** utility
`snap_individual_utility_allowance.yaml` — HI HH3 electricity-only (366) is the only HH-size-branch
isolate

A water-only HI case (`spm_unit_size: 3`, `water_expense: 240`, FY2025 → **81**) would prove the
`hh_size_utilities` set is applied per-utility rather than only to electricity.

#### S6. `period` vs `period.this_year` on the incurrence read — correct today; do not "align" the two variables blindly
`snap_individual_utility_allowance.py:45`

The six expense variables are all SPMUnit / `definition_period = YEAR` / `value_type = float`, so
core classifies them `QuantityType.FLOW` (`policyengine_core/variables/variable.py:223-235`) and
`spm_unit(var, period)` from this MONTH formula routes through `calculate_divide`
(`policyengine_core/simulations/simulation.py:788-793`), returning **annual ÷ 12**. Division by 12
preserves sign exactly, so `> 0` is unaffected and the result is correct.

Two roles reached the same mechanics with slightly different advice, worth recording together.
The code role notes `period.this_year` would express the value the formula actually means (an
annual incurrence flag) and would remove the hazard of a later edit reusing
`spm_unit(expense_variable, period)` as a dollar amount at monthly scale. The regulatory role
notes the precedent is subtly *different* rather than identical:
`count_distinct_utility_expenses` is `value_type = int` → STOCK → read at `period.this_year` with
**no** division. Both code paths are correct; the two sides of this rule simply evaluate
incurrence at different scales, with identical outcomes purely by sign invariance. **Do not
"fix" one to match the other** on the assumption they are the same mechanism.

#### S7. The `heating_cooling` asymmetry is not a defect — but it deserves one clarifying comment
`count_distinct_utility_expenses.py:13` includes `heating_cooling`;
`utility/single/utility_types.yaml` has no heating/cooling entry and this formula does not iterate
one

All three roles converged on "correct as written", on two independent grounds:
(1) `snap_utility_allowance_type.py:27-38` evaluates `has_heating_cooling_expense | always_sua`
**first** in the `select`, so any unit with a heating/cooling expense is typed SUA and never
reaches the IUA branch — within the IUA population `heating_cooling_expense` is identically 0 and
contributes nothing to the count either; (2) regulatorily, 273.9(d)(6)(iii)(C) directs households
incurring heating/cooling to a standard *with* a heating or cooling component (the HCSUA), not to
an individual standard, so there is no standard to add. There is also no dead-zero hole: IUA
requires `distinct_utility_bills > 0` with `heating_cooling = 0`, so at least one of the six mapped
types is positive.

It is nonetheless exactly the kind of one-entry divergence that gets "fixed" incorrectly later, and
it is invisible unless you read both files. A one-line note in the `EXPENSE_VARIABLE_OVERRIDES`
comment block would prevent that.

#### S8. `gas_expense` does not aggregate the non-heating fuel components (pre-existing)
`snap_individual_utility_allowance.py:9` (`gas_and_fuel_expense → gas_expense`) —
7 CFR 273.9(d)(6)(ii)(C)

(d)(6)(ii)(C) covers "electricity or fuel used for purposes other than heating or cooling". The
model has `bottled_gas_expense`, `metered_gas_expense`, `fuel_oil_expense`, `coal_expense` and
`cooking_fuel_expense`, but `gas_expense` is a bare input with no formula and does **not**
aggregate them — all five were grepped and are referenced only by IL AABD
(`gov/states/il/dhs/aabd/payment/utility/utility_types.yaml`) and state LIHEAP variables. A
household whose only fuel input is `metered_gas_expense` plus a water bill now gets the water
standard alone, where the regulation would also entitle it to the gas/fuel standard. Not
introduced by this PR — `count_distinct_utility_expenses` has the same blind spot, so such a
household was already mis-routed, and the old all-six sum only covered it by accident. Follow-up
(make `gas_expense` add the fuel components, or extend both lists together), not a blocker.

#### S9. Pre-subsidy electricity: rationale verified sound; one narrow regulatory caveat recorded, no change recommended
`snap_individual_utility_allowance.py:3-10,44`

`ca_care_amount_if_eligible` is `p.discount * pre_subsidy_electricity_expense` — a percentage
**rate discount**, so the household is still billed and still pays; under 273.10(d)(1)(ii) it
"makes a money payment for the service" and under 273.9(d)(6)(iii)(A) it incurs an electricity
expense. Reading pre-subsidy is therefore both the cycle-breaker and the better regulatory answer
in the common case, and it adds no new dependency edge (that variable is already pulled in at
MONTH via `snap_utility_allowance_type`). For the record: 273.10(d)(1)(i) does disallow "a utility
expense which is reimbursed or paid by an excluded payment", and a flat-dollar subsidy (LA EZ
Save) could in principle exceed a very small bill, making pre-subsidy `> 0` marginally
over-inclusive. That case is rare, is not modelable without reintroducing the cycle, and cuts the
opposite way from 273.9(d)(6)(iii)(C)'s explicit LIHEAA carve-out. **No change recommended.**

---

### Regression history — will a third revert be caught?

PR **#3867** originally applied this rule; PR **#3903**, a decomposition refactor, reverted it by
rewriting the formula from scratch. The structural cause of that revert was **the test suite, not
the code layout**: before this PR the test file asserted `snap_individual_utility_allowance: 380`
for an AK_C **trash-only** household, with a comment spelling out
"elec=110 + gas=114 + water=56 + sewer=55 + trash=30 + phone=15". The tests actively *encoded the
bug*, so #3903's rewrite passed CI unchallenged.

**This PR fixes that, and that is the real safeguard.** Six assertions now pin the correct
behaviour and fail loudly under the sum-all-six formula, and the comments say "not the sum of all
six standards". A fourth-generation rewrite that reintroduces the sum cannot pass CI.

Two residual gaps a maintainer should weigh, both listed above:
- **A2** — the outer `where(allowance_type == IUA, …, 0)` is still unpinned, so a rewrite that
  drops the type gate (causing SUA/LUA double-counting in `snap_utility_allowance`) would still
  pass the whole suite.
- **A3** — the type→variable mapping remains duplicated, so a future edit to
  `count_distinct_utility_expenses` alone could desync *routing* from *amount* (household typed
  IUA, allowance $0) with no test failure.

Closing A2 and A3 is what would make the invariant, rather than just the arithmetic,
regression-proof.

---

### PDF Audit Summary

**Not Applicable.** This PR cites no PDF or URL source documents and changes no parameter values,
so there was **no PDF phase** in this review. The only citation added is 7 CFR
273.9(d)(6)(iii)(A), as code and test comments. The regulatory baseline was instead established
from primary legal text pulled verbatim from GPO/govinfo XML
(`CFR-2023-title7-vol4-sec273-9.xml` and `-sec273-10.xml`; eCFR was returning a 302 to an unblock
page) and cross-checked against Cornell LII. Phase 5E (page-anchor check) is likewise N/A — no
`#page=` references are added.

- PDFs downloaded: 0
- Parameter values audited against a source document: 0 (no parameter files changed)
- XREF / EXT / MISMATCH items: 0
- Page-anchor checks: 0 (N/A)

---

### Validation Summary

| Check | Result |
|---|---|
| CRITICAL findings | **0** — all three roles independently reported none |
| Regulatory conformance (7 CFR 273.9(d)(6)(iii)(A), (C), (D), (E); 273.10(d)) | PASS — formula matches the rule as researched from govinfo XML, and is *better* than the #3867 version it restores (additive vs. first-match `select`) |
| Test arithmetic reproduced from parameter YAML | PASS — all 14 cases; recomputed independently by the regulatory, code, and test roles; three-way agreement, no discrepancy |
| Regression pinned (old sum-all-six formula fails) | PASS — 6 discriminating cases |
| CI | PASS — all 33 checks green |
| `ruff format --check` on the changed variable | PASS (clean) |
| `policyengine-core test` on the changed YAML | PASS (14 passed) |
| Changelog fragment | PASS — `changelog.d/fix-snap-iua-filter-incurred.fixed.md`, top-level, name matches branch (`/`→`-`), `fixed` is the right type for a behaviour correction |
| Hard-coded values introduced by the diff | PASS — none; only `0` (accumulator seed) and the `> 0` comparison. Pre-existing `MAX_SPM_SIZE = 10` is S4 |
| Vectorization | PASS — the `for` loop iterates a static six-element **parameter** list, never household data; `incurs_expense * allowance` broadcasts elementwise; the `if util_name in hh_size_utilities` branch is on a loop constant, not entity data |
| Iteration determinism | PASS — `expense_types` is a YAML sequence (file order), not a dict/set; summation order cannot move the result at these magnitudes |
| Entity levels | PASS — SPMUnit throughout; `snap_utility_region_str` correctly reached via `spm_unit.household(...)` |
| Period usage | PASS — all test periods `YYYY-01`; FY26 cases at `2026-01` correctly resolve past the 2025-10-01 effective date; in-formula YEAR-at-MONTH read is sign-preserving (see S6) |
| Reinvented-variable check | PASS — grepped before flagging; no existing six-type mask or per-type allowance helper exists; `bool * amount` has in-program precedent (`snap_emergency_allotment.py:29`, `snap_min_allotment.py:28`) |
| Microsim / population impact | **Measured, not inferred: nil.** SNAP delta $0.0000B, 0 weighted units changed, IUA branch 0% of units. Cause: utility expense variables are empty groups in `enhanced_cps_2024.h5` and absent from `populace_us_2024.h5`. Household-calculator impact is real (see A1) |
| Branch staleness | PASS — BEHIND=0, AHEAD=2. **The branch is current; no rebase needed.** |
| Partner contract tests | Not touched — no files under `tests/policy/baseline/partners/**` in the diff |
| **Reference-checker (scope adaptation)** | **Not run as a separate role.** This PR adds no parameter files and no YAML `reference:` entries, so the reference-checker was **folded into the regulatory-reviewer**. The regulatory role verified the rule text from primary sources and reproduced all parameter arithmetic. The *absence* of a `reference` attribute on the Variable class is reported as A6 |
| **PDF audit (scope adaptation)** | **No PDF phase ran.** No source documents are cited and no parameter values changed, so there was nothing to audit against. See PDF Audit Summary — Not Applicable |
| Phase 5 verification queue | **NONE** — nothing required escalation. Both potentially-escalating items (the microsim impact and the `heating_cooling` asymmetry) were settled with direct evidence by the delegate roles |

---

### Review Severity: COMMENT

No CRITICAL findings, and the change is regulatorily correct and well tested arithmetically — but
nine SHOULD ADDRESS items go beyond style. The two that most warrant a maintainer's attention
before merge are **A1** (the PR body states neither the nil aggregate impact nor the large
household-calculator impact, which is precisely what a partner triaging an API behaviour change
needs) and **A2** (the type-gate line is unpinned, leaving the door open to the same class of
silent regression that produced #3903). Neither blocks correctness of the code as written.

---

### Next Steps

To auto-fix issues: run the fix-pr workflow for this PR.

Suggested ordering if fixing selectively:
1. **A1** — PR-body/disclosure only, no code change; highest value per unit of effort.
2. **A2**, **A8**, **A4** — three small, mechanical test additions that close the highest-value
   coverage gaps (type gate, gas isolate, pre-subsidy override).
3. **A6** — one-line `reference` attribute; also retires the redundant in-formula comment.
4. **A7** — a note or follow-up issue so the telephone-only-region $0 result is not later
   mistaken for a regression of this PR.
5. **A3**, **A5**, **A9** — larger refactor / integration-test work; reasonable as follow-ups.
6. **S1–S9** — optional polish and follow-up issues. S8 (fuel aggregation) and S4
   (`MAX_SPM_SIZE`) are pre-existing and belong in their own issues, not this PR.

PR #9375, opened 2026-09-02
Fix container metadata propagation

### LINKED ISSUE #5208: Propagate unit and period metadata to child parameters
The original example was `gov.irs.deductions.itemized.salt_and_real_estate.cap`: the YAML declares `unit: currency-USD` and `period: year` on the filing-status container, but its value-bearing children receive neither field because the container omits `propagate_metadata_to_children: true`. Household API serializes the child metadata directly, so clients receive `unit: null` and cannot format the values correctly.

An initial repository-wide scan on 2 September 2026 found 676 breakdown parameter files with this problem, accounting for 58,237 generated parameter leaves. A subsequent scan of all 102,894 processed leaf parameters found 25 additional containers that declare a unit or period without passing it to their children. The additional scan did not require the parent to declare `metadata.breakdown`.

This issue covers the 700 containers that can be corrected without changing parameter values:

- Enable `propagate_metadata_to_children` on 696 containers.
- Define unit and period explicitly on children of four containers where broad propagation would also activate a currently unapplied parent uprating rule.
- Verify that the SALT cap filing-status children expose `currency-USD` and `year` after the normal parameter-processing pipeline.
- Keep all loaded parameter values and calculation behavior unchanged.

`gov.usda.snap.asset_test.limit` is the one identified propagation case excluded from this issue. Its parent contains an uprating rule, so enabling broad propagation would generate additional future values. Selective propagation using an explicit deny-list is tracked in PolicyEngine/policyengine-core#537; applying that capability to SNAP will be separate follow-up work.

Containers whose parent metadata is missing or malformed cannot be corrected through propagation. That audit is tracked in #9377.

anth-volk:
Fixes #5208

## Summary

- propagate units and periods from 696 container parameters to their value-bearing children
- define metadata explicitly for four containers where broad propagation would also activate previously unapplied uprating rules
- leave the SNAP asset-limit container unchanged pending selective metadata propagation in PolicyEngine/policyengine-core#537

Missing or malformed source metadata is outside this PR and tracked in #9377.

After this merges, adding `uprating` to any of the 696 containers marked for metadata propagation will immediately apply it to every child. Until strict validation is available in PolicyEngine/policyengine-core#537, review such changes for effects on projected values.

## Verification

- `make format`
- `uv run --no-sync pytest policyengine_us/tests/test_parameter_files.py -q` (7 passed)
- compared all 102,892 processed parameter value timelines with `main`; the serialized values and SHA-256 digest are identical


hua7450 (changes requested):
Requesting one small change before merge; the value-neutrality claim itself checks out.

**Requested change**

`gov/states/ny/tax/income/deductions/itemized/phase_out/start.yaml` is the one flagged container whose parent metadata has `uprating`. It is a no-op today because every child already carries the identical rule, but once the flag is on, the five per-child `uprating` blocks are redundant and should be removed here so the file has a single source of truth for the rule.

**Independent verification**
- Built the parameter tree from this branch and from its base and compared all 102,892 leaf timelines: zero value differences from this PR.
- Leaves missing `unit`: 59,257 → 531. Leaves missing `period`: 78,196 → 19,726. No leaf had a pre-existing unit or period overwritten to a different value.
- The 696 flagged containers include no folder-level `index.yaml` (so no subtree cascade) and no bracket scale (so no threshold-uprating side effect). The only extra keys that newly propagate are `breakdown_label(s)`.

**Non-blocking suggestion**

The three explicit-metadata files that keep a dead parent `uprating` line (AZ standard deduction, CO CollegeInvest cap, IRS exemption phase-out start) now read as "uprate" at the top while the children deliberately do not. A one-line YAML comment pointing at PolicyEngine/policyengine-core#537 would stop the next editor from adding the flag and changing values.

**Note for future editors**, worth a line in the PR description: after this merges, adding `uprating` to any of these 696 containers becomes live for every child immediately. That is the correct semantics, but it is a new way to change values by accident until the strict validation in core#537 lands.

anth-volk:
Addressed both review items in `35a8e0255b`:

- removed the five redundant child-level `uprating` blocks from the New York itemized-deduction phase-out threshold; the parent definition now provides the single rule propagated to each filing-status child;
- added the suggested PolicyEngine/policyengine-core#537 comment to the Arizona standard deduction, Colorado CollegeInvest cap, and federal exemption phase-out threshold containers;
- added the future-editor warning to the PR description.

`make format` and all seven parameter-file tests pass.

hua7450 (approved):

PR #541, opened 2022-02-05
Add California Clean Vehicle Rebate Project

MaxGhenis:
# New variable: `ca_cvrp`

- [x] Label field added
- [x] Documentation field added
- [x] Unit field added
- [x] Default value field added if relevant
- [x] Variable name follows conventions
- [x] Unit test(s) added
- [x] Integration test(s) added if relevant
- [x] Issues this PR fixes linked

Issue #542, opened 2022-02-06
California Clean Vehicle Assistance Program


MaxGhenis:
Up to $5,000 for a clean vehicle and up to $2,000 for home charging for families with income up to 400% of FPL: https://cleanvehiclegrants.org/faq/

$19 million combined outlays from Nov 2020 to Oct 2021, though this would have been higher if they hadn't run out funds (they're no longer taking people on the waitlist): https://cleanvehiclegrants.org/program-data/

This program appears to have the same structure, but only operates in certain counties: https://communityhdc.org/dcap/#Submit

MaxGhenis:
Done

PR #9392, opened 2026-09-04
Add policyengine-us dependency-map and publish it per release

PavelMakarchuk:
## Why

Downstream tools need to know which model variables a parameter path moves and which variables feed which: validation matching in the app (PolicyEngine/policyengine-app-v2#1180 matches a reform's provisions to scorecard programs through it, and the calibration-dashboard side is next), the calibration dashboard's model-coverage page (which currently derives the same edges by grepping formula source), and reform classifiers. Static scans miss bracket, scale, and vectorised reads, so the map is traced from the model at run time instead. The map is a fact about the model, so it ships with the model.

## What

- **`policyengine_us/tools/dependency_map.py`** runs simulations under policyengine-core's `FullTracer` and folds the trace trees into `readers` (parameter path → variables whose formula read it) and `consumers` (variable → variables whose formula read it). Bracket reads are recorded at the scale node (`p.base.calc(age)` → `gov.irs.credits.ctc.amount.base`). `gov.abolitions.*` switches are dropped as noise.
- **Populations.** `tests` (default) builds a simulation per baseline YAML test and calculates its outputs: deterministic, no data download. Tests in a file mostly vary inputs for the same outputs and record the same edges, so by default one test is kept per newly covered output variable in the file (5,338 of 27,593); `--every-test` traces all of them. `microdata` calculates every variable over a subsample of the default dataset (broader, but a `defined_for` formula only runs if someone in the sample qualifies); `both` unions them.
- **Provenance.** The output is stamped with the package version, git sha, the data-build fingerprint from `build_metadata.py` (the exact cache key: the map can only change when entities, parameters, variables, or system change), and the core version.
- **CLI.** `pyproject.toml` pointed the `policyengine-us` console script at `policyengine_us:main`, which does not exist. It now routes to `policyengine_us/cli.py`, whose first subcommand is `dependency-map`.
- **Workflow.** `.github/workflows/dependency-map.yaml` runs on the same version-bump commit that triggers Publish, but as its own workflow so a tracing failure never blocks a release. It traces the map, gzips it, and publishes it as a GitHub release asset for that version, creating the release if none exists. The release is created with `GITHUB_TOKEN`, which deliberately triggers no other workflow. Consumers fetch `https://github.com/PolicyEngine/policyengine-us/releases/download/<version>/dependency-map.json.gz`.

## policyengine-core workarounds

Two tracer gaps make the map near-empty without patches, filed as PolicyEngine/policyengine-core#541 (the yearly `ParameterNodeAtInstant` is cached before core's tracing recast, so yearly formulas record no parameter reads) and PolicyEngine/policyengine-core#542 (`TracingParameterNodeAtInstant` only records scalar leaves, so scale/bracket reads are invisible). Both are worked around here behind `CORE_TRACER_FIXED_VERSION`; set that constant once a core release ships the fixes and the patches switch off.

## Test

- `policyengine_us/tests/test_dependency_map.py` traces the CTC test directory and asserts the scale read, the yearly read, the downstream chain to `ctc` and `refundable_ctc`, and the abolition filter; plus selection and merge unit tests. ~45s.
- `uv run policyengine-us --help` and `uv run policyengine-us dependency-map` resolve through the new entry point.
- Full `tests` population run locally: numbers below.

## Local run

`policyengine-us dependency-map --population tests` on this branch (policyengine-us 1.808.0, core 3.30.2, M-series laptop):

| | |
|---|---|
| tests traced | 5,337 (1 failed to build) |
| wall time | 62 min |
| parameter paths with readers | 5,074 |
| variables with consumers | 5,497 |
| output size | 1.2 MB raw, 175 KB gzipped |

Against a `microdata` run over 2,000 households (8 min): 4,602 parameter paths appear in both, 472 only in the tests run, 125 only in microdata; 5,200 consumed variables in both, 297 only in tests, 210 only in microdata. `both` is the most complete map and costs the extra 8 minutes, but needs the dataset download in CI; the workflow defaults to `tests` until that is wired in (a `workflow_dispatch` input already accepts `both`).

## Follow-ups

- Set `CORE_TRACER_FIXED_VERSION` when core ships #541/#542.
- policyengine-app-v2: fetch the release asset for the API's model version at build time instead of committing the map.
- calibration-diagnostics: replace the static scan in `generate_model_coverage.py` with the asset.

### DIFF
--- a/.github/workflows/dependency-map.yaml
+++ b/.github/workflows/dependency-map.yaml
@@ -1,0 +1,67 @@
+# Trace the parameter -> variable dependency map for each released version
+# and publish it as a GitHub release asset, so downstream tools (the app's
+# validation matching, the calibration dashboard's model-coverage page) can
+# fetch the map that matches the model version they run against.
+#
+# Runs on the same version-bump commit that triggers the PyPI publish, but as
+# its own workflow: a tracing failure never blocks a release. The release is
+# created with GITHUB_TOKEN, which deliberately does not trigger other
+# workflows.
+name: Dependency map
+
+on:
+  push:
+    branches: [main]
+  workflow_dispatch:
+    inputs:
+      population:
+        description: "tests, microdata, or both"
+        default: tests
+        required: false
+
+permissions:
+  contents: write
+
+jobs:
+  trace:
+    name: Trace and publish
+    runs-on: ubuntu-latest
+    if: |
+      github.repository == 'PolicyEngine/policyengine-us'
+      && (github.event_name == 'workflow_dispatch'
+          || github.event.head_commit.message == 'Update PolicyEngine US')
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v6
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: 3.14
+      - name: Install uv
+        uses: astral-sh/setup-uv@v8.1.0
+      - name: Install dependencies
+        run: uv sync --extra dev
+      - name: Read version
+        id: version
+        run: |
+          VERSION=$(python -c "import re; print(re.search(r'version = \"(.+?)\"', open('pyproject.toml').read()).group(1))")
+          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
+      - name: Trace the dependency map
+        run: |
+          uv run policyengine-us dependency-map \
+            --population "${{ github.event.inputs.population || 'tests' }}" \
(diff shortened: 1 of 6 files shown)


PavelMakarchuk:
Closing in favour of doing this in policyengine-core: fix PolicyEngine/policyengine-core#541 and #542 properly and add a country-agnostic `dependency-map` command that takes the model package as an argument, so -us and -uk need no wrapper and no workaround. Branch kept for reference.

PR #9355, opened 2026-08-27
Remove the inert federal CCDF implementation

hua7450:
## Summary

Removes the inert federal CCDF implementation under `gov/hhs/ccdf/`. The "federal" encoding was legacy New York data (market rates, copay percentages, county clusters), and the whole subsystem contributed $0 to every household: `spm_unit_ccdf_subsidy` is not listed in `child_care_subsidy_programs.yaml` or any benefit aggregate. New York now has a proper state implementation (#9282) that reads none of these files.

Fixes #9283

## Changes

Commit 1 — inert NY legacy:
- Variables: `ccdf_age_group`, `ccdf_market_rate`, `ccdf_county_cluster`, `spm_unit_ccdf_subsidy`, `spm_unit_total_ccdf_copay`, `is_ccdf_home_based`, and the orphan input stubs `is_ccdf_initial_income_eligible`, `is_ccdf_continuous_income_eligible`
- Parameters: `amount.yaml` (NY market rates), `copay_percent.yaml`, `county_cluster.yaml`
- Their 5 test files under `tests/policy/baseline/gov/hhs/ccdf/`

Commit 2 — the eligibility chain:
- Variables: `is_ccdf_eligible`, `is_ccdf_income_eligible`, `is_ccdf_reason_for_care_eligible`, `ccdf_income_to_smi_ratio`, `ccdf_income`, and their 5 test files
- Prose touch-ups in NY CCAP and SD CCA files that named the deleted variables
- The intermediate CI run at this commit failed exactly on `Household API Partners`, confirming the partner contract tests guard this surface

Commit 3 — partner test update:
- Deleted `edge_cases/federal/childcare/ccdf/federal.yaml` and `ccdf/ma.yaml` (existed solely to contract-test the removed chain)
- `signatures/ma.yaml`: removed 20 `is_ccdf_eligible` output pins and 20 `is_ccdf_reason_for_care_eligible` input lines
- 7 Head Start edge-case files: removed 38 inert `is_ccdf_reason_for_care_eligible` input lines (nothing reads them; no expected outputs change)

## Partner impact

The API partner was informed and approved removing `is_ccdf_eligible` (appears in their requests as an output) and `is_ccdf_reason_for_care_eligible` (as an input).

## Kept — genuinely shared federal infrastructure

[...]

Issue #9283, opened 2026-08-14
Remove the inert federal CCDF implementation (New York data under gov/hhs/ccdf)


hua7450:
## Background
The "federal" CCDF implementation under `gov/hhs/ccdf/` is actually a legacy New York encoding: its market rates (`amount.yaml`), copay percentages (`copay_percent.yaml`), and county clusters (`county_cluster.yaml`) are New York OCFS data cited to NY documents. It is **inert on main** — `spm_unit_ccdf_subsidy` is not listed in `child_care_subsidy_programs.yaml` or any benefit aggregate, so it contributes $0 to every household.

Once New York has a proper state implementation (#9282), these files should be removed.

## Scope — delete
Parameters:
- `parameters/gov/hhs/ccdf/amount.yaml` (NY market rates, 938 lines)
- `parameters/gov/hhs/ccdf/copay_percent.yaml`
- `parameters/gov/hhs/ccdf/county_cluster.yaml` (NY counties)

Variables:
- `ccdf_age_group`, `ccdf_market_rate`, `ccdf_county_cluster`, `spm_unit_ccdf_subsidy`, `spm_unit_total_ccdf_copay`, `is_ccdf_home_based`
- `is_ccdf_initial_income_eligible`, `is_ccdf_continuous_income_eligible` (orphan input stubs with zero readers)

Tests:
- `tests/policy/baseline/gov/hhs/ccdf/`: `ccdf_age_group.yaml`, `ccdf_copay.yaml`, `ccdf_county_cluster.yaml`, `ccdf_market_rate.yaml`, `spm_unit_ccdf_subsidy.yaml`

Partner analytics-coverage tests do not reference any of these (verified by a green Household API Partners run with them deleted).

## Keep — genuinely shared federal infrastructure
- `is_ccdf_asset_eligible`, `is_ccdf_immigration_eligible_child`, `meets_ccdf_activity_test`, `is_ccdf_age_eligible`, `is_enrolled_in_ccdf`, `child_care_subsidies` + their parameters (`asset_limit`, `age_limit`, `income_limit_smi`, `child_care_subsidy_programs`) — used by many state programs.
- `ccdf_duration_of_care` — retained for Idaho ICCP (`id_iccp_time_category`), its only consumer; its hour thresholds originated in a NY market-rate letter, so it should eventually be re-homed under Idaho with its own parameters and citation.

## Out of scope
[...]

Issue #7766, opened 2026-03-12
Clean up NY-specific federal CCDF variables and parameters


hua7450:
## Summary

The federal CCDF directory (`gov/hhs/ccdf/`) contains a mix of truly federal variables and NY-specific implementations. No state CCAP implementation (MA, CO, TX, CA, NE) currently uses any of these federal variables. This issue tracks cleaning up the NY-specific files and making the federal layer actually reusable.

Related: #7764 (federal CCDF infrastructure), PR #7765

### NY-specific files (not generalizable)

**Variables:**
- `ccdf_market_rate.py` — NY market rates only
- `spm_unit_total_ccdf_copay.py` — NY county copay percentages
- `ccdf_county_cluster.py` — Maps NY counties to 5 clusters
- `ccdf_age_group.py` — NY OCFS age cutoffs (different for home vs center)
- `ccdf_duration_of_care.py` — NY OCFS duration definitions
- `spm_unit_ccdf_subsidy.py` — Uses NY-specific market rate and copay
- `is_ccdf_home_based.py` — Used by NY age group logic

**Parameters:**
- `amount.yaml` — NY market rates (5 clusters × 5 provider types × 4 durations × 4 age groups)
- `copay_percent.yaml` — All 62 NY counties with copay percentages
- `county_cluster.yaml` — NY county-to-cluster mapping
- `age_limit.yaml` — Only has NY-dated value (2021-10-01)

### Stub variables (no formulas)

- `is_ccdf_initial_income_eligible.py` — Empty, no formula
- `is_ccdf_continuous_income_eligible.py` — Empty, no formula
- `is_enrolled_in_ccdf.py` — Empty, no formula
- `meets_ccdf_activity_test.py` — Empty, intentionally a stub (states implement their own)
- `ccdf_income.py` — Only `adds = ["market_income"]`, undocumented

### Working federal variables (keep as-is)

- `is_ccdf_asset_eligible.py` — $1M federal limit, truly universal
- `is_ccdf_immigration_eligible_child.py` — Federal qualified alien list (added in PR #7765)
- `is_ccdf_age_eligible.py` — Under 13 federal default
- `is_ccdf_income_eligible.py` — 85% SMI federal ceiling
- `is_ccdf_eligible.py` — Composite eligibility
- `is_ccdf_reason_for_care_eligible.py` — Activity test or protective services
- `ccdf_income_to_smi_ratio.py` — Income/SMI ratio calculation
[...]

PR #9400, opened 2026-09-07
Fix CCDF age group crashing on households with adults

MaxGhenis:
## Summary

Calculating `spm_unit_ccdf_subsidy` (or `ccdf_market_rate`, or `ccdf_age_group` itself) crashed for any household containing a person aged 13 or older. For a Virginia household with two adults (35 and 33) and a 3-year-old, `Simulation(situation=...).calculate("spm_unit_ccdf_subsidy", 2026)` raised:

```
ValueError: Invalid value(s) ['0', 'CCDFAgeGroup.PRESCHOOLER'] for enum CCDFAgeGroup. Valid values are: ['INFANT', 'TODDLER', 'PRESCHOOLER', 'SCHOOL_AGE']
```

The exact text depends on who is listed first. A child listed first gives `AttributeError: 'int' object has no attribute 'index'`, and an adults-only household gives `Invalid value(s) ['0']`. The state path (`va_child_care_subsidies`) computed fine; only the federal generic path failed.

## Root cause

`ccdf_age_group` built its result with `select(conditions, [CCDFAgeGroup items])` and no `default`. `select` here is numpy's `np.select`, whose default is the integer `0`. For every person no condition matched (age ≥ 13, so every adult) the result was an object array mixing `0` with `CCDFAgeGroup` members.

policyengine-core's `Enum.encode` takes its Enum-item fast path only when the *first* element is an Enum. With an adult first, the array is cast to strings (`'0'`, `'CCDFAgeGroup.PRESCHOOLER'`) and the string lookup rejects both. With a child first, the fast path hits `AttributeError` on `0.index`.

Before policyengine-core 3.22.0 the string path encoded unmatched values as index 0 (INFANT) with no warning, so the adults-first case silently produced INFANT for adults and, because `str(CCDFAgeGroup.PRESCHOOLER)` is not a member name, for the children too. Core 3.22.0 added a warning and 3.23.0 (December 2025) made it raise. This repo requires core ≥ 3.30.1, which is what surfaced the bug. The child-first ordering has always raised.

## Fix

Drop the redundant `age < 13` branch and give the select an explicit `default=CCDFAgeGroup.SCHOOL_AGE`. Children under 13 get the same answer as before (verified over a dense age grid for both home-based and center-based care). Anyone 13 or older (the CCDF age limit in `gov.hhs.ccdf.age_limit`) now resolves to SCHOOL_AGE instead of an unencodable 0. `is_ccdf_age_eligible` already screens them out of the subsidy, so no subsidy amount changes for eligible children. `in_ccdf_age_group` already uses this pattern.

## Tests

- `ccdf_age_group.yaml`: age 13 boundary, an adult, and a mixed adult/preschooler household with the adults listed first.
- New `integration.yaml`: the reported Virginia household (computes, subsidy 0 with no childcare hours) and a Nassau County household with a preschooler in full-time center care ($310/wk × 52 = $16,120 market rate, $3,536 copay, $12,584 subsidy).
- All 71 tests under `policyengine_us/tests/policy/baseline/gov/hhs/ccdf` pass. All 5 new cases fail on unfixed `main` (04a961a) with the same ValueError: the three mixed-household cases with the exact message above, the two single-person cases with `Invalid value(s) ['0']`.
- A 55-person vectorised simulation with adults first across five states computes the whole chain without error.

## Related

- #9401 (commit 1 of #9355, split out) deletes this whole path; if it lands first, this PR closes. Issue #9283 and PR #9355 propose deleting the federal (legacy New York) CCDF path as inert; #7766 is the older tracking issue. This fix is small and self-contained so the crash stops for anyone calling these variables today. If #9355 lands first, this PR can simply close. If this lands first, #9355 needs a rebase that also deletes the new `integration.yaml`.
- Swept every Enum-valued variable for `select` calls without a `default` keyword: 111 calls, 5 without it. Three are exhaustive by construction (`pell_grant_household_type`, `state_filing_status_if_married_filing_separately_on_same_return`, `age_group`) and `il_personal_exemption_eligibility_status` passes its default positionally. `md_ccs_service_unit` would hit the same failure only for negative or NaN childcare hours, which its bracket maps to unit 0; left for a follow-up.

### DIFF
--- a/changelog.d/fix-ccdf-age-group-enum.fixed.md
+++ b/changelog.d/fix-ccdf-age-group-enum.fixed.md
@@ -1,0 +1,1 @@
+Fixed `ccdf_age_group` crashing on anyone aged 13 or older by defaulting them to school age, so `ccdf_market_rate` and `spm_unit_ccdf_subsidy` compute for households that include adults.
--- a/policyengine_us/tests/policy/baseline/gov/hhs/ccdf/ccdf_age_group.yaml
+++ b/policyengine_us/tests/policy/baseline/gov/hhs/ccdf/ccdf_age_group.yaml
@@ -43,3 +43,38 @@
     age: 10
   output:
     ccdf_age_group: SCHOOL_AGE
+
+# Everyone in the simulation gets an age group, not only eligible children.
+# Ages at or above the CCDF age limit (13) fall through to SCHOOL_AGE; without
+# an explicit select default they used to become an integer 0 that could not
+# be encoded as a CCDFAgeGroup, crashing every household with an adult in it.
+- name: Age 13, the CCDF age limit, still resolves to school age
+  period: 2026
+  input:
+    age: 13
+  output:
+    ccdf_age_group: SCHOOL_AGE
+
+- name: Adult resolves to school age rather than an unencodable value
+  period: 2026
+  input:
+    age: 35
+  output:
+    ccdf_age_group: SCHOOL_AGE
+
+- name: Household with two adults and a preschooler, adults listed first
+  period: 2026
+  input:
+    people:
+      worker:
+        age: 35
+      home:
+        age: 33
+      kid:
+        age: 3
+    households:
+      household:
+        members: [worker, home, kid]
+        state_code: VA
+  output:
+    ccdf_age_group: [SCHOOL_AGE, SCHOOL_AGE, PRESCHOOLER]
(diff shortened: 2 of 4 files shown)


MaxGhenis:
Superseded by #9401, which removed `ccdf_age_group` and the rest of the legacy New York CCDF path outright (merge commit f627b0b659). The crash this PR fixed no longer has a file to live in.

PR #9466, opened 2026-09-14
Allow explicit SPM calculator post-release consumers

### LINKED ISSUE #9465: Allow explicit SPM calculator post-release development consumers
The country requirement `spm-calculator==1.0.0` rejects an explicitly requested `1.0.0.post1` calculator, including the source-development consumer for PolicyEngine/spm-calculator#45. The dependency resolver reports an unsatisfiable requirement before the packages can be tested together.

Narrowly admit `1.0.0.post1` alongside `1.0.0` while retaining the default registry lock at `1.0.0`. This concerns dependency metadata compatibility only: source development still needs immutable commits for unpublished fixes, and managed data certification remains separate. Keep the existing wrapper bundle pins and country #9462 source qualification unchanged.

MaxGhenis:
Fixes #9465

An explicitly requested `spm-calculator==1.0.0.post1` conflicts with the country's unconditional `==1.0.0` requirement. Admit `>=1.0.0,<=1.0.0.post1` so either version can resolve. Keep the default calculator lock at 1.0.0; all 147 nonroot lock records remain unchanged.

This three-file change covers package metadata only. It does not include #9462, change wrapper bundle pins, or certify a data release. Development consumers of unpublished calculator fixes must still use an immutable source revision rather than a version label alone.

Validation:

- Actual uv 0.12.13 resolver fails before and succeeds after when constrained to 1.0.0.post1. Version checks admit 1.0.0/post1 and reject 0.3.1, post2, 1.0.1, and 2.0.0.
- 58 existing SPM cases and 25 threshold cases pass with Python 3.14.4, Core 3.30.2, Microdf 1.3.0, pandas 3.0.0, NumPy 2.4.1, and editable calculator source `bcf45768003bb79addfafb0e6d9c7d2d5e547d9d` from PolicyEngine/spm-calculator#45. Ordinary installation and dependency checking pass without an override. Runtime origins and the full package freeze were stable across the valid test runs.
- `make format`, Ruff checks, and the committed registry-lock guard pass. Independent review of the committed three-file diff found no actionable issues.

The initial local test run overlapped an environment sync and was discarded; the counts above are from the restored, complete reruns. No native population data was used. This remains a draft; managed-data compatibility and the separate #9462 CI failures are not resolved by this dependency change.


MaxGhenis:

PR #9478, opened 2026-09-15
Count MA TAFDC once in household benefits

### LINKED ISSUE #9470: MA TAFDC counted twice: in tanf and in household_state_benefits
**Observed** (API, 2026): for the household in the TAFDC formula issue at $26,000 of earnings, `tanf` = $9,880 **and** `household_state_benefits` = $9,880. Both feed `household_net_income`, so the grant is counted twice. Setting `tanf` to 0 as an input leaves the $9,880 in `household_state_benefits`.

**Cause**: `ma_tafdc` is listed in `gov/household/household_state_benefits.yaml` while it also reaches `tanf` (the federal aggregate in `household_benefits`).

**Expected**: counted once. Either drop `ma_tafdc` from `household_state_benefits` or stop it flowing into `tanf`, whichever matches how other state TANF programs are handled.

**Repro**: same request as the formula issue, adding `household_benefits: null` and `household_state_benefits: null` on the household.

mikewolfd:
Fixes #9470

`ma_tafdc` is one of the state TANF programs summed into `tanf`, which `household_benefits` already counts, and it was also listed in `household_state_benefits`, which `household_benefits` counts as well — so an MA household receiving the grant had it added twice to household net income. No other implemented state TANF program (`il_tanf`, `ca_tanf`, `ny_tanf`, `co_tanf`, `dc_tanf` and the rest of `STATE_TANF_VARIABLES`) appears in that list, so it is removed from all four dated lists, with a test that the grant reaches `household_benefits` exactly once and that `ma_eaedc`, which is not a TANF program, still flows through `household_state_benefits`.

Tests: `uv run policyengine-core test policyengine_us/tests/policy/baseline/household/income/household -c policyengine_us` — 37 passed; `... partners gov/states/ma` — 1160 passed.

https://claude.ai/code/session_018deKYfPGDB3bQneeC5Ww8z


hua7450 (approved):
Verified the double count on main: `ma_tafdc` is in `STATE_TANF_VARIABLES` (summed into `tanf`) and was also listed in all four dated blocks of `household_state_benefits`, both of which feed `household_benefits`. Regression traces to 2b1518c (2026-02-16), which folded MA into the TANF aggregate without pruning the state benefits list. Removing it here is consistent with how every other state TANF program is handled, and `ma_eaedc` correctly stays. CI green.

PR #9486, opened 2026-09-15
Test latest compatible PolicyEngine Core

### LINKED ISSUE #8753: CI can miss failures against newer policyengine-core because uv.lock pins stale core
## Summary

Household API staging deployment tests recently surfaced a `policyengine-core` compatibility failure that `policyengine-us` CI did not catch. The likely reason is that `policyengine-us` test jobs install from the committed `uv.lock`, which currently pins an older `policyengine-core`, while downstream consumers can resolve newer compatible `policyengine-core` versions from the package's lower-bound dependency range.

## Observed downstream failure

The household API staging tests failed with:

```text
The formula for 'is_ssi_recipient_for_medicaid' called numpy.random.seed(), but rules-engine formulas must be deterministic...
```

That failure appeared in the household API deployment path after resolving newer package versions, including a newer `policyengine-core` version than the one used by the locked `policyengine-us` test environment.

Related temporary downstream pin:

- PolicyEngine/policyengine-household-api#1575
- PolicyEngine/policyengine-household-api#1576

## Why this matters for policyengine-us

`policyengine-us` declares a compatible lower-bound dependency on `policyengine-core`, but normal CI runs appear to use `uv sync --extra dev`, which respects the committed lock file. In the current checkout, `uv.lock` pins `policyengine-core==3.26.0`, so PR/push tests can pass against that stale core even when downstream installs resolve a newer compatible `policyengine-core` that changes runtime behavior.

This means `policyengine-us` CI may not catch regressions introduced by newer `policyengine-core` versions until a downstream application, such as `policyengine-household-api`, installs and exercises that newer core.

## Suggested fixes

- Add a CI job that tests against the latest compatible `policyengine-core`, independent of the committed lock file.
- Consider a scheduled lock refresh/upgrade workflow that validates the package against upgraded dependencies.
- Add a focused regression test for the Medicaid/SSI path that triggered the household API failure, especially around `is_ssi_recipient_for_medicaid`.
- Consider porting or mirroring relevant household API partner/customer-input fixtures into `policyengine-us` when they exercise package behavior that normal YAML tests miss.
- If newer `policyengine-core` behavior is intentionally incompatible, add an upper bound or compatibility gate until `policyengine-us` is updated.

## Notes

The failing formula itself does not obviously contain a direct random seed call in the current source. This issue is about the test-environment gap first: `policyengine-us` should catch compatibility failures against newer allowed `policyengine-core` versions before they reach downstream deployments.

anth-volk:
Fixes #8753

## Summary

- Resolve the latest compatible policyengine-core before dependency installation in the existing pull-request and post-versioning test jobs, keeping the current parallel test split.
- Remove the separate compatibility jobs so Python and household API partner suites run only once per workflow.
- Preserve committed-lock installs for candidate-wheel builds and publication. No scheduled workflow is added.

## Testing

- make format
- 43 release-workflow tests passed (2 registry probes skipped locally)
- actionlint passed for both workflows, excluding two existing warnings in unrelated pull-request steps
- uv lock upgraded policyengine-core from 3.30.2 to 3.32.5; 10 focused Medicaid tests passed on 3.32.5

The committed uv.lock remains unchanged. Test jobs resolve Core at run time; builds continue to use the committed lock.


hua7450 (commented):
Could we also add `uv lock --upgrade-package policyengine-core` before the install step in `BundleMetadataContract`?

This job validates actual runtime metadata, including Core’s metadata, against the bundle contract. The regular `Rest` job skips this integration test because `policyengine_bundles` isn’t installed there, so it doesn’t cover bundle compatibility with newer Core.

I verified the proposed steps locally on macOS with Python 3.14.6, Core 3.32.5, and the workflow’s pinned bundles commit: **all four tests in `test_build_metadata.py` passed without skips**. This would extend coverage without adding another runner.

hua7450 (approved):
Thanks for adding the Core upgrade to BundleMetadataContract. The requested coverage gap is addressed, and its Ubuntu CI job passed all four metadata tests without skips on Core 3.32.6. No remaining review concerns. The Rest (Python + variables) check is still running and should pass before merging.

PR #9500, opened 2026-09-16
Defer directly changed reform tests out of Quick Feedback

MaxGhenis:
## Problem

Quick Feedback defers the `policyengine_us/tests/policy/reform` directory to the per-file Reform job, because each reform YAML force-applies a reform and deep-copies the full parameter tree (5.5 to 8 GB per file, per the Makefile note). But a directly changed reform test file is not covered by that rule: `limit_test_paths` only subtracts exact directory matches, and the direct-file fallback keeps every changed test file. #9497 added four reform YAML tests; its Quick Feedback job ran the first one under coverage and the runner received a shutdown signal (run 35042783076, twice), while the per-file Reform job passed all four.

## Change

- Files under `policyengine_us/tests/policy/reform/` are deferred like the directory, in both the mapped targets and the direct-file fallback, with a log line saying so.
- Unit test: a changed reform module plus a changed reform YAML yields only the module's contrib directory, never the reform file.

## Verification

- `pytest policyengine_us/tests/core/test_run_selective_tests.py`: 13 passed.
- `ruff format --check` and `ruff check` on both files: clean.


MaxGhenis:
Closing in favor of #9501. The failed Quick Feedback job on #9497 (job 104653005722) died inside the `policy/contrib/congress/tlaib` directory target, before any of the directly changed reform file targets started, so deferring the reform files was the wrong lever. #9501 defers the congress contrib directories and keeps file targets running.

PR #9503, opened 2026-09-16
Count all child care subsidies in household benefits

### LINKED ISSUE #9405: household_state_benefits omits 28 of 51 implemented state child care subsidy variables (child care assistance never reaches net income in those states)
## Summary

`gov.hhs.ccdf.child_care_subsidy_programs` lists 51 state/DC child care subsidy variables, but `gov.household.household_state_benefits` (the list that `household_state_benefits` → `household_benefits` → `household_net_income` sum) includes only 23 of them in every period block, including 2026. In the other 28 states an implemented child care subsidy never reaches household benefits or net income, so the household app, the marriage calculator, and any microsimulation that reads `household_net_income` show zero child care assistance there.

Follow-up to #9234 (closed) and #9236, which added NC, DC and NY. Verified on `main` at 1ca1d4ee (2026-09-07) by parsing both YAML files.

## Present in `household_state_benefits` (23)

al ar az ca co dc de ga id in md mo mt nc ne nv ny oh ok sc va wa wv

## Implemented but missing from `household_state_benefits` (28)

ak ct fl hi ia il ks ky la ma me mi mn ms nd nh nj nm or pa ri sd tn tx ut vt wi wy

## Per-period counts in `household_state_benefits.yaml`

| Block | entries | `*_child_care_subsidies` |
|---|---|---|
| 2023-01-01 | 54 | 16 |
| 2024-01-01 | 60 | 22 |
| 2025-01-01 | 61 | 23 |
| 2026-01-01 | 62 | 23 |

## Suggested fix

Either add the 28 variables to every applicable block of `household_state_benefits.yaml`, or replace the per-state entries with the single aggregate `child_care_subsidies` (which already `adds` `gov.hhs.ccdf.child_care_subsidy_programs`) and remove the 23 per-state entries to avoid double counting. The second option keeps the two lists from drifting again; a test that asserts every `child_care_subsidy_programs` entry is reachable from `household_benefits` would lock it in.

Context: needed for a PolicyEngine analysis of the HHS "parent-based child care" draft rule (Sept 2026), where CCDF amounts must flow into net income consistently across all states.

RajveerKapoor:
Fixes #9405.

## Problem
household_state_benefits listed only 23 of the 51 state child care subsidy variables in the CCDF aggregate. Implemented subsidies in the other states therefore never reached household benefits or household net income.

## Fix
Replace the per-state child care entries in household_state_benefits with the existing child_care_subsidies aggregate in every period. This makes the CCDF list the single source of truth and removes the existing direct entries to prevent double counting.

## Tests
- Added an Alaska regression showing a previously omitted subsidy reaches household state benefits.
- Added a North Carolina regression showing a previously direct-listed subsidy is counted exactly once.
- Verified all 51 programs in the 2025 CCDF list reach child_care_subsidies and household_state_benefits exactly once.


mikewolfd:
Independent confirmation of the problem this fixes, and one observation about its size.

We (HotGap, a benefits-cliff calculator on PolicyEngine) hit #9405 from the outside on 2026-09-15 against the hosted API (policyengine-us 1.764.6). The omission is worse than a missing line item: because `childcare_expenses` is defined net of `child_care_subsidies`, an omitted state's household looks **poorer** for having the subsidy — SNAP's dependent-care deduction shrinks, and the subsidy that ate it never arrives in `household_net_income`.

Connecticut, single parent, one 3-year-old, $9,600/yr of care, $25,000 of pay, `meets_ccdf_activity_test: true`:

| | `ct_child_care_subsidies` modeled | forced to 0 |
|---|---|---|
| `child_care_subsidies` | 8,850 | 0 |
| `household_state_benefits` | **0** | 0 |
| `snap` | 2,577 | 6,559 |
| `household_net_income` | **34,121** | **38,102** |

Colorado (in the list) is the control: the same household at $25,000 shows `household_state_benefits` 8,913 = the subsidy to the dollar, and net income moves by the subsidy when `co_child_care_subsidies` is forced to 0 (45,226 → 39,944).

On the diff itself: I pulled the branch and checked the parameter — the aggregate is in all four periods (2023–2026) and none of the 84 per-state entries survive, so the double-count the PR description guards against can't happen. Swapping to `child_care_subsidies` also matches what `childcare_expenses` already subtracts, so the two sides of the household stay on one variable.

Downstream note for anyone else adding the subsidy back in the omitted states: this PR moves it inside `household_net_income` everywhere, so that workaround has to stand down on the release that carries it.

MaxGhenis (commented):
Thanks for picking up #9405, and sorry this sat for four days without a review.

Replacing the hand-listed state entries with `child_care_subsidies` is the right fix. There are 51 `*_child_care_subsidies` variables and `gov.hhs.ccdf.child_care_subsidy_programs` lists all 51 for 2025, so the aggregate cannot drift the way the hand list did.

One thing blocks merge. `Full Suite - Rest` fails in `test_aggregate_list_integrity.py::test_members_are_state_gated`:

```
gov/household/household_state_benefits.yaml members without a StateCode gate in their defined_for chain (dc_ctc-style leak risk): ['child_care_subsidies']
```

That guard is recent (#9497) and only walks `defined_for`, so it cannot see that every member of the aggregate's `adds` list is state-gated. The problem is the guard, not your change. Either teach it to accept a member with no `defined_for` when everything in its `adds` list is itself state-gated, or add an explicit exemption for `child_care_subsidies` with a comment saying why. The first is better because the next national aggregate will hit the same wall.

Three things to know about, none blocking:

- **Vermont.** `vt_ccfap` pays the state rate whatever the provider charges, which is Vermont's rule since December 2023, so it can exceed the family's child care cost. With this PR that amount reaches `household_net_income` for the first time. A Vermont case that pins it would make the number a decision instead of a surprise.
- **SPM side.** `spm_unit_benefits` still lists only `co_ccap_subsidy`, `wa_child_care_subsidies` and `wv_child_care_subsidies`, so household and SPM net income will disagree on child care in most states. Fine as a follow-up, but worth a line in the PR body.
- **Tests.** The two new cases set the state variable as an input, so they check list membership only. One case that computes a real subsidy in a newly connected state would be stronger.

After the guard fix, please merge main so CI runs on a current base.

PR #9506, opened 2026-09-16
No marketplace premium for a person in the Medicaid coverage gap

### LINKED ISSUE #9472: Adults in the Medicaid coverage gap are charged the full benchmark premium
**Observed** (API, 2026): a single childless 30-year-old in Wyoming with **$0 income** gets `spm_unit_medical_out_of_pocket_expenses` = $11,769 with `medicaid` = 0 and `premium_tax_credit` = 0. Texas: $6,962. The premium is flat below 100% FPL and drops to $336 at $16,000, so `household_net_income` jumps by $11,433 when the household crosses the poverty line — a large negative cliff in the wrong direction.

**Cause**: `takes_up_aca_if_eligible` defaults to `True`, and `pays_aca_premium` tests immigration status, other coverage and age, but not whether any subsidy exists. Adults below 100% FPL in a non-expansion state have no Medicaid pathway and no PTC (26 CFR 1.36B-2(b)(1)), yet the model charges them the full benchmark premium.

**Expected**: these households are uninsured in practice — Wyoming DOH's 2026 coverage chart labels 0–99% FPL "No Coverage" for non-disabled adults. Premium should be $0 there (or take-up should require subsidy eligibility). Affects AL, FL, GA, KS, MS, SC, TN, TX, WY (WI covers adults to 100% FPL).

**Repro**: `POST /us/calculate`, one person age 30, `employment_income` 0 / 8000 / 12000 / 16000, `state_name: WY`, requesting `medicaid`, `premium_tax_credit`, `spm_unit_medical_out_of_pocket_expenses`.

Related: #8168 (marketplace take-up calibration).

mikewolfd:
Fixes #9472, on the gate you specified there.

`pays_aca_premium` never asked whether a subsidy existed, so an adult under the premium tax credit's income floor with no Medicaid pathway was charged the full benchmark premium — $11,769 a year for a Wyoming adult at $0 of income, $6,962 for a Texas adult at $8,000.

**The gate is the coverage gap only:**
- under the floor read from `ptc_income_eligibility` (not "no subsidy", so above the ceiling — where the 2026 subsidy cliff returns — full price still applies);
- with no Medicaid pathway (already in `ineligible_coverage`; a non-expansion state or a status bar);
- and without the below-poverty exception for lawfully present immigrants, honored while 26 U.S.C. 36B(c)(1)(B) was in force and gone from 2026 under P.L. 119-21, so the pre-2026 behavior for that group is unchanged.

**Checked on six edges** (single adult, 2026): WY $0 → pays no premium; TX $8,000 → none; TX $16,500 (just over the floor) → pays, PTC $6,615; TX $90,000 → pays full price; CA $8,000 → Medicaid, unchanged; PA permanent resident under the five-year bar at $12,000 → in the gap, none.

**Tests:** four new cases in `pays_aca_premium.yaml` (gap, floor, the 2025 exception, its 2026 repeal). One CSR case re-pinned — a below-FPL lawfully present person in Texas after the repeal now has no plan rather than an unsubsidized silver one. Suites green: aca 282, hhs/medicaid 501, household 551, ca 670, co 610, nm 548, nj 620, wa 801, partners 620.

https://claude.ai/code/session_01Q7xBzQqLTkE1HJLYJbG5S3


hua7450 (changes requested):
The coverage-gap fix makes sense. Could we make the new gate respect the eligibility scale’s result as well as its threshold?

Currently, a reform setting the first `ptc_income_eligibility` bracket’s `amount` to `true` would make people below 100% FPL income-eligible, but `thresholds[1]` would remain `1`. The new gate would still set `pays_aca_premium=False`, preventing that reform from providing PTC.

One possible adjustment is:
```python
in_coverage_gap = (
    (magi_frac < floor)
    & ~p.ptc_income_eligibility.calc(magi_frac)
    & ~below_fpl_exception
)
```

Please add YAML regression coverage for that reform, plus a below-poverty TX citizen household in 2025 using income and state inputs to exercise the new gate through to premiums and medical out-of-pocket expenses.

MaxGhenis (commented):
Agree with @hua7450's two requests. Two more things to settle in the same pass.

**`pays_aca_premium` is reused by state programs as a test with no income in it.** After this PR it carries an income floor, and these seven variables read it: `ca_premium_subsidy_eligible`, `co_premium_assistance`, `co_premium_assistance_eligible`, `nj_njhps_member_eligible`, `nm_premium_assistance_eligible`, `nm_premium_assistance_mih_eligible` and `wa_cascade_care_savings_member_amount`. New Jersey's says so in a comment ("`pays_aca_premium` is a Person-level test with no income component") and then applies its own band, `magi_fraction >= p.fpl_floor`. New Mexico's says it reuses "the non-income internals of `is_aca_ptc_eligible`". With the new gate, a below-poverty member with no Medicaid pathway fails these state tests through `pays_aca_premium` whatever the state's own floor is. That may be right for some of these programs and wrong for others, for example a state program built for people the federal credit excludes. Could you check each reader, say in the PR which ones are meant to inherit the gate, and update the two comments?

**The comment states a behavioural assumption as a fact.** "Nobody in that gap buys the full-price benchmark plan" is a modelling choice, and a reasonable one: [45 CFR 155.305(a)](https://www.law.cornell.edu/cfr/text/45/155.305) has no income minimum for enrolling in a plan, so a person in the gap may buy at full price. Wording it as an assumption ("the model assumes no one in the gap buys...") would keep the variable honest about what it encodes.

hua7450 (approved):
The requested changes are resolved: the coverage-gap gate respects the eligibility scale, and derived-input regressions cover the below-poverty reform and the 2025 Texas household through premiums and medical expenses. The state-program checks and modeling-assumption documentation are also addressed. The 86 focused tests passed, along with lint and formatting checks. No remaining blockers from my review.

mikewolfd:
@hua7450 — done, in the shape you suggested.

I verified the parameter's API rather than assuming it: `gov.aca.ptc_income_eligibility` is a `SingleAmountTaxScale` with `amount_unit: bool`, and policyengine-core preserves the bracket dtype, so `.calc` returns a boolean array and `~` is the logical negation the gate needs — your snippet works verbatim.

One addition: I kept the `magi_frac < floor` term. Without it the gate also fires *above* the eligibility ceiling, where `~calc` is likewise true in 2026, which would hand a full-price enrollee a free plan.

Your case is real, measured both ways on a Texas single adult at $8,000 in 2026 with `gov.aca.ptc_income_eligibility[0].amount` set true:

| | threshold-only gate | scale-result gate |
|---|---|---|
| `pays_aca_premium` | false | true |
| `aca_ptc` | $0 | $7,670.46 |
| `marketplace_net_premium` | $0 | $168.00 |

Four YAML cases added to `pays_aca_premium.yaml`: that reform, the same household without it, and the below-poverty Texas citizen you asked for — 2025, income and state inputs only — carried through to `selected_marketplace_plan_premium_proxy`, `marketplace_net_premium`, `spm_unit_health_insurance_premiums`, `spm_unit_medical_out_of_pocket_expenses` and the effective actuarial value ($6,194.58 → $0), with a $30,000 Texas companion that still pays $588 at AV 0.87 so the tests show the gate touching only the gap.

`gov/aca` is 286 passing; the branch is current with `main`.

mikewolfd:
@MaxGhenis — checked all seven, and the answer is that none of them moves. But the reason is not the one I first reached for, so let me give the whole dependency picture, because your worry was better founded than a `grep` suggests.

**The direct readers.** Only two of the seven actually call `pays_aca_premium`: `nj_njhps_member_eligible` and `nm_premium_assistance_mih_eligible`. The other five call `is_aca_ptc_eligible` and mention `pays_aca_premium` only in comments.

**That alone does not settle it**, which is the part worth flagging. `co_premium_assistance` reads `slcsp`, and the SLCSP family-tier build-up (`slcsp_family_tier_category`, `slcsp_age_curve_amount_person`, `slcsp_family_tier_multiplier`) reads `pays_aca_premium` directly. So Colorado reaches the gate indirectly, and a mixed household — one member gated, another not — can move where a single adult cannot. **The path is live**: across 50 households, `slcsp` moves in 40 of them, including every Colorado case (CO 2026 mixed-status below-poverty: $15,735.81 → $0).

**No state amount moves anyway, and here is why.** The gate is the exact complement of `is_aca_ptc_eligible`'s own income test — `in_coverage_gap` requires `~calc(magi_frac) & ~below_fpl_exception`, which is `~is_income_eligible`, the other conjunct in that variable. So a person in the gap was already PTC-ineligible; setting `pays_aca_premium` false cannot change it. Checked rather than argued: 10,116 values over 2024–2027 × three immigration statuses × 281 FPL points from 0 to 700% — no difference in `is_aca_ptc_eligible`, `aca_ptc` or `person_receives_aca`. And every one of the seven gates on PTC eligibility or on a band above the gap, so the SLCSP movement never reaches an amount.

**Measured on the households a uniform sweep cannot show** — mixed-status (a lawfully present immigrant under the five-year bar beside citizens), all-barred, and citizen families, below poverty and at 150% FPL, 2025 and 2026, all five states, every Medicaid pathway forced off, run against `main` and against this branch: **0 differences in any state amount or eligibility variable, in 50 of 50 cases.** The 120 differences are all in `pays_aca_premium`, `slcsp` and `marketplace_net_premium` — what the PR intends to change.

Per reader: CA unchanged (Covered California's band starts at exactly 1.00 FPL); CO eligible and CO amount unchanged (both gate on `is_aca_ptc_eligible`, which is invariant, even though the amount's SLCSP input moves); NJ unchanged (138% floor); NM base unchanged (requires `is_aca_ptc_eligible`); NM MIH unchanged (requires income above 400% FPL); WA unchanged (Group 1 requires `is_aca_ptc_eligible`, Group 3 is undocumented members who fail the status test either way). The geography agrees: all five states hosting these programs expanded Medicaid, so a below-poverty adult there already failed on `ineligible_coverage` before any income test — the gap lives in TX, WY, FL and GA, none of which has one of these programs.

**On the shape**, I kept the gate inside `pays_aca_premium` rather than splitting it. The label is what it now computes — a person who does not enrol pays no marketplace premium — and every reader uses it as the enrolment test (the per-person SLCSP/LCBP build-up, the family-tier count, the plan proxy, the state wraps); a state wrap on a plan nobody bought is not owed. A status-only twin would have no consumer today. What I did instead was make the contract explicit: `pays_aca_premium` now carries a `documentation` field saying it is an enrolment test with an income component in it, and that a reader wanting the status and coverage conditions alone should not reuse it; and both misleading comments are corrected — New Jersey's "no income component" and New Mexico's "non-income internals" now say what the variable carries and why it cannot bind where they read it. **If you would still rather have the split, say so and I will do it in this PR** — it is small from here, and the SLCSP path above is a fair argument for making the two concepts separately nameable.

Your second point is taken: the comment no longer states behaviour as fact. It says the model assumes nobody in the gap buys the full-price benchmark plan, with 45 CFR 155.305(a) cited on the variable for there being no income minimum on enrolling.

Also merged current `main` (the branch was 44 behind; only Washington's Cascade Care Savings files moved under anything this PR touches) and re-ran: aca 286, ca 678, co 611, nj 628, nm 548, wa 808, hhs/medicaid 501, household 558, partners 620, contrib/aca 44.

PR #9514, opened 2026-09-17
Register the remaining five state premium programs in household_health_benefits

mikewolfd:
## Problem

`healthcare_benefit_value.py` adds all nine state marketplace premium programs, but `gov.household.household_health_benefits` — the list `household_health_benefits` sums, and so the one that reaches `household_benefits` and `household_net_income` when `gov.simulation.include_health_benefits_in_net_income` is on — carried only four: California, Colorado, Maryland and New Mexico. Covered Connecticut, ConnectorCare, NJ Health Plan Savings, Vermont Premium Assistance and Washington Cascade Care Savings were computed and dropped.

This is the same omission caught for Maryland in review (C2, a6acf1f: "register md_premium_assistance in household_health_benefits.yaml … not only healthcare_benefit_value.py"); the five programs added since repeated it.

## Fix

Add the five variables to `household_health_benefits.yaml` in every period they exist.

## Tests

Seven new cases in the `household_health_benefits` YAML test pin the flow: five forced-value cases, one derived Washington unit (12 × $55 = 660, under the cap) and one derived New Jersey unit (12 × $100 × 1 member = 1,200, under the residual), so each registration is checked end to end. `policyengine-core test policyengine_us/tests/policy/baseline/household`: 558 pass.

https://claude.ai/code/session_01Q7xBzQqLTkE1HJLYJbG5S3


hua7450 (approved):
Reviewed the five registrations and aggregation paths. The default health-benefits flag remains false, so baseline outputs are unchanged and the combined health-inclusive output counts these subsidies once. Validated 26 repository tests and 21 additional checks. The separate combined-output double-counting issue when the flag is explicitly enabled can be addressed in a follow-up.

Issue #1189, opened 2022-07-28
Georgia income tax


MaxGhenis:
* [2021 booklet](https://dor.georgia.gov/document/booklet/2021-it-511-individual-income-tax-booklet/download)
* [Chapter 7](https://advance.lexis.com/documentpage/?pdmfid=1000516&crid=fcb8a062-e7f6-47c7-b75d-20561b282b9a&nodeid=ABWAALAAB&nodepath=%2FROOT%2FABW%2FABWAAL%2FABWAALAAB&level=3&haschildren=&populated=false&title=CHAPTER+7+Income+Taxes&config=00JAA1MDBlYzczZi1lYjFlLTQxMTgtYWE3OS02YTgyOGM2NWJlMDYKAFBvZENhdGFsb2feed0oM9qoQOMCSJFX5qkd&pddocfullpath=%2Fshared%2Fdocument%2Fstatutes-legislation%2Furn%3AcontentItem%3A6348-G0H1-DYB7-W3J7-00008-00&ecomp=vgf5kkk&prid=84953788-e42b-4311-a60c-458ea00138ef) (income tax) and [chapter 7A](https://advance.lexis.com/documentpage/?pdmfid=1000516&crid=6abe9219-3f97-4123-bbf1-3304170588ff&nodeid=ABWAAMAAB&nodepath=%2FROOT%2FABW%2FABWAAM%2FABWAAMAAB&level=3&haschildren=&populated=false&title=CHAPTER+7A+Tax+Credits&config=00JAA1MDBlYzczZi1lYjFlLTQxMTgtYWE3OS02YTgyOGM2NWJlMDYKAFBvZENhdGFsb2feed0oM9qoQOMCSJFX5qkd&pddocfullpath=%2Fshared%2Fdocument%2Fstatutes-legislation%2Furn%3AcontentItem%3A6348-G0H1-DYB7-W3SG-00008-00&ecomp=vgf5kkk&prid=84953788-e42b-4311-a60c-458ea00138ef) (tax credits) of Official Code of Georgia Annotated

MaxGhenis:
Closing as implemented in current main. Georgia now has modeled income tax rate schedules, standard deduction, exemptions, low-income credit, child/dependent care credit, credit ordering, AGI/taxable-income variables, and integration/unit tests under policyengine_us/tests/policy/baseline/gov/states/ga/tax/income/. The remaining open Georgia state-income-tax child issues identified here are now closed as implemented.

PR #9529, opened 2026-09-18
Apply Montana 2021 income tax rebate once per return

### LINKED ISSUE #9528: Montana 2021 income tax rebate clipped per spouse and distorts the filing election
## Bug

Montana's 2021 income tax rebate ([MCA 15-30-2191](https://archive.legmt.gov/bills/mca/title_0150/chapter_0300/part_0210/section_0910/0150-0300-0210-0910.html)) is the lesser of the return's 2021 income tax liability or a per-return cap ($1,250, or $2,500 for a joint return). PolicyEngine modeled it as a **per-person non-refundable credit** ($1,250 in each spouse's column). Two problems on a combined-separate (MFS-on-same-return) filing:

1. **Per-column clipping** — the rebate was capped at each spouse's separate-column liability, so a couple whose combined liability exceeds the cap could still lose part of the rebate when one spouse's column liability is below their $1,250 share.
2. **Election distortion** — because the rebate was a non-refundable credit, it entered the separate-vs-joint election (which minimizes tax after non-refundable credits). Zeroing it (as the TAXSIM emulator does to report `srebate`) flipped the election, so the measured rebate under-reported.

Found via [policyengine-taxsim #1189](https://github.com/PolicyEngine/policyengine-taxsim/issues/1189): a joint couple with a 2021 liability of $4,701 (Form 2 columns $4,195 + $506) received only **$1,949** of the $2,500 rebate.

## Fix

Model the rebate once per return:
- Remove `mt_income_tax_rebate` from the non-refundable credit list so the filing election no longer depends on it.
- Add `mt_income_tax_before_2021_rebate` — the elected income tax before refundable credits and the rebate (Form 2 line 20, excluding the rebate).
- Make `mt_income_tax_rebate` a tax-unit variable = `min(per-return cap, mt_income_tax_before_2021_rebate)`, and subtract it in `mt_income_tax_before_refundable_credits_unit` (read by `state_income_tax`) and `mt_income_tax`.

The rebate now reports its capped value and reaches the full $2,500 for unequal joint filers whose combined liability exceeds the cap, without altering which filing status minimizes tax.

DTrim99:
Closes #9528.

Montana's 2021 income tax rebate (MCA 15-30-2191) is the lesser of the return's 2021 liability or a per-return cap ($1,250, or $2,500 joint). It was modeled as a per-person non-refundable credit ($1,250 per spouse's column), which on a combined-separate return (1) clipped the rebate to a spouse's separate-column liability and (2) entered the separate-vs-joint election, so zeroing it (as the TAXSIM emulator does for `srebate`) flipped the election and under-reported the rebate — [policyengine-taxsim #1189](https://github.com/PolicyEngine/policyengine-taxsim/issues/1189): $1,949 of a $2,500 rebate for a couple with $4,701 of 2021 liability.

## Changes
- `credits/non_refundable.yaml`: drop `mt_income_tax_rebate` from the 2021 list so the election is rebate-independent.
- New `mt_income_tax_before_2021_rebate`: elected income tax before refundable credits and the rebate (Form 2 line 20 without the rebate).
- `mt_income_tax_rebate`: now a tax-unit variable = `min(per-return cap, mt_income_tax_before_2021_rebate)`.
- `mt_income_tax_before_refundable_credits_unit` (read by `state_income_tax`) and `mt_income_tax`: subtract the rebate once per return.
- Tests: rewrote the rebate test (unit-level, capped, plus a #1189 regression where the lower spouse's $206.80 column keeps the full $2,500); moved the election-min tests to `mt_income_tax_before_2021_rebate`; updated the issue-9012 integration case and moved the TAXSIM-parity rebate zeroing to the tax-unit level.

## Validation
The taxsim emulator on the #1189 record now returns siitax **$2,201.39** and srebate **$2,500** (was $1,949), with `staxbc` corrected to $4,701 (separate filing, matching the taxpayer's Form 2 columns). The full Montana test suite passes (one unrelated pre-existing `mt_taxable_social_security` case aside, which is a dataset/version difference, not touched here).


MaxGhenis (commented):
Thanks for digging into this. I think the rebate rule here goes the wrong way for couples who file separately on the same form, so I would hold off merging.

**The cap follows the Montana return, and status 2a is two returns.** The DOR [rebate report](https://archive.legmt.gov/content/Committees/Interim/2023-2024/Revenue/Meetings/May-2024/5.1-DOR-rebate-report.pdf#page=6) says: "For individuals, or those who filed married filing separately, the maximum rebate was $1,250 or the amount the taxpayer reported at line 20 ... whichever was lower. For married couples who filed a joint return, the maximum rebate was $2,500." On page 10 it adds that "married couples who filed separately were ... treated as separate individuals," and that those couples asked why they did not get $2,500. The 2021 Form 2 labels column B "for spouse when filing separately using filing status 2a."

This PR sets `cap = p.amount[filing_status]` from the federal filing status, so a federally joint couple gets the $2,500 cap even when `mt_files_separately` is true. The new case "Unequal joint filers keep the full rebate without per-column clipping" pins exactly that: `mt_files_separately: true` with `mt_income_tax_rebate: 2_500`. Under the DOR rule, filing separately gives that couple min(1,250, 5,099.51) + min(1,250, 206.80) = 1,456.80 of rebate and 3,849.51 of tax, so they file jointly and pay 3,123.22.

**Main already gets the tax right.** Flooring each column at zero after a $1,250 credit is the same as subtracting min(1,250, line 20) per spouse, and the joint path does the same with 2,500. `mt_files_separately` then picks the cheaper of the two. For the 90,000 / 15,000 couple, main returns that 3,123.22 and this PR returns 2,806.31.

**What is wrong on main is the reported value.** `mt_income_tax_rebate` returns the uncapped $1,250 per spouse, so it overstates the rebate whenever line 20 is below the cap. Reporting min(cap, line 20), per column when the couple files separately, would fix #9528's second point without moving anyone's tax.

Smaller things:

- `mt_income_tax_rebate` moves from Person to TaxUnit. Anyone who sets it per person (the TAXSIM parity tests did) gets a parsing error after this.
- The three references are replaced by one URL that now renders only "15-30-2191. Terminated." The archived statute and the DOR report carry the amounts.
- `mt_income_tax_before_refundable_credits_unit` changes formula, and its test file is renamed onto the new variable, so it ends up with no test of its own. The head of household and surviving spouse cases are dropped.
- `amount.yaml` gives `SURVIVING_SPOUSE` $2,500. The statute's $2,500 is only for "a married couple filing a joint return," so that one looks like $1,250. That predates this PR.

DTrim99:
## Review fixes applied (@MaxGhenis)

Thanks for the detailed review. All points addressed:

- **B1 (blocking) — per-column cap implemented, tax unchanged.** The tax mechanics are reverted to `main` (byte-identical), and the 2021 rebate is now a reporting-only value capped per column at `min($1,250, that column's pre-2021-rebate liability)` via a new reporting-only cap-base variable. No one's tax moves. Verified: `mt_income_tax` / `mt_income_tax_before_refundable_credits_unit` diffs vs `main` are empty, and all 20/20 `main` integration cases reproduce identically.
- **Person entity kept** on `mt_income_tax_rebate` (no Person→TaxUnit change), so per-person setters/TAXSIM-parity inputs still parse.
- **3 references restored** on the rebate variable (archived MCA 15-30-2191, HB 192 enrolled-bill PDF, DOR May 2024 rebate report), matching `amount.yaml`.
- **Tests:** HOH and surviving-spouse (QSS) rebate cases restored; separate-filer cap boundary cases (just-below / just-above $1,250) added; the unit variable's own test restored. **33/33 green.**

Two notes for you:

- **(a) One example figure in the review needed correction.** The 90k/15k couple elects **separate** filing under `main`'s optimizer, so its tax is **$5,306.31**, not the joint $3,123.22 cited in the review (engine-verified). The rebate for that case reports `[1250, 206.80]` per column.
- **(b) The pre-existing `SURVIVING_SPOUSE: $2,500` in `amount.yaml` was left as-is.** It predates this PR and the review only hedged on it ("looks like $1,250"), so I kept it out of scope here — happy to open a follow-up issue if it should be $1,250.

PavelMakarchuk (changes requested):
The revised per-column caps and retained Person entity resolve the earlier concerns. However, deleting `mt_income_tax_rebate` from `credits/non_refundable.yaml` removes its aggregation into household resources without adding a replacement payment path. The new rebate output is calculated but no longer contributes to household net income.

The PR's own single-filer test increases tax from $1,089.51 to $2,339.51; the joint example increases it from $18.22 to $2,518.22. Keeping a separate payment outside Form 2 tax can be intentional, but the payment still needs to reach disposable income under the chosen booking year. Please restore that effect and add an integration assertion for household resources. The claim that tax is unchanged also needs updating.

Please add source references to `mt_income_tax_before_2021_rebate` and assert actual rebate/cap values in the advertised boundary test, which currently checks only the filing election. The current PR description also still describes a TaxUnit implementation that this head no longer uses.

DTrim99:
## Fixes applied (4a3b9156a7): Pavel's review

**Payment path restored.** `mt_income_tax_rebate` is now the third `adds` entry of `mt_refundable_credits`. It is booked to TY2021 as a Montana refundable payment, following the repo's existing prior-year rebate convention (GA HB 162 is the closest analog; CT, NM and RI work the same way). It sits in exactly one aggregate. It is not in any benefits list or `spm_unit_benefits`, and not in `credits/refundable.yaml`, because that list feeds the elderly homeowner/renter credit's gross-income line and would inflate 2021 household income for that credit. SPM resources pick it up through `state_income_tax`.

| 2021 case | `household_net_income` before → after | `state_income_tax` before → after |
|---|---|---|
| Single, $50k | 40,940.49 → **42,190.49** (+1,250) | 2,339.51 → **1,089.51** |
| Joint, $60k / $0 | 51,901.78 → **54,401.78** (+2,500) | 2,518.22 → **18.22** |

Both cases are now integration tests at the bottom of `mt/tax/income/integration.yaml`. They also assert `household_refundable_state_tax_credits`, `mt_refundable_credits` and `household_tax_before_refundable_credits`.

**The separate-vs-joint election stays rebate-free.** The same per-person rebate is subtracted on the separate and joint paths, so it drops out of `mt_files_separately`. This matches the DOR rebate report ([printed p.11](https://archive.legmt.gov/content/Committees/Interim/2023-2024/Revenue/Meetings/May-2024/5.1-DOR-rebate-report.pdf#page=13)): couples' rebates "were affected by the filing status they chose on their 2021 Montana tax return", and amending to change status "would not result in a greater rebate." The 2021 returns were filed in 2022, before HB 192 passed.

**Rebate zeroed outside 2021.** Checking for leaks turned up that parameter backdating had copied the 2021 amounts into 2015–2020, so booking the rebate as a refundable credit would have paid $1,250 in those years. `amount.yaml` now starts with `0000-01-01: 0` for every filing status, the same pattern as GA HB 162, and a 2020 guard case covers it.

**Other asks**
- `mt_income_tax_before_2021_rebate` now cites MCA 15-30-2191(1)(a), (2), (6) (archived), HB 192 Sec. 2(1)(a) and (2), 2021 Form 2 lines 18–20 and Column B, and DOR report printed pp.10–11.
- **Boundary test.** The old "just below the cap" case was mislabeled: the second column's line 20 was 1,994.51, above the cap. The pair now sits on either side of the $1,250 cap and asserts the cap base, the rebate, `mt_files_separately` and `mt_income_tax`:
  - $34,000: line 20 is 1,235.51, so the rebate is 1,235.51.
  - $34,500: line 20 is 1,270.01, so the rebate is 1,250.
- **Tax claims.** Montana tax values are back to main's, except where the rebate-free election changes the filing status (below). The changelog and the PR description are updated, and the description now describes the Person-level implementation.

**@MaxGhenis: the $90k/$15k couple.** Main gives 3,123.22 (joint, $2,500 rebate). This head gives **3,849.51** (separate, 1,456.80 rebate). That matches the figure you worked out for separate filing. Without the rebate, separate filing is cheaper (5,306.31 vs 5,623.22), which is the status they would have filed in 2022. My earlier 5,306.31 note left the rebate out entirely; that gap is what this commit closes.

**Verification.** Local runs: `credits/rebate/` 14/14, `tax_calculation/` 3/3, MT `integration.yaml` 22/22. `ruff format` is clean. An independent validator pass came back clean, and no partner files were touched.

PR #9531, opened 2026-09-18
Add AZ refundable dependent credit and MA/UT dependent exemption age-split contribs

DTrim99:
Closes the remaining model-side gaps for converting state dependent provisions into under-18 credits (the CPID conversion toolkit; dashboard-side wiring follows separately).

## Arizona — refundable dependent tax credit
`gov.contrib.states.az.dependent_credit.refundable.in_effect` pays the portion of the dependent tax credit stranded past the nonrefundable liability cap as a refund (same shape as the GA refundable CTC contrib: recompute the ordered-credit limit, refund `max(potential − applied, 0)`, per-period `in_effect` gate). The new `az_refundable_dependent_tax_credit` is appended to the AZ refundable-credit list from 2024 via `modify_parameters`. It reads `az_dependent_tax_credit_potential`, so it composes with the existing AZ dependent-credit contrib (flat amount / age limit) — covered by a composition test.

## Massachusetts — dependent exemption age split
`gov.contrib.states.ma.dependent_exemption` reprices the Part B dependent exemption per dependent with an optional age limit: dependents under the threshold take the reform amount (0 = repeal), older dependents keep the baseline $1,000. The amount uses a negative sentinel meaning "track baseline" (AZ/RI pattern), so enabling only the age limit is a pure no-op — asserted in tests. Personal, blind, and aged exemptions are untouched, as is the No Tax Status worksheet (leaving NTS at baseline means the repeal never harms an NTS-protected filer).

## Utah — dependent exemption age split
`gov.contrib.states.ut.dependent_exemption` does the same for the per-dependent exemption inside the Taxpayer Tax Credit, splitting it from the shared (uprated) personal-exemption amount — the piece the blanket `repeal_state_dependent_exemptions` contrib could only neutralize whole. The 59-10-1018(1)(g) newborn additional exemption follows the reform amount (newborns are under any threshold), asserted in a dedicated test.

## Notes
- All three creators use the standard 5-year `in_effect` lookahead.
- NC needs no equivalent: `nc_child_deduction` counts `ctc_qualifying_children` (under 17), so a plain repeal is already the under-18 repeal.
- 12 YAML tests across the three contribs: baseline-equivalence sentinels, under-18 repeals preserving an older dependent, AZ liability-cap arithmetic + contrib composition, UT newborn interaction. `make format` applied; ruff clean.


MaxGhenis (commented):
Thanks for these three. CI is green, but one thing in the Arizona reform needs a fix before this merges.

**The Arizona refund is registered more than once.** `modify_parameters` appends `az_refundable_dependent_tax_credit` to `gov.states.az.tax.income.credits.refundable` without checking whether it is already there:

```python
current = node("2024-01-01")
node.update(..., value=current + ["az_refundable_dependent_tax_credit"])
```

A reform's `modify_parameters` runs more than once per build. On main at 25380edf72 I applied the same unguarded append through `Simulation(reform=...)`: the function ran three times and the 2026 list held the new name three times, so `az_refundable_credits` would add the refund once per copy. Every list-append reform on main guards against this, for example [`ct_tax_rebate_2026.py`](https://github.com/PolicyEngine/policyengine-us/blob/main/policyengine_us/reforms/states/ct/tax_rebate_2026/ct_tax_rebate_2026.py#L29-L37) and [`id_ctc_reform.py`](https://github.com/PolicyEngine/policyengine-us/blob/main/policyengine_us/reforms/states/id/ctc/id_ctc_reform.py#L88-L95):

```python
if "az_refundable_dependent_tax_credit" not in current:
    node.update(..., value=list(current) + ["az_refundable_dependent_tax_credit"])
```

**No test would catch it.** The four Arizona cases assert only `az_dependent_tax_credit_potential` and `az_refundable_dependent_tax_credit`. Could one case also assert `az_refundable_credits` and `az_income_tax`? That is the aggregate where a duplicate shows up. The Massachusetts and Utah cases have the same gap: none asserts the state income tax.

**Smaller: the Massachusetts and Utah formulas never read their own `in_effect`.** The Arizona variable gates per period with `where(p.refundable.in_effect, ...)`, and its comment explains why: the reform is installed for the whole simulation once it is active in any of the next five years. The Massachusetts and Utah replacements read `amount` and `age_limit.in_effect` only, so a year with `in_effect: false` and a non-negative `amount` still gets the reform amount. Gating on `pc.in_effect` per period would make the three consistent.

DTrim99:
Thanks Max — all three addressed in 4d986ec77e:

1. **Registration guard**: `modify_parameters` now checks `if "az_refundable_dependent_tax_credit" not in current` before appending (the ct_tax_rebate_2026 / id_ctc pattern), so repeated runs register the refund exactly once.
2. **Aggregate asserts**: the no-liability AZ case now asserts `az_refundable_credits: 250` and `az_income_tax: -250` — a double registration would surface as an extra $250 per copy. (Moved that case's AGI above the excise-credit limit so the aggregate isolates the new credit.) The MA and UT repeal cases now assert the state income tax with hand-checkable arithmetic: MA $2,020 → $2,070 (+5% of the repealed $1,000), UT $1,379 → $1,508 (+6% of the repealed ~$2,159 through the taxpayer credit).
3. **Per-period gates**: the MA and UT replacement formulas now gate on their own `in_effect` per period, mirroring the AZ refund's gate and comment; each file gains an inertness case asserting that a set `amount` does nothing in a period where `in_effect` is false.

14 YAML tests pass locally (4 AZ, 5 MA, 5 UT); ruff clean.

DTrim99:
CI note for the remaining red check: **Quick Feedback (Selective Tests + Coverage) has been runner-evicted twice at the tail of the job with every test passing.** Both attempts show identical behavior — the selective batches (code-health file, then the az/ma/ut contrib dirs and direct YAMLs) all complete with green dots, and then the runner receives a shutdown signal during the final batch/coverage step (`##[error]The runner has received a shutdown signal`, exit 143 — the same OOM-eviction signature that hit the Rest job on this PR before the rebase). Attempt 1 died 90 minutes in, attempt 2 at 50 minutes, both right after `contrib/states/ut` finishes.

All 34 other checks pass, including the sharded Full Suite contrib jobs that run these same tests without coverage instrumentation. I've kicked off a third attempt; if it survives, great — otherwise this looks like the coverage job's single-process memory ceiling rather than anything in the diff, and I'll defer to you on whether to treat the check as advisory here.

hua7450:
**Arizona: the refund inherits a baseline credit-ordering issue that first affects dollars here**

The refund is computed correctly for the ordering the model uses: `az_refundable_dependent_tax_credit` keys `state_non_refundable_credit_limit` off the same `gov.states.az.tax.income.credits.non_refundable` list and the same `"az_dependent_tax_credit"` ordering key as baseline `az_dependent_tax_credit`, so potential − applied is exactly the stranded portion. The problem is the baseline list itself.

`parameters/gov/states/az/tax/income/credits/non_refundable.yaml` from 2021 on is:

```yaml
- az_charitable_contributions_credit
- az_dependent_tax_credit
- az_family_tax_credit
```

The forms apply the dependent and family credits **before** any Form 301 credit, including the charitable ones:

- 2025 Form 140 booklet: line 49 Dependent Tax Credit, line 50 Family Income Tax Credit, line 51 nonrefundable credits from Form 301, line 52 "Subtract lines 49, 50, and 51 from line 48" ([p. 26](https://azdor.gov/sites/default/files/document/FORMS_INDIVIDUAL_2025_140Booklet.pdf#page=26), [p. 30](https://azdor.gov/sites/default/files/document/FORMS_INDIVIDUAL_2025_140Booklet.pdf#page=30)).
- 2025 Form 301 instructions, Part 2: line 31 tax subtotal; line 32 "the total amount of family income tax credit and dependent tax credit claimed on your income tax form" (Form 140 lines 50 and 49); line 33 = line 31 − line 32, floored at zero; "The total amount of tax credits you use cannot exceed the amount entered on line 33" ([p. 3](https://azdor.gov/sites/default/files/document/FORMS_CREDIT_2025_301_i.pdf#page=3)). The qualifying charitable organizations credit is line 39 of that section, drawing only on what is left ([p. 4](https://azdor.gov/sites/default/files/document/FORMS_CREDIT_2025_301_i.pdf#page=4)).
- The 2022 forms have the same sequence (140i [p. 21](https://azdor.gov/sites/default/files/2023-03/FORMS_INDIVIDUAL_2022_140i-2D.pdf#page=21) and [p. 24](https://azdor.gov/sites/default/files/2023-03/FORMS_INDIVIDUAL_2022_140i-2D.pdf#page=24); 301i [p. 3](https://azdor.gov/sites/default/files/2023-03/FORMS_CREDITS_2022_301i.pdf#page=3)).

In baseline the order is invisible: whichever credit is labelled "applied," the total nonrefundable credits and `az_income_tax` are the same, so no existing test catches it. Once this refund is on, the order changes a dollar outcome.

Example, 2026, single filer, one dependent under 17 (potential $125), `az_income_tax_before_non_refundable_credits` $300, qualifying charitable organizations credit $300, `refundable.in_effect: true`:

| | Form order | Model order |
|---|---|---|
| Dependent credit applied | $125 | $0 (charitable takes the full $300 first) |
| Stranded dependent credit → refund | $0 | $125 |

The form gives the charitable credit a five-year carryforward and pays no refund; the model pays $125.

Suggested handling: this is a baseline fix (reorder the list to dependent, family, then charitable from 2021) rather than a change to this PR, so I'd file it as a follow-up and add a note in the PR that the refund depends on the baseline ordering. Population affected is small (AZ filers with charitable credits and liability below the combined credits), but the direction is one-sided.

DTrim99:
@hua7450 Ordering-dependency note added at `policyengine_us/reforms/states/az/dependent_credit/az_dependent_credit_reform.py` line 89, explaining that the refund inherits the baseline `gov.states.az.tax.income.credits.non_refundable` ordering (charitable before dependent/family, contrary to Forms 140/301). The baseline reorder itself is tracked separately in issue #9559 rather than fixed in this PR. AZ contrib tests unchanged: refundable 4/4 and dependent_credit_reform 5/5 pass.

hua7450 (approved):
Approving at head e316956.

Reviewed the three contribs against A.R.S. 43-1073.01, MA G.L. c.62 §3 and §5, Utah Code 59-10-1018, and the 2022/2025 Forms 140 and 301. The MA and UT inert paths and the negative-sentinel paths reproduce baseline exactly; the AZ refund is exactly the stranded portion and composes with the existing AZ dependent-credit and EITC contribs; the registration guard, aggregate assertions, and per-period `in_effect` gates from Max's review are all in place. The AZ baseline credit-ordering dependency is documented in the formula and tracked in #9559.

34 of 35 checks pass on this head. The remaining Quick Feedback coverage job was runner-evicted (shutdown signal) with every selective batch passing, including all MA and UT contrib cases; the sharded Full Suite Contrib jobs that run the same YAML files pass.

Non-blocking follow-ups noted from the review: the MA No Tax Status / Limited Income Credit thresholds keep the baseline per-dependent add-on (c.62 §5 cross-references the §3(B)(b)(3) exemption), which is worth stating in the parameter description as a deliberate hold-harmless choice; and an at-threshold age case (dependent exactly 18) plus an AZ case where liability absorbs the whole credit would pin the boundaries.

